PR #8, opened 2019-03-12
Plug DRAMSim as an DRAM simulator

ndesh26:
As part of my project, I replaced the native DRAM simulator with DRAMSim simulator. I am creating the pull request to see if having this as an alternative in ChampSim would be beneficial to the community.

The current code is just a proof of concept and *cannot* be merged as is. I am willing to improving/fixing the code if you think that this might be a helpful addition for the ChampSim community.

Thanks


zoumo4913:
Hi, 
I am new at simulator and architecture and I am trying to run Champsim simulator with DRAMSim, but there seem some problems.
When I run build_champsim.sh, such warning occurs with every .cc file:
截屏2020-09-10 上午1 03 32
 If I ignore it and run run_champsim.sh, a segmentation fault occurs in the running. I think this may be caused by introducing DRAMSim simulator. What do you think that I should do to check the issue?
Thank you very much!

ndesh26:
It might be because you are using the latest code of ChampSim. Can you try using the code from [here](https://github.com/ndesh26/ChampSim/tree/wip/dramsim)?

zoumo4913:
I get the code from https://github.com/Prakharj24/ChamSim-CS422A for both ChampSim and DRAMSim2. I will try to change the ChampSim code. Thank you for your help.

zoumo4913:
Hi, 
Thank you for your help. I can run the code on my computer now. My system is MacOs, here are some changes that may help others who use MacOS.
First, in _DRAMSim2/makefile_, the option **-soname** should be modified to **-install_name** to match MacOS.
Second, in _run_champsim.sh_, I need to add one line to export dynamic library in MacOS, like this: **export DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH:$(pwd)/DRAMSim2/**
I think this is caused by the system. Hope these will help to others. Thank you for your help and the work that combine the simulator with dram is really meaningful!

ngober:
This patch has drifted too far from the mainline, and would need significant work to make it mergeable. If you are still interested in it, see the discussions in #390 and submit a new patch.

PR #10, opened 2019-03-26
Buildscript

ngober:
In a project I've been working on, I wanted to pass flags to the preprocessor (-D), but the existing build system did not support it. So, I wrote one that did.

The configuration script in this PR now supports user files from outside the build path (for example, the L1 prefetcher does not necessarily need to be in prefetcher/). It also supports passing compiler flags via CFLAGS, CXXFLAGS, etc. Note that passing compiler flags is currently only possible via the configuration file.

The configuration file can even be overridden. Try './config.sh -h' for more information.

It is strongly extensible, so new features are easy to add to the configuration file.


ngober:
My build script has evolved since the initial pull request. It is now a python (2.7-compatible) script. I chose python because it gave me more flexibility with the CLI than the bash script did. I have also added support for prefetchers that have multiple source files (a huge win!), although the `*.l1d_pref` file must be specified first. This doesn't have to remain this way, but I kept it for the sake of making minimal changes to the structure of ChampSim.

ngober:
Closing this because I feel that #69 is a preferable solution, plus the commit history on this branch has become a headache.

PR #11, opened 2019-03-28
[WIP] Enabling more flexibility in the development using a Cmake build system

itisntalex:
This pull request may sound a bit like a previous one (namely #10) but this work allows ChampSim to be way more flexible in the way we build and develop new methods. It's based on CMake instead of plain old Makefiles which translates into increased flexibility.

Hope this work can help some of you.


ngober:
Hi, I'm the guy from #10! I like your idea to use CMake. It's certainly more clean than my bash script.

One of the common use cases for ChampSim is to compare a bunch of different prefetcher designs in sequence. Say I had 20 different L1 prefetchers that I wanted to test, all against the same trace, What do you think would be the easiest way to handle that? To me, it looks like the only option is to edit the CMakeLists file, where the syntax is a bit arcane.

danielsuo:
I have my own fork of ChampSim that also uses CMake. Nowhere near in a
condition for public consumption, but chiming in because CMake can use glob
patterns, so a typical user wouldn't need to edit the cmake file.

itisntalex:
Hello guys! First of all, I'm glad that some people are kind of enthusiastic about my small contribution. :smile: 

> Hi, I'm the guy from #10! I like your idea to use CMake. It's certainly more clean than my bash script.
> 
> One of the common use cases for ChampSim is to compare a bunch of different prefetcher designs in sequence. Say I had 20 different L1 prefetchers that I wanted to test, all against the same trace, What do you think would be the easiest way to handle that? To me, it looks like the only option is to edit the CMakeLists file, where the syntax is a bit arcane.

To answer your question, I'll start by giving you some background about my use case and why I designed this. I think it'll be of great help to answer your question.

# Story #

As part of my current Ph.D position, I'm working on cache replacement policies and just like you have to try a whole bunch of L1 prefetchers, I have to test me cache replacement policies on many benchmarks (traces). It means that I have to build many variations of ChampSim and eventually execute them.

# Leveraging on the CMake Build system #

As I choose CMake as a build system for my project, I tried to design it in a very flexible way. It translates in some knobs (CMake variables) that you can use to customize your build. Currently, there's one variable per component:
- One for each level of prefetching;
- One for the branch predictor;
- And one for the cache replacement policy.

For instance, if you want to build a variant of ChampSim using a *gshare* branch predictor and a *Next Line* L1 prefetcher, you just have to type the following in your terminal:

```bash
cmake -G "Unix Makefiles" -DCHAMPSIM_BRANCH_PREDICTOR:STRING=gshare -DCHAMPSIM_PREFETCHER_L1D:STRING=l1d_next_line
```

**N.B.:** the exemple above assumes that we are using the directory hierarchy I'm using in my pull request as well as the naming convention.

To give you more insights on how it works in depth; the main CMake script will look for the matching directories in directory hierarchy following the naming conventions I'm using (maybe we should agree on them also...?). if the appropriate folder is found, it will look for a `.cmake` file which contains all the rules to properly build this specific component.

Here is a small example taken from my proposal, it only deals with the branch predictor, but it's all the same for every single component.

```cmake
# Looking for the branch predictor related-files and knobs.
if (EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/src/branch/${CHAMPSIM_BRANCH_PREDICTOR}/${CHAMPSIM_BRANCH_PREDICTOR}.cmake")
    include(${CMAKE_CURRENT_SOURCE_DIR}/src/branch/${CHAMPSIM_BRANCH_PREDICTOR}/${CHAMPSIM_BRANCH_PREDICTOR}.cmake)
else ()
    message(FATAL_ERROR "The branch predictor ${CHAMPSIM_BRANCH_PREDICTOR} could not be found. Aborting!")
endif ()
```

It might sound a bit heavy for a build system on such a project, but in my opinion it allows us leverage on a new scale of flexibility:
- Multiple file predictors and components :pray: 
- Possibility to easily include third-party libraries (if you want to compute some fancy linear algebra?)
- And so on... :wink: 

I hope this story helped you to understand better what is the idea behind this proposal and also how it can help people working with ChampSim. In conclusion, building many variants of ChampSim can be achieved using a simple bash script to change knobs to match your expectations.

ChampSim:
Hi guys,

Thanks for the contribution for Cmake approach. I also like to keep all the files in a uniform/flexible way so there is no reason to refuse this request. However, we'd like to hold all the changes until DPC-3 submission gets finished. At this point, changing the build flow might cause confusion to DPC-3 participants. We will revisit this pull request after DPC-3 gets done.

Jinchun

itisntalex:
> Hi guys,
> 
> Thanks for the contribution for Cmake approach. I also like to keep all the files in a uniform/flexible way so there is no reason to refuse this request. However, we'd like to hold all the changes until DPC-3 submission gets finished. At this point, changing the build flow might cause confusion to DPC-3 participants. We will revisit this pull request after DPC-3 gets done.
> 
> Jinchun

Hello @ChampSim,

Thanks for the reply. As I'm actually working with Prof. Daniel Jimenez, I'm aware of the ongoing contest and I totally agree with your point of view on not breaking the current workflow. I'll just keep on improving it as I'm working. Also, feel free to suggest any ideas or features. I'd be glad to here them and discuss them.

Alexandre V. Jamet.

ngober:
Getting back around to this... With the merge of #69, I think it is unlikely that ChampSim will move in the direction that we need a system like CMake, so I am closing this PR. But, many of the ideas presented here, such as the directory structure, made it into #69, so thank you for the suggestions.

PR #27, opened 2020-02-26
Make build process more flexible

jnider:
There are many options that need to be passed to the build script. The number of options has changed over time, but the 'readme' can't keep up. In fact, many users only want to change a few of these parameters each time, and want to leave the rest as 'default'. Currently there is no way to use default parameters - all parameters must be specified each time. This patchset relaxes that restriction by using environment variables instead of positional parameters. If the environment variable is not set, a default value is used. The user also has the option to set the variable (semi-permanently) in the environment, or set it temporarily for a single build (by prefixing the shell script). All parameters can also accept the value 'help' which will print out valid options, to make it easier for casual/first-time users to learn the system.


GinoAC:
We've updated the current build system to use a configuration file format that fulfills this pull request's intent.

PR #33, opened 2020-05-10
Initialization cleanup

ngober:
This pull request improves the clarity of the initialization steps. The explicit initialization (especially to zero) of member variables is moved out of the constructor. In most cases, this allows the explicit default constructor to be removed entirely, reducing the number of code lines significantly.

Also, a long-standing TODO is fixed in main.cc and the initialization of most of the core model is moved to the O3_CPU constructor. The L1I operate functions are still local to main.cc and should eventually be moved.


ngober:
@sethpugsley I was reviewing #38 and noticed that this may have been accidentally closed in our confusion this past weekend. Did you intend to close this without merging, or was there a reason it is rejected?

sethpugsley:
I did not intentionally reject this, sorry.  Feel free to fix my mistake.

sethpugsley:
This PR has some conflicts that need to be resolved, and it doesn't take into account the modifications to core initialization done by #61, so I think #61 should be merged first, then this should be updated to move those new core initializations also to the O3 constructor, and the conflicts for this one should be resolved, then this should be merged in.

Other than that, I like these changes and it looks good to me.

ngober:
I've resolved the conflicts that were introduced in develop. My basic tests don't show any difference in performance between this patch and the base branch, but we now gain some clarity, because this removes many lines of code that are simply initialization. I also took the liberty to remove a few members that weren't being used, which will also improve code clarity.

One consequence of this is that I was able to merge the CORE_BUFFER and LOAD_STORE_QUEUE classes into one. It doesn't make any difference, but it's nice.

sethpugsley (approved):

sethpugsley:
I think we should finally merge this.  Sorry for not getting to it earlier.  @GinoAC, will you please make one final check and then pull the trigger?

GinoAC (approved):
Looks like a good pass on cleaning up and clarifying the initialization process. I'll merge it into dev now.

PR #34, opened 2020-05-21
Bugfix: Underutilized scheduling width

ngober:
The ROB.next_fetch member is never modified, so it always keeps its default value of zero. This causes the scheduler to always run from ROB.head to ROB.SIZE, which is a bug. This pull request changes it so that it runs from ROB.head to ROB.tail, still halting early if it reaches a scheduled instruction or the machine's schedule width. This results generally in an increased performance in the simulations.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -788,8 +788,6 @@ int main(int argc, char** argv)
 	      ooo_cpu[i].update_rob();
 
 	      // schedule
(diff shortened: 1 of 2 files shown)


ngober:
Please squash this PR when merging. I had introduced a compilation error with the first commit (I was using C++11 arrays), but fixed it in the second.

PR #36, opened 2020-05-22
Added support for streaming traces with wget

ngober:
This PR allows the execution of remote traces, removing the need for them to be stored locally.


GinoAC:
Supports previous format with only difference a check for http as the trace file name.

PR #38, opened 2020-05-26
Cleanup

elbs:
Okay, here's my cleanup. Sorry about the hubbub beforehand--Git is not my forté. I've focused mainly on src/ and inc/, making all files indented (4 spaces) and cleaned from trailing whitespace. I've also made some style choices regarding brackets and for-loops.


ngober:
Before I say this, we really should settle on a style and write it down somewhere. Maybe the wiki would make a good place?.

I like the changes you've made here. I think the whitespace tweaks increase readability. I tend to disagree on your choice of brace placement. I prefer to put braces on their own lines. I think it creates visual space around the surrounded code and allows me to see it in isolation. It also works well with vim's [[ command. Is placing braces on the same line better, in your opinion?

arminvakil:
Hello,
You can check google's code style for C++ so everyone can follow the same standards while developing.

elbs:
I definitely agree on the document. I was thinking of setting up one, but I
wanted to see if this was received well as a first step.

With regards to the brackets, yeah, I can see benefits to both. (Also,
Vim's % command works with these brackets too.) Mainly I did it for line
count reduction, but readability is important too. We also need to make
choices with regards to if-statements that only have one command, etc.

I'll look up others' style documents and try to make a template for what
ours should look like and contain.

Elba

ngober:
I checked out Google's C++ style guide, and I like a lot of things in it. The rules on headers and #include statements, as well as preprocessor macros, are great and avoid a lot of bugs.

One issue we may run into is deciding on non-trivially-destructible objects with static duration. A lot of our models use static durations to emulate hardware, because it's always present. We might relax any rules on that, but we should be clear about what to do with destructors in that case.

ngober:
I was reading through some other projects on GitHub, and came across [this comment left on a Rails PR](https://github.com/rails/rails/pull/13771#issuecomment-32746700). It argues not to permit PRs that are only cosmetic changes, but rather to make the cosmetic changes as we substantially improve the code in question. The point that stands out to me is that it pollutes the git history. If I notice that a thing is broken, I will look at the blame output and try to figure out how it got there and what the intention was. If the blame output just says "whitespace changes", then I do not gain any useful information. It also makes rebasing (other feature branches once this is merged) extremely hard. Git's merge tool is powerful, but naive when it comes to whitespace and brace positioning (it must be). 

What do you think about this? Should we drop this PR and instead write a style document, saying "All future contributions must follow this style", then allow the code style to steadily improve?

itisntalex:
> I was reading through some other projects on GitHub, and came across [this comment left on a Rails PR](https://github.com/rails/rails/pull/13771#issuecomment-32746700). It argues not to permit PRs that are only cosmetic changes, but rather to make the cosmetic changes as we substantially improve the code in question. The point that stands out to me is that it pollutes the git history. If I notice that a thing is broken, I will look at the blame output and try to figure out how it got there and what the intention was. If the blame output just says "whitespace changes", then I do not gain any useful information. It also makes rebasing (other feature branches once this is merged) extremely hard. Git's merge tool is powerful, but naive when it comes to whitespace and brace positioning (it must be).
> 
> What do you think about this? Should we drop this PR and instead write a style document, saying "All future contributions must follow this style", then allow the code style to steadily improve?

As a follow up to your answer, there a couple of tools available around to enforce a style in a project. _e.g._: [editorconfig](https://editorconfig.org/). Coupled with a proper contribution guideline document that would help build consistency in the code. :smile:

sethpugsley:
I like the idea of a single big style correction, but I think we just have to time it correctly.  I propose that the next time we get the develop branch into a position where we want to merge it into master, that we do the merge and then we do all the style changes and push another master update within a couple days of that, before the develop branch has time to wander off into the weeds again.

What if we plan on doing this later this summer?

ngober:
I don't know that I understand the difference between a sweeping style correction today and a sweeping style correction two months from now when we merge to master. Both still pollute the git history and make rebasing feature branches difficult. I think Alex's suggestion to find an editor-agnositc way to enforce style is interesting, but I worry about interoperability. For example, I know that some people on this project use Emacs, I use Vim, yet others might use something less feature-rich, like Mousepad. I don't know that there can be a single solution to make things "just work". 

Instead, I think we should draft up a document that says things like "Name variables like this, don't use implicit conversion constructors, prefer const static variables to preprocessor defines", etc. and take a more step-by-step approach. I think it's possible that the things I'm working on will touch most of the codebase in the next year, so there's a good chance that any problematic lines can be fixed as other things are improved.

itisntalex:
@ngober Well, the tool I proposed I quite interoperable as it provides plugins for most editors around (_e.g._, emacs, vim, atom, etc.), and some others provide built-in support. But anyway, I believe a document stating rules is definitely a go for. The tool I suggested would be a fancy addition.

elbs:
Hi all,

I agree that a large change to the whole source would be a jarring move, and without rigorous step-by-step testing, may unwillingly add bugs from manually-necessitated changes. Maybe over time, we can schedule changes to the base, say, a file at a time, with testing to ensure all is computationally identical. 

I’ve been very sick these past couple weeks (hence few responses) but I do have the style document as a major TODO. Maybe once I have a document done, and we all amend it as we see fit, we can start rolling out changes to the code in a controlled, slow manner. 

Elba

truls:
I've considered creating a PR like this myself so I'm happy to see it discussed.

I think the proper way to deal with this is to use ClangFormat to reformat the code automatically together with a CI hook on GitHub that checks if the code in pull requests is properly formatted. This ensures that a consistent coding style is followed by the current code base and is enforced in future commits. Further, using an automatic code formatter reduces the risk of accidentally introducing semantic changes to the code while reformatting it.

The ClangFormat approach is editor agnostic and is usually implemented in projects by introducing an appropriately named make target, often `make format`.

@ngober I agree that commits containing purely cosmetic changes are, in principle, a bad idea, however, the current inconsistent formatting of the champsim code hinders code comprehension. This is a problem for a project which expects its users to understand and modify its code. MongoDB has a nice blog series ([part 1](https://engineering.mongodb.com/post/succeeding-with-clangformat-part-1-pitfalls-and-planning), [part 2](https://engineering.mongodb.com/post/succeeding-with-clangformat-part-2-the-big-reformat), [part 3](https://engineering.mongodb.com/post/succeeding-with-clangformat-part-3-persisting-the-change)) on ClangFormat. They argue in favor of reformatting the whole code base in a single commit and discuss how to deal with migrating old feature branches onto the reformatted code.

sethpugsley:
I'm a big fan of the idea to use ClangFormat.  I don't have any experience setting it up, though.  I like the idea of not needing to worry about how I format my code, and then have it automatically formatted for me before I can commit it.

Once we have that in place, I like the idea of one big monolithic auto-reformatting push that puts everything into the correct format, never to depart from it again.

elbs:
This sounds fantastic, I agree. Is the setup difficult?

truls:
Cool. It's not hard to set up and I'm happy to provide a PR. The biggest decision to be made is which coding style to enforce. I can try to come up with something that matches the current code.

ngober:
I've created issue #84 to discuss this. Let's move our general formatting and tools discussion there, then we can return to this thread if we want to focus more on Elba's patch.

ngober:
I'm going to close this, since it has gone pretty stale. I think we've settled on something like #93 to handle this idea.

PR #41, opened 2020-06-15
Pipeline decoder

lpha-z:
I think decoder should be pipelined.
This pull request addresses this issue.

If it's intentionally non-pipelined, I think it should be noted somewhere.


ngober:
Thanks for the pull request! I see what this is doing: It is counting only instructions that have not already been checked. I'm merging this into master. In the future, would you please submit pull requests to the develop branch? This makes sure that we can keep our master stable when there are multiple changes being made.

PR #45, opened 2020-06-20
Fix a bug that fetched instructions become unfetched state

lpha-z:
Fetched instructions become unfetched (inflight) state when fetching another instruction that is in the same cache line is started.
This means that fetched instructions sometimes cannot be decoded, leading to performance degradation.


sethpugsley:
This is another good catch.  Thanks for fixing this error.

PR #46, opened 2020-06-21
Reduced complexity of register dependency resolution

ngober:
Recent fixes to the instruction scheduler have placed more significant weight on the efficiency of the scheduler itself. This is one step in the effort to make the scheduler simulate faster. Now, rather than iterate through the ROB for each instruction being scheduled, we store an iterator to the most recent instruction that produces the register we are waiting on, so it can be accessed directly.

A side effect of this change is that the ROB entries are now stored in a std::vector instead of a C-style array. It also removes 0 as an indicator of "none" in source and destination register lists.

My initial tests show that this takes approximately a minute off of the runtime of bwaves-2609B for 10M/100M with no prefetchers. My profiling analysis suggests that, while the scheduler still dominates the runtime, it is at least 25% more efficient.


sethpugsley:
Does this have any impact on reported IPC or any other statistics?

ngober:
I ran it locally on perlbench-570B and bwaves-2609B for 10M/100M, no prefetchers, both before and after this commit, and there was no change in the reported statistics. I can do a more extensive test this evening if you would like.

sethpugsley:
I was just able to test this out, and the reg_dependency changes, merged with the current state of develop, do have different performance than just develop by itself.  Are you still seeing no changes in stats when these new changes are merged in with the latest state of develop?  I'm seeing your new version give <1% additional IPC for the few traces I tested so far.

I'm not seeing any crashes or other problems, but the reported IPC is different, and I'm seeing a wall clock simulation time savings of about 5-15%, depending on the trace.  Is that time savings also in line with what you're expecting?

ngober:
Taking another pass through this, I noticed that my changes interfere with our discussion in #47. I changed the way that registers are stored in the instruction, so it required small changes to `mem_reg_dependence_resolved()`. Maybe that's where the variance comes from? Should I change it to match our discussions? It looks like @lpha-z created a patch but it hasn't been merged (https://github.com/lpha-z/ChampSim/commit/df9ae87e46ac43d69b5f2cecd3a7b4aec2207636).

ngober:
I have submitted #59 because I have noticed as I test this PR that the IPC is highly sensitive to the scheduler. Hopefully, #59 will reduce that sensitivity and allow more reliable timing. Until I can figure that out, I've marked this as draft.

ngober:
I'm withdrawing this PR for now. I have a lot of other work in flight, so this one has atrophied. It's still on my list, but I'll let other stuff stabilize before I return to it.

PR #48, opened 2020-06-24
Upstream Georgios Vavouliotis's Page Walker

JamesPC44:
## Change Summary

Realistic Page Walker provided by Georgios Vavouliotis in `simulated_page_walker()` called by `va_to_pa()`

`README.md` is modified for usage instructions, which are runtime

`va_to_pa()` is modified to accept `ip` and `type` as arguments which are used by the
page walker. This undoes some of Seth's work in [`ee93e2`.](https://github.com/ChampSim/ChampSim/commit/ee93e223aefb311ac820e72013c1f4c46157ab5d) Unsure if that is a problem. If it is, there will need to be another way to get `ip` to `simulated_page_walker()`, through `va_to_pa()`

## Some possible areas of concern:

+ Line 683 of `cache.cc` - if the type passed is appropriate (`RQ.entry[index].type`)
+ Lines 467 and 814 of `ooo_cpu.c`, if the type `PREFETCH` is correct here.
+ Line 826 of `ooo_cpu.c`, which reads: `pf_packet.ip = pf_v_addr;` after modification in `ee93e2` to not be `pf_packet.ip = ip`. This could now be changed back. Should it?

### DIFF
--- a/src/cache.cc
+++ b/src/cache.cc
@@ -2,6 +2,11 @@
 #include "set.h"
 
 uint64_t l2pf_access = 0;
+/* you can implement each page walk cache level as a 2d array */
+uint64_t pml4[PML4_SET][PML4_WAY], pdp[PDP_SET][PDP_WAY], pd[PD_SET][PD_WAY];
+
+/* implementation of LRU policy with counters */
+uint64_t pml4_lru[PML4_SET][PML4_WAY], pdp_lru[PDP_SET][PDP_WAY], pd_lru[PD_SET][PD_WAY];
 
 void CACHE::handle_fill()
 {
@@ -675,7 +680,7 @@ void CACHE::handle_read()
 			  // TODO: need to differentiate page table walk and actual swap
 			  
 			  // emulate page table walk
-			  uint64_t pa = va_to_pa(read_cpu, RQ.entry[index].instr_id, RQ.entry[index].full_addr, RQ.entry[index].address, 0);
+			  uint64_t pa = va_to_pa(read_cpu, RQ.entry[index].instr_id, RQ.entry[index].full_addr, RQ.entry[index].address, 0, RQ.entry[index].ip, RQ.entry[index].type);
 			  
 			  RQ.entry[index].data = pa >> LOG2_PAGE_SIZE; 
 			  RQ.entry[index].event_cycle = current_core_cycle[read_cpu];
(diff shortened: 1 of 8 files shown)


itisntalex:
Hello @JamesPC44,

As I just saw this pull request and I'm the mad dog about consistency I will place some comments here:
- First, I see that you have a couple of hard-coded values/constants in your code. In my views, that's not a good practice as it is easy to forget to update one when one will apply further changes. I would recommend using either old-school `#defines` or `constexpr` and computing the values you need at compile-time. An example of this is the flag `0x1ff` you use multiple times in `main.cc`.
- I did see some inconsistency in terms of the types used in function declarations. You use both build-in types and system agnostic types provided by `#include <cstdint>` or `#include <stdint.h>` header. For consistency, I would suggest to choose one and stick to it.
- One last thing would be about choosing between C++ I/O streams or the old-school interface provided by the C standard library. As a matter of fact, there's now a blend of `cout` and `printf` in the code and I believe we would do better sticking with one.

I might also place comments directly on the code. 😉

src/main.cc
```diff
@@ -177,7 +178,28 @@ void finish_warmup()
...
+// Runtime PAGE_TABLE_LATENCY
+// if environment var is defined, set to that value
+char* s = getenv("CS_PAGE_TABLE_LATENCY");
+if(!s) {
     PAGE_TABLE_LATENCY = 100;
+	printf("Page Table latency set to default of %d\n", PAGE_TABLE_LATENCY);
+}
+else {
+	sscanf(s, "%d", &PAGE_TABLE_LATENCY);
+	printf("Page Table latency set to %d\n", PAGE_TABLE_LATENCY);
```
itisntalex, line 190: Here's an example of blending C++ I/O and C-style I/O. You can do exactly the same with `cin`, `cout` and `clog`. More details [here](https://en.cppreference.com/w/cpp/io).

It would basically look like this:
```suggestion
        std::cin >> PAGE_TABLE_LATENCY;
	std::cout << "Page Table latency set to default " << PAGE_TABLE_LATENCY << std::endl;
```

src/main.cc
```diff
@@ -296,8 +318,653 @@ uint64_t rotr64 (uint64_t n, unsigned int c)
...
+{georgios.vavouliotis, lluc.alvarez, marc.casas}@bsc.es
+
+Ported by James Coman - james in the domain tamu.edu
+*/
+int simulated_page_walker(uint32_t cpu, uint64_t vpage, int swap, uint8_t is_code, uint64_t instr_id, uint64_t ip, int type){
+
+	// 4KB: LOG2_PAGE_SIZE == 12
+	// 2MB: LOG2_PAGE_SIZE == 21
+
+	int cstall = 2; // the latency cost of the current page walk
```
itisntalex, line 334: Here, why not using the same types as in the method parameters? That would add consistency to the code.

```suggestion
int32_t simulated_page_walker(uint32_t cpu, uint64_t vpage, int swap, uint8_t is_code, uint64_t instr_id, uint64_t ip, int type){

	// 4KB: LOG2_PAGE_SIZE == 12
	// 2MB: LOG2_PAGE_SIZE == 21

	int32_t cstall = 2; // the latency cost of the current page walk
```

src/main.cc
```diff
@@ -296,8 +318,653 @@ uint64_t rotr64 (uint64_t n, unsigned int c)
...
+    	if(LOG2_PAGE_SIZE == 12)
+        	mmu_hit[2] = ooo_cpu[cpu].STLB.search_pd(compare_pd); // check for hits in the MMU cache of the PD level
+
+    	// update the MMU cache levels
+        ooo_cpu[cpu].STLB.update_pml4(ooo_cpu[cpu].STLB.mmu_timer, compare_pml4); // update the content of the MMU cache of the PML4 level if needed
+        ooo_cpu[cpu].STLB.update_pdp(ooo_cpu[cpu].STLB.mmu_timer, compare_pdp);   // update the content of the MMU cache of the PDP level if needed
+        if(LOG2_PAGE_SIZE == 12)
+        	ooo_cpu[cpu].STLB.update_pd(ooo_cpu[cpu].STLB.mmu_timer, compare_pd); // update the content of the MMU cache of the PD level if needed
+
+		uint64_t cr3 = 0x200000; // defining the CR3 register pointing address
```
itisntalex, line 369: Here you define a variable that is in fact a constant. Instead, you could use a `#define` or a `constexpr`.

src/main.cc
```diff
@@ -296,8 +318,653 @@ uint64_t rotr64 (uint64_t n, unsigned int c)
...
+            pt_index   = (vpage       & 0x00000000001ff);
+            pd_index   = ((vpage>>9)  & 0x00000000001ff);
+            pdp_index  = ((vpage>>18) & 0x00000000001ff);
+            pml4_index = ((vpage>>27) & 0x00000000001ff);
+        }
+        else{
+            pt_index   = (vpage       & 0x00000000001ff);
+            pdp_index  = ((vpage>>9)  & 0x00000000001ff);
+            pml4_index = ((vpage>>18) & 0x00000000001ff);
+        }
```
itisntalex, line 383: The same flag is used multiple times here. So I would suggest the same as in the previous comment. Set it in a `#define` and that would look much cleaner. 😉

src/main.cc
```diff
@@ -296,8 +318,653 @@ uint64_t rotr64 (uint64_t n, unsigned int c)
...
+            pdp2s  = cr3 + 512 * 8 \
+                     + pml4_index * 512 * 8 \
+                     + pdp_index * 8;
+
+            pt2s   = cr3 + 512 * 8 \
+                   	 + 512 * 512 * 8 \
+                   	 + pml4_index * 512 * 512 * 8 \
+                   	 + pdp_index * 512 * 8 \
+                   	 + pt_index * 8;
+        }
```
itisntalex, line 421: Maintainance here might be difficult in the future. Consider putting some kind of `#defines`, etc. 😉

src/main.cc
ngober: Careful! Your line 189 does not keep the same functionality.

```suggestion
	PAGE_TABLE_LATENCY = std::atoi(s);
	std::cout << "Page Table latency set to " << PAGE_TABLE_LATENCY << std::endl;
```

src/cache.cc
```diff
@@ -2,6 +2,11 @@
 #include "set.h"
 
 uint64_t l2pf_access = 0;
+/* you can implement each page walk cache level as a 2d array */
+uint64_t pml4[PML4_SET][PML4_WAY], pdp[PDP_SET][PDP_WAY], pd[PD_SET][PD_WAY];
+
+/* implementation of LRU policy with counters */
+uint64_t pml4_lru[PML4_SET][PML4_WAY], pdp_lru[PDP_SET][PDP_WAY], pd_lru[PD_SET][PD_WAY];
```
ngober, line 9: My preferred style is to create structs for each table entry to keep the data grouped together.
```suggestion
/* you can implement each page walk cache level as a 2d array */
struct pt_entry
{
    uint64_t address;
    uint64_t lru;
};
pt_entry pml4[PML4_SET][PML4_WAY], pdp[PDP_SET][PDP_WAY], pd[PD_SET][PD_WAY];
```

This makes debugging easier (data is logically grouped with its metadata) and in my opinion more accurately represents the way we think about this data.

src/main.cc
ngober: I think an argument can be made here that the width of the integer does not matter because this variable is not microarchitecture. So, we can simply let the compiler choose the width. I think we generally underuse `int`, `bool`, etc in ChampSim in situations where it would communicate better.

src/cache.cc
```diff
@@ -675,7 +680,7 @@ void CACHE::handle_read()
 			  // TODO: need to differentiate page table walk and actual swap
```
ngober, line 680: Would you please delete this TODO? You are resolving it with this PR.

src/main.cc
ngober: I agree. It took me a few minutes to figure out where all of the magic numbers came from.

src/main.cc
```diff
@@ -296,8 +318,653 @@ uint64_t rotr64 (uint64_t n, unsigned int c)
...
+                        cstall += 200;
+                    }
+                }
+            }
+        }
+    }
+
+    return cstall;
+}
+
```
ngober, line 965: This seems to duplicate a lot of code that is in cache.cc. Is it wrong here to add the miss entries to the L1 read queues and allow the cache architecture to propogate the request?

My primary concern here is that all of these fills happen instantaneously, which is unrealistic. The LLC fill should take some time, and we should reflect that. However, all of these requests need to be filled _as a part of the STLB miss_. This is a difficult problem to solve. Could we set the STLB MSHR event_cycle to some absurdly large value, ensuring the core will wait on it until the requests from the page walk are resolved?

This section of code is the most important part of this contribution. We should be careful to get this right.

sethpugsley:
First of all, this pull request is to merge into master, which is not right for this kind of change.  I'm not even sure it's appropriate to be merged into develop in its current state.

ChampSim definitely needs a better virtual memory system, including realistic TLB miss penalties.  This change starts with the already-not-great existing virtual memory system and makes it even more complicated and messy, while only slightly increasing the realism in the process.

Seeing the code in this pull request makes me realize we need an even greater, more drastic and fundamental change to the way the virtual memory system is modeled in ChampSim.  I will start a discussion in a new Issue thread about this to explain what I have in mind, and you can judge for yourself how it compares to this proposal.

src/main.cc
itisntalex: Yes, you’re right I read too fast and took it for a scanf.

src/cache.cc
itisntalex: I agree. Much cleaner in my view.

PR #51, opened 2020-06-28
MSHR merging:- Keep prior fill information

vishalgupta97:
The bug:- If an instruction prefetch request from L1I prefetcher is present in L2C and demand request comes from L1D for the same cache block, then fill_l1i information will be lost. If a subsequent instruction demand request comes to same cache block, it will get merged in L1I MSHR and it will wait forever. 
This will also not create a deadlock situation because deadlock is checked once the request reaches ROB.

### DIFF
--- a/src/cache.cc
+++ b/src/cache.cc
@@ -763,7 +763,15 @@ void CACHE::handle_read()
                         if (MSHR.entry[mshr_index].type == PREFETCH) {
                             uint8_t  prior_returned = MSHR.entry[mshr_index].returned;
                             uint64_t prior_event_cycle = MSHR.entry[mshr_index].event_cycle;
-                            MSHR.entry[mshr_index] = RQ.entry[index];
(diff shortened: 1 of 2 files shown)


ngober:
Are you confident that this is a bug? Do you see situations where the same block is requested for both instruction and data?

vishalgupta97:
Yes. In traces which I am using has this bug. It contains both user and kernel level traces. The bug appeared for address in kernel address space.
One example that I can think of in kernel address space is when registering an interrupt handler by the kernel. Some code is written to IDT and then it is executed if some interrupt comes. But that will happen during the boot time, not when some application is running. 
I am not sure what type of program in kernel will be accessing the same block both in instruction and data. Might be some self modifying code.

ngober:
This looks to be an extension of #20, patching something it missed, so I would like to merge it. Can you please change the merge branch to `develop`?

PR #52, opened 2020-07-01
Virtual Address Space data prefetching

sethpugsley:
Added basic virtual address space data prefetching, along with example prefetchers.  This builds on the existing interface for data prefetchers, giving them the virtual address as well as the physical address.  A cache line can now be prefetched using its virtual address and the new function va_prefetch_line().  This allows you to prefetch outside of page boundaries, and gives visibility into which physical pages are contiguous in the virtual address space.

For this first version, VA->PA translation happens magically after a fixed number of cycles.

### DIFF
--- a/src/cache.cc
+++ b/src/cache.cc
@@ -127,18 +127,21 @@ void CACHE::handle_fill()
         if (do_fill){
             // update prefetcher
 	  if (cache_type == IS_L1I)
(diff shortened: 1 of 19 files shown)


src/cache.cc
```diff
@@ -1464,6 +1485,119 @@ int CACHE::kpc_prefetch_line(uint64_t base_addr, uint64_t pf_addr, int pf_fill_l
...
+	      vapq_index = 0;
+	    }
+	}
+    }
+
+  // TEMPORARY SOLUTION: mark prefetches as translated after a fixed latency
+  vapq_index = VAPQ.head;
+  for(uint32_t i=0; i<VAPQ.SIZE; i++)
+    {
+      if((VAPQ.entry[vapq_index].address == VAPQ.entry[vapq_index].v_address) && (VAPQ.entry[vapq_index].event_cycle <= current_core_cycle[cpu]))
```
ngober, line 1571: I think this could be a dangerous condition. There is an (unlikely) chance that the physical address and the virtual address could be identical. Could we instead include a `bool va_translated` flag to the `PACKET` class?

PR #58, opened 2020-08-17
Unified ready-to-execute queues; fixed bug where instrs w/ loads executed for free

sethpugsley:
1) Merged RTE0 and RTE1 into a single buffer, called ready_to_execute.  NOTE: this new buffer is still just a simple FIFO, not a priority queue (yet ...).

2) Before, all instrs w/ loads would bypass the RTE queues, and would just get marked as executed when their loads resolved, ignoring the limits of the core's EXEC_WIDTH.  Now, all instructions must respect the EXEC_WIDTH before they can pass through the complete_execution() function.

### DIFF
--- a/inc/ooo_cpu.h
+++ b/inc/ooo_cpu.h
@@ -15,7 +15,7 @@ using namespace std;
 // CORE PROCESSOR
 #define FETCH_WIDTH 6
 #define DECODE_WIDTH 6
-#define EXEC_WIDTH 6
+#define EXEC_WIDTH 4
 #define LQ_WIDTH 2
 #define SQ_WIDTH 2
 #define RETIRE_WIDTH 5
@@ -62,8 +62,7 @@ class O3_CPU {
     uint64_t STA[STA_SIZE], STA_head, STA_tail; 
 
     // Ready-To-Execute
-    uint32_t RTE0[ROB_SIZE], RTE0_head, RTE0_tail, 
-             RTE1[ROB_SIZE], RTE1_head, RTE1_tail;  
+    uint32_t ready_to_execute[ROB_SIZE], ready_to_execute_head, ready_to_execute_tail;
 
     // Ready-To-Load
     uint32_t RTL0[LQ_SIZE], RTL0_head, RTL0_tail, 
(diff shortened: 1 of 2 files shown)


inc/ooo_cpu.h
```diff
@@ -15,7 +15,7 @@ using namespace std;
 // CORE PROCESSOR
 #define FETCH_WIDTH 6
 #define DECODE_WIDTH 6
-#define EXEC_WIDTH 6
+#define EXEC_WIDTH 4
```
ngober, line 18: Why are we decreasing execution width here? This would create a bottleneck.

src/ooo_cpu.cc
```diff
@@ -1786,19 +1767,19 @@ void O3_CPU::reg_RAW_release(uint32_t rob_index)
...
-                        cout << " head: " << RTE0_head << " tail: " << RTE0_tail << endl; }); 
+                        cout << "[ready_to_execute] " << __func__ << " instr_id: " << ROB.entry[i].instr_id << " rob_index: " << i << " is added to ready_to_execute";
+                        cout << " head: " << ready_to_execute_head << " tail: " << ready_to_execute_tail << endl; }); 
 
-                        RTE0_tail++;
-                        if (RTE0_tail == ROB_SIZE)
-                            RTE0_tail = 0;
+                        ready_to_execute_tail++;
+                        if (ready_to_execute_tail == ROB_SIZE)
+                            ready_to_execute_tail = 0;
```
ngober, line 1782: Could you check that this does not bypass the scheduler width? This section is very similar to L916-924.

inc/ooo_cpu.h
sethpugsley: The idea here is that the default config approximately matches an Intel Sunny Cove core, which has 4 non-AGU ALUs, hence the 4-wide execution.  I had previously set it to 6 in an attempt to compensate for how I believed the simulator worked, which was wrong.  I believed that instructions with loads would have to go through execute twice,- once when their registers were ready, and again when their loads were ready, but really they were going through execute zero (0) times.  After this fix, all instructions now respect the EXEC_WIDTH limits at the complete_execution() phase, so making the EXEC_WIDTH 4 now matches the original intention.

src/ooo_cpu.cc
sethpugsley: I agree.  I'll see if I can make it so that we only add to ready_to_execute[] in do_scheduling().  reg_RAW_release() should only mark the dependencies as resolved, not also do the scheduling.

ngober: If you think those changes should be in a separate PR, then go ahead and merge this. I like it overall.

sethpugsley: I was able to make it so that instructions that have had their dependencies resolved need to go through scheduling again, but now that I've thought about this more, I think I like this original behavior better.  Allowing instructions to execute as soon as possible after their dependencies are resolved is kind of like a bypass network.  And since they don't have have their executed field set to anything, they still count as being in the scheduler, so they are still taking up scheduler width anyway.

I like it better this way, so I think this should be merged as-is, but I think we should eventually replace the circular buffer with something like a priority queue to make the oldest instructions execute first.

PR #59, opened 2020-08-19
Added L0 instruction buffer

ngober:
Programs, especially those with small instruction footprints, tend to access the same instruction cache line many times before moving on to another line. While the L1I does hold the cache line, it has a hit latency of 4 cycles. Once a demand returns from the L1I, all instructions in the ROB from that cache line are marked as fetched. However, this excludes instructions that enter the ROB just after their line returns. For these instructions, we must create a new demand.

This patch introduces an instruction buffer that keeps the most recent instruction block and marks an instruction in this block as immediately fetched if it is read from the trace. As submitted, this patch has only one entry, but is configurable in the number of sets and ways, managed in an LRU fashion.

My tests suggest that this reduces the number of L1I cache hits by 35% while not changing the number of L1I cache misses. This can permit better instruction prefetchers to be developed because there is less noise from repeated requests to the MRU cache line.

### DIFF
--- a/inc/instruction.h
+++ b/inc/instruction.h
@@ -115,8 +115,6 @@ class ooo_model_instr {
             branch_taken,
             branch_mispredicted,
             branch_prediction_made,
-            translated,
-            data_translated,
             source_added[NUM_INSTR_SOURCES],
             destination_added[NUM_INSTR_DESTINATIONS_SPARC],
             is_producer,
@@ -130,11 +128,12 @@ class ooo_model_instr {
     uint8_t branch_type;
     uint64_t branch_target;
 
-    uint32_t fetched, scheduled;
+    uint8_t translated = 0,
+            fetched = 0,
+            decoded = 0,
+            scheduled = 0,
+            executed = 0;
     int num_reg_ops, num_mem_ops, num_reg_dependent;
-
-    // executed bit is set after all dependencies are eliminated and this instr is chosen on a cycle, according to EXEC_WIDTH
-    int executed;
 
     uint8_t destination_registers[NUM_INSTR_DESTINATIONS_SPARC]; // output registers
 
@@ -185,14 +184,9 @@ class ooo_model_instr {
         branch_taken = 0;
         branch_mispredicted = 0;
 	branch_prediction_made = 0;
-        translated = 0;
-        data_translated = 0;
         is_producer = 0;
         is_consumer = 0;
         reg_RAW_producer = 0;
-        fetched = 0;
-        scheduled = 0;
-        executed = 0;
         reg_ready = 0;
         mem_ready = 0;
         asid[0] = UINT8_MAX;
--- a/inc/ooo_cpu.h
+++ b/inc/ooo_cpu.h
@@ -1,6 +1,8 @@
 #ifndef OOO_CPU_H
 #define OOO_CPU_H
 
+#include <array>
+
 #include "cache.h"
 
 #ifdef CRC2_COMPILE
@@ -24,6 +26,10 @@ using namespace std;
 //#define SCHEDULING_LATENCY 0
 //#define EXEC_LATENCY 0
 //#define DECODE_LATENCY 2
+
+// Dimensions of instruction buffer
+#define L0I_SET 1
+#define L0I_WAY 1
 
 #define STA_SIZE (ROB_SIZE*NUM_INSTR_DESTINATIONS_SPARC)
 
@@ -51,6 +57,17 @@ class O3_CPU {
              next_print_instruction, num_retired;
     uint32_t inflight_reg_executions, inflight_mem_executions, num_searched;
     uint32_t next_ITLB_fetch;
+
+    struct l0i_entry_t
+    {
+        bool valid = false;
+        unsigned lru = L0I_WAY;
+        uint64_t addr = 0;
+    };
+
+    // instruction buffer
+    using l0i_t= std::array<std::array<l0i_entry_t, L0I_WAY>, L0I_SET>;
+    l0i_t L0I;
 
     // reorder buffer, load/store queue, register file
     CORE_BUFFER IFETCH_BUFFER{"IFETCH_BUFFER", FETCH_WIDTH*2};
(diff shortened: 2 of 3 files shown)


sethpugsley:
Does this impact performance at all?  How much?

This seems a bit like a half-measure on our way to implementing a decoded micro op cache.  Hits in the uop cache would skip L1-I lookup as well as the decode latency penalty.

As far as implementation goes, I imagine that this would be implemented as hits being marked as translated and fetched as soon as they enter the IFETCH_BUFFER, and then the instructions would be marked as decoded as soon as they enter the DECODE_BUFFER.  Instructions coming out of the DECODE_BUFFER would be added/have their LRU state updated in the uop cache.

In the early days of getting ChampSim ready for IPC-1 I had a uop cache in there, but I ended up removing it because we wanted the instruction prefetchers to have a more complete view of all instructions the program was executing.  It's hard to predict what will happen to the instruction prefetchers in the presence of a uop cache.  Will the best ones still be the best?  Will they require extensive modification to correctly function?  Who knows?  If we eliminate almost all of the L1-I accesses, then most of the information the prefetchers could get would come from the branches, but most of the IPC-1 submissions focused on L1-I cache accesses and ignored the branches.

Does anyone else have any thoughts about this?

ngober:
The performance improves the DPC3 traces by a geomean 2.5%. All of this gain appears to come from time not spent hitting the L1.

I'm not convinced that a uop cache is feasible under the current setup. Do we have enough information in the trace to do full decoding?

sethpugsley:
The decoding is all pretend, as is almost everything else in ChampSim.  The decode and instruction fetch stages were added to ChampSim when I was getting it ready for IPC-1.

Right now, after an instruction is finished being fetched, it's added to a decode queue (called the DECODE_BUFFER), and after DECODE_LATENCY amount of time, they move out of the DECODE_BUFFER and into the ROB.  Right now the DECODE_LATENCY is set to 2 after warmup is complete.  No decoding actually happens to instructions, and the DECODE_BUFFER is just someplace they have to sit for X cycles to slow down the pipeline.

ngober:
We may be getting off topic here, but it seems to me that in order to do proper uops, you would need to know the opcodes. I've read "Improving CISC Instruction Decoding Performance Using a Fill Unit" by Smotherman and Franklin, and it seems to suggest that we could infer uops from just the addresses and registers that we currently have available.

Are you suggesting that I should mark hits to my L0 as decoded in addition to fetched?

sethpugsley:
I think I understand the source of confusion.  Yes, "decoded uop cache" is a misnomer because there are no uops in ChampSim.  Instead, I'm just saying that we could have a structure that performs a similar function to that in the ChampSim, i.e., allowing instructions that are hits in this new structure to rush through the instruction fetch and decode stages without having to actually look up the L1-I cache or pay the decode penalty.

It's possible that you extend the L0 instruction cache to allow instructions to also skip decode, but if you do that, then you should rename it to "decoded instruction cache."

I think that this feature should be one that we control with a knob, so that IPC-1 prefetchers can still operate by looking up the L1-I cache for all instructions if we want them to.

ngober:
I can hide this behind an ifdef if we want. One thing of note is that ChampSim does read merging, that is, if a demand comes into the RQ for a cache line that is already in the RQ, the second will be discarded. So, there is not precisely one L1I lookup per instruction. This change was intended to cover the case where an instruction is read from the trace just barely too late to be covered by an earlier demand (which apparently happens 35% of the time) and so causes an additional hit to the MRU cache line.

What are the dimensions that you envision on this decoded instruction cache?

sethpugsley:
Hiding behind an ifdef is a good idea for now.  When we eventually move to a config file-based setup, then it should be a knob in there.

Here are some public sources about decoded instruction caches in Intel CPUs.  The first tells a little bit about how it works (under the section "New uop cache & x86 tax"), the second gives some more details, and the last gives updated sizes for a Sunny Cove core.  We can base the dimensions off these:
https://en.wikichip.org/wiki/intel/microarchitectures/sandy_bridge_(client)
https://en.wikichip.org/wiki/intel/microarchitectures/skylake_(client)
https://en.wikichip.org/wiki/intel/microarchitectures/sunny_cove

I encourage you to read those, but according to my understanding, the way I believe it should be implemented is that our decoded instruction cache should be either 32 sets x 12 ways, or 64 sets x 6 ways.

Each entry in this cache will represent FETCH_WIDTH worth of instructions, meaning everything that could be read from the trace file in a single cycle.  So you only ever look up the decoded instruction cache with the head of the IFETCH_BUFFER at the beginning of the cycle, and if it's a hit, then everything in the IFETCH_BUFFER gets fetched+decoded "for free," and if the head of the IFETCH_BUFFER misses, then everything in the IFETCH_BUFFER has to go through the L1-I and pay the decode latency.  This is very different from it behaving like an L0, which has to respect the cache line boundaries in the virtual address space.

This is starting to turn into a bigger project, because I think to do this right we will have to make sure that multiple "fetch windows" aren't in the IFETCH_BUFFER at the same time, so that the scheme I outlined above can work.

Alternatively, we could make it operate more like an L0 cache, which is organized according to the virtual address space, and where you check each individual instruction to see if it's present in the L0 cache, and those that aren't have to be fetched and decoded.  This would be much easier to code up, but I don't believe represents as well what's done in real processors.

If we go for the easier-to-implement L0 style, then I think we should cache 72 64B cache lines, and the dimensions should be 8 sets and 9 ways.  This gives us a similar number of instructions cached to Sunny Cove's 2.25k uop cache, assuming instructions are on average ~2 bytes.  If you want to assume x86 instructions are closer to 2.5 bytes on average, then we'd need 96 cache lines, and could use 16 sets and 6 ways.

ngober:
I refactored the L0I into a Decoded Instruction Buffer, as we discussed, and caused hits to it to skip decode. As expected, this improves the core's performance across the board. I also took the time while I was doing surgery to clean up some messy code, so it should be easier to maintain now.

sethpugsley:
I looked over the code and tested a bunch of traces, and everything looks good to me.

PR #61, opened 2020-08-27
Cache write bandwidth

sethpugsley:
Cache write bandwidth can now be something other than 1 cache line per cycle.  It also makes it so that cache fill and write back operations both share the same limited resource, rather than each having their own dedicated, hard-coded 1/cycle.

In each cache, you can set MAX_WRITE, which does the same thing for fills and write backs that MAX_READ does for reads and prefetches.

Also, added a feature that automatically quits at the end of the trace for single core simulations, rather than wrapping back around to the beginning of the trace.  Multi-core simulations should be unaffected.

### DIFF
--- a/inc/cache.h
+++ b/inc/cache.h
@@ -89,8 +89,8 @@ class CACHE : public MEMORY {
     uint32_t LATENCY;
     BLOCK **block;
     int fill_level;
-    uint32_t MAX_READ, MAX_FILL;
-    uint32_t reads_available_this_cycle;
+    uint32_t MAX_READ, MAX_WRITE;
+    uint32_t reads_available_this_cycle, writes_available_this_cycle;
     uint8_t cache_type;
 
     // prefetch stats
@@ -152,8 +152,8 @@ class CACHE : public MEMORY {
         lower_level = NULL;
         extra_interface = NULL;
         fill_level = -1;
-        MAX_READ = 1;
-        MAX_FILL = 1;
+        MAX_READ = MAX_READ_PER_CYCLE;
+        MAX_WRITE = MAX_WRITE_PER_CYCLE;
 
         pf_requested = 0;
         pf_issued = 0;
--- a/src/cache.cc
+++ b/src/cache.cc
@@ -5,6 +5,8 @@ uint64_t l2pf_access = 0;
 
 void CACHE::handle_fill()
 {
+  for(uint32_t i=0; i<MAX_WRITE; i++) {
+
     // handle fill
     uint32_t fill_cpu = (MSHR.next_fill_index == MSHR_SIZE) ? NUM_CPUS : MSHR.entry[MSHR.next_fill_index].cpu;
     if (fill_cpu == NUM_CPUS)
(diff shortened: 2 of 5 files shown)


sethpugsley:
I will say that the majority of the lines of change in cache.cc are due to a change in indentation after I added a top level for-loop to the handle_fill() and handle_writeback() functions, so try not to get distracted by those.

src/cache.cc
```diff
@@ -1060,8 +1094,16 @@ void CACHE::handle_prefetch()
 
 void CACHE::operate()
 {
+    // perform all writes
+    writes_available_this_cycle = MAX_WRITE;
     handle_fill();
-    handle_writeback();
+
+    if(WQ.occupancy && (writes_available_this_cycle > 0))
```
ngober, line 1101: It's probably good to be explicit here and write `WQ.occupancy > 0`

src/ooo_cpu.cc
```diff
@@ -30,6 +30,24 @@ void O3_CPU::read_from_trace()
 
         if (knob_cloudsuite) {
             if (!fread(&current_cloudsuite_instr, instr_size, 1, trace_file)) {
+	      if((SINGLE_CORE_TRACE_END_EXIT==1) && (NUM_CPUS==1))
```
ngober, line 33: I don't think I like the way you've implemented this here. It feels very ham-handed to have this as a preprocessor define. I feel like there are multiple ways you can handle terminating on trace end: terminate when the first trace ends, when the last trace ends, or never. Could we change this to a CLI option like `--trace-end=(first|last|loop)`?

ngober, line 33: Also, does this belong in its own PR? I feel like you have two different changes here and it is difficult to comment on both of them.

src/cache.cc
```diff
@@ -74,162 +76,178 @@ void CACHE::handle_fill()
...
-        }
+	      writes_available_this_cycle--;
+
+	      update_fill_cycle();
+	    }
+	  }
     }
+    else
+      {
+	return;
```
ngober, line 237: I don't like the `else-return` idiom here. I think it would be more clear to invert the condition on the `if` and have that return at the top of the function. A similar thing goes for the `if` statement that follows.

inc/cache.h
```diff
@@ -152,8 +152,8 @@ class CACHE : public MEMORY {
         lower_level = NULL;
         extra_interface = NULL;
         fill_level = -1;
-        MAX_READ = 1;
-        MAX_FILL = 1;
+        MAX_READ = MAX_READ_PER_CYCLE;
+        MAX_WRITE = MAX_WRITE_PER_CYCLE;
```
ngober, line 156: Why do we initialize here when we redefine it later in `src/main.cc`?

sethpugsley:
I removed the end-of-trace simulation completion feature.  Now this PR has just the changes to cache write bandwidth, which is implemented using the same code patterns that cache read bandwidth's implementation uses.

src/ooo_cpu.cc
GinoAC: I think it'll do until we move more into reworking how champsim is configured at run time.

GinoAC:
Based on SPEC17 results, this change does not seem to significantly change performance over the current develop branch. Aside from the changes suggested by @ngober, it looks solid to me.

PR #62, opened 2020-08-30
Prefetch stats

GinoAC:
A simple change to main.cc to reset the prefetch statistics after warm up. Prior to this, the number of useful/useless prefetches could exceed the number of actual prefetch requests recorded during the simulation period.


sethpugsley:
Can this be changed to merge into develop instead?

GinoAC:
It should be, its an error on my part that it's listed on master.

PR #64, opened 2020-09-03
Separated trace reading from core operations

ngober:
This patch pulls the file reading tasks out of the `O3_CPU` class. This will allow us greater flexibility to create future trace file formats without changing any operations in the core model. Furthermore, it reduces a large amount of code duplication between the current traces and the old cloudsuite trace format.

This patch causes no change in the performance of the core model based on my tests of the dpc_max_simpoint traces run for 100M/200M with no prefetching.


GinoAC (approved):
Looks solid and doesn't break anything or seem to change performance.

sethpugsley (approved):
Looks good, and I confirmed that performance is unchanged.

PR #69, opened 2020-09-16
JSON-based configuration and build system

### LINKED ISSUE #31: Choosing a build system
I think I'm not alone in thinking that the current method of building ChampSim is clunky. Over the past year, there have been three pull requests opened to address issues (#10, #11, #27). The three of them are incompatible, but they all attempt to address the same issues: The current build system
- Does not allow for models that span multiple files. This leads to monolithic, non-reusable code.
- Does not permit compilation flags or parameters (such as CXXFLAGS) to be passed to the compiler.
- Requires that models be named `*.l2c_pref`, or similar. This causes problems with tools that would expect `*.cc` or `*.cpp`.
- Does not scale well with the number of options available. With the recent addition of L1I prefetchers, `build_champsim.sh` now takes 7 parameters, while it is likely that only one is important on any build.
- Rebuilds the entire project on every build. Why do we need to do this when we have tools like Make available to us?

I think that continuing to use Make is a smart idea. It is a standard and it is well-understood by developers. There are multiple ways to prepare the project for Make:
- Autoconf is the GNU standard for building projects, but is targeted more to enable compatibility, not to select between source files
- CMake is a possible alternative, but introduces a new dependency (#11)
- SCons is used by GEM5, and seems to be a powerful tool.
- A home-cooked solution may also be good, since it varies only marginally from what we have now. (#10 uses a Python script to parse a configuration file, #27 uses environment variables)

What features do we need out of our build system? What can we do to make it better?

ngober:
Yet another build system. Resolves #31.

This one takes inspiration from all of the others #10, #11, #27 to produce one that I'm very happy with, due to its extensibility.

The config.sh script generates inc/champsim_constants.h, which contains all of the machine parameters previously scattered across various headers, src/core_inst.cc, which contains instantiations of the elements of the machine we're simulating (again, previously scattered across various source files), and a makefile.

The input to the configuration file is a JSON object and I have included an example file that specifies all available options. But, and this is the most important part, all items in this configuration are optional. If not specified, the system will assume defaults that match what we currently specify. 

I want to eventually expand this configuration. There are currently no options for prefetching in the TLBs, nor replacement anywhere but the L3, but these will be easy additions to the build configuration if we add them later. I also have a vision of future possibilities of heterogeneous cores, which this build system almost already supports.

I don't want to overload us with options, but I think this is the best solution we have yet.


sethpugsley:
I like this, but I think it's unwise to rush into such a large change without input from more people.  I hope people other than just me and Gino can try it out before we adopt this strategy.

A few things:

I noticed that you have #define DEBUG_PRINT turned on by default in this version, so that makes it very surprising to run for the first time.

Do I need to do a make clean & make every time I make a code change?  I experimented a little bit by changing some things here and there and the only way I could reliably get it to compile again was by doing a make clean & make.

I believe I'm seeing that the reported IPC is different than the current version of the develop branch for what I believe is an identical configuration.  Is anyone else seeing that, or does it have an explanation?  Are my configurations not as identical as I believe?  The configuration I made doesn't change any of the numbers in champsim_config.json, but it does use hashed_perceptron for the branch predictor, and uses next_line_l1d for the L1D prefetcher, and spp_dev for the L2C prefetcher.  Can anyone else confirm that this produces different IPC?

ngober:
The #define DEBUG_PRINT is definitely a mistake, I'll remove that.

I've seen the same thing with the make clean and remake. I'm not entirely sure why that is, but I have an idea to fix it by only rebuilding part of the makefile. I'll try a couple things.

I would be surprised if the IPC is different, but the most likely case is that I improperly copied values over to my defaults. I'll double check that also.

ngober:
Re: involving more people, is this the kind of thing that we should notify the mailing list about? The reason I ask is that I have some changes in mind to the way we integrate the prefetchers that will involve at least a name change in the prefetcher files themselves. If we need to notify the mailing list, would it be better to do both of these in one email?

ngober:
I've added a commit that addresses the `make clean` issue. It now builds the prefetchers, branch predictor, and replacement policies as shared libraries, then links them. The configure script now detects when the configuration has changed and forces a rebuild if we select a new prefetcher. It was the configuration change that was causing the binary incompatibilities between the builds.

ngober:
I found the source of the IPC changes. My default associativity for the LLC was incorrect. Changing that appears to have fixed the problem. Let me know if you see any discrepancies on your end.

GinoAC (approved):
Updated build system that is intuitive for new users and provides easier access to advanced configuration options for experienced architects. Able to successfully build configurations that ship with base download.

PR #70, opened 2020-09-16
Removed magic in instruction VA translation

ngober:
As I was making other changes, I noticed that the ITLB was receiving no requests, because instructions were entering the IFETCH_BUFFER magically translated. This change removes that magic in favor of actually using the TLB.


sethpugsley:
I put in the magical translation for IPC-1, but forgot to remove it afterwards.  There is something similar going on for VA prefetches, which also don't touch the DTLB, but that's a fix for another day.

Issue #68, opened 2020-09-16
Possible bug in MSHR match


mahmoodn:
Hi
Part of the [handle_read()](https://github.com/ChampSim/ChampSim/blob/master/src/cache.cc#L638) for checking a miss block in MSHR seems to be buggy. The code is
```
                // check mshr
                uint8_t miss_handled = 1;
                int mshr_index = check_mshr(&RQ.entry[index]);

		if(mshr_index == -2)
		  {
		    // this is a data/instruction collision in the MSHR, so we have to wait before we can allocate this miss
		    miss_handled = 0;
		  }
```
So, if we find a match the `mshr_index` must be -2. That means, upon a match, [check_mshr()](https://github.com/ChampSim/ChampSim/blob/master/src/cache.cc#L1654) must return -2. Now look at part of the code:

```
  for (uint32_t index=0; index<MSHR_SIZE; index++)
    {
      if (MSHR.entry[index].address == packet->address)
	{
	      return index;
	}
    }
```
So, that function, returns index upon a match and not -2. 
In fact [return -2](https://github.com/ChampSim/ChampSim/blob/master/src/cache.cc#L1681) is commented in that function.

That can be verified easily by debugging. I see that there is a match in the MSHR and the function returns 0 (first entry in MSHR) but that is considered as collision.

Any comment on that? If you agree with that, I will create a poll request.

ngober:
I don't think that's what returning -2 means in this context. The -2 return was introduced during preparations for IPC1, to resolve instruction-data collisions by waiting for one to return before demanding the other. I'm pretty sure this was resolved before IPC1, then trimmed up in #51, where we just return both instead of one or the other. When the `check_mshr()` function returns nonnegative, it indicates a match in the given MSHR index.

PR #72, opened 2020-09-17
Fixing mshr_index for collision

mahmoodn:
Please refer to https://github.com/ChampSim/ChampSim/issues/68

### DIFF
--- a/src/cache.cc
+++ b/src/cache.cc
@@ -306,7 +306,7 @@ void CACHE::handle_writeback()
                 uint8_t miss_handled = 1;
                 int mshr_index = check_mshr(&WQ.entry[index]);
 
-		if(mshr_index == -2)
+		if(mshr_index >= 0)
 		  {
 		    // this is a data/instruction collision in the MSHR, so we have to wait before we can allocate this miss
 		    miss_handled = 0;
@@ -639,7 +639,7 @@ void CACHE::handle_read()
                 uint8_t miss_handled = 1;
                 int mshr_index = check_mshr(&RQ.entry[index]);
 
-		if(mshr_index == -2)
+		if(mshr_index >= 0)
 		  {
 		    // this is a data/instruction collision in the MSHR, so we have to wait before we can allocate this miss
 		    miss_handled = 0;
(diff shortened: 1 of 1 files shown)


ngober:
I don't think I understand what this patch is trying to fix, and I'm actually concerned that it is, in fact, broken. Take the change on line 309 as an example, it's the same in the other places.

The if statement has five cases, in descending order of precedence:
- `check_mshr()` returned -2. This is an instruction-data collision. Previously, we would let the second request sit in the WQ until the first was cleared from the MSHR, but now we just don't worry about it. The block is sent to the appropriate cache when it is returned anyway.
- `check_mshr()` returned -1, and there is an available MSHR. Forward the request to the lower-level cache and pop it off of this queue.
- `check_mshr()` returned -1, and there is no available MSHR. Leave the request in the queue and stall.
- `check_mshr()` returned nonnegative. There is an in-flight miss, and this block will soon be in the cache. Pop off the queue.

It looks like this will convert the last case into a do-nothing case, which is not correct. It might even cause deadlocks in a worst-case.

I'm fine with a patch that removes the -2 return for instruction-data collisions, but this isn't that.

mahmoodn:
Right now, the check_mshr returns either -1 or index as the -2 is commented.
I didn't understand the second and third cases you mentioned as they both return -1.

In the current code base, after check_mshr(), the [-2](https://github.com/ChampSim/ChampSim/blob/master/src/cache.cc#L642) is checked. Then [-1](https://github.com/ChampSim/ChampSim/blob/master/src/cache.cc#L647) is checked. The [else](https://github.com/ChampSim/ChampSim/blob/master/src/cache.cc#L687) relates to neither -1 nor -2 which is actually the non-zero value. In that else block and the [line](https://github.com/ChampSim/ChampSim/blob/master/src/cache.cc#L688) after that, the index is again checked with -1 which I think never happens and I don't understand its position.
Finally, the non-zero index check is [here](https://github.com/ChampSim/ChampSim/blob/master/src/cache.cc#L694). So, it seems that I was confused with the in-flight misses in this else block. I will close the pull request.

If -2 is no longer needed, the mshr_index check must be commented too. Is that right?

PR #74, opened 2020-10-06
Removed unused functions in O3_CPU

ngober:
I found a function that is unused in O3_CPU, and its presence was confusing me. To reduce potential confusion for new users, I have removed it.


GinoAC (approved):
These changes seem reasonable to improve readability and don't appear to cause any impact to the performance of the simulator.

PR #75, opened 2020-10-20
Refactor of cache.cc

ngober:
This is part 1 of my efforts to clean up the cache model.

This patch should be performance-neutral. It is focused on reducing the number of long if-statements to make the code more readable. It removes 500 lines of code from src/cache.cc. Hopefully, this will make it easier for researchers to make modifications.

In addition, the replacement functions are now virtualized. This means that it is possible, with this change, to run DRRIP in the L1D, though there is currently no user-facing interface to do this and the source code must be edited directly. I leave that change for a following patch.

I would have virtualized the prefetchers also, but that change would require a change to the prefetcher interfaces, since they are not identical at different levels, and that's a user-facing change I'm not eager to do immediately.


GinoAC (approved):
Functionally the same while looking a lot better functioning. No changes to performance statistics from the test I ran.

sethpugsley:
I got a crash on my first test of cache_refactor.  Could someone else please give it a try?  I started with the develop branch, then applied cache_refactor on top of it.

Here are the changes I made to champsim_config.json:
branch predictor - hashed_perceptron
L1I prefetcher - the FNL_MMA prefetcher from IPC-1
L1D prefetcher - next_line_l1d
L2 prefetcher - spp_dev

I ran the trace 600.perlbench_s-570B.champsimtrace.xz from DPC3, and I got this error during warmup:
[L1D_MSHR] return_data instr_id: 0 cannot find a matching entry! full_addr: a7d778c0 address: 29f5de3 event: 632471 current: 632472
cache_test: src/cache.cc:1342: virtual void CACHE::return_data(PACKET*): Assertion `0' failed.

Let me know if you're able to reproduce the bug.  If not, I'll start over my testing from scratch and see if I made a errors on my end.

sethpugsley:
This problem is similar to what I've seen before.  I think this issue is caused by an L1-I prefetch erroneously ending up at the L1-D, and not finding a matching MSHR.  Did you test your cache_refactor code with prefetchers enabled in every cache simultaneously?

GinoAC:
I was able to reproduce the bug with the same configuration and with an L1-I and L2 prefetcher without L1D. It does appear to jump from the I-side to the D-side. I did not test with multiple prefetchers enabled.

sethpugsley:
I believe we've already solved this problem before, so you could just use the solution we already had in place.  See here:

https://github.com/ChampSim/ChampSim/blob/1a2f732ca7d924cb83c1b5bd30bfe9914e05da2a/src/cache.cc#L58

I believe that the packet members fill_l1i and fill_l1d are already being set correctly everywhere in your code.  You would just need to use the "if(fill_level == FILL_L2) ..." code everywhere where you're doing a return_data().

ngober:
I added a commit that follows Seth's suggestions. I don't really like this as a long-term solution. I feel that as little cache functionality as possible should depend on knowing what level of the cache we are. I have a patch in the works that approaches this from a different angle (by attaching a `this` pointer to packets), but that will have to wait for part 2.

sethpugsley:
I can appreciate that.  One way to generalize this is to check to see if the upper_level_icache is the same as the upper_level_dcache.  If they're the same, that means that the upper level cache is also a unified cache, and if they're different, then that means that this is the highest level unified cache, and it's time to do this special behavior.  This makes it so we don't require the L2 cache to be a unified cache with two higher level non-unified caches, and instead we could put that branching point anywhere in the hierarchy.

GinoAC (approved):

sethpugsley:
@ngober, do you want to implement my latest change, or should we merge it as-is?  I'll approve this as-is, and let Gino actually do the merge.

ngober:
I think it's better to go ahead and merge. I have a few changes I want to make, but they probably deserve their own PRs.

Just as a preview, I'm working on
- Removing the special return behavior for L1 caches (including TLBs) where packets are placed in a processed queue. I want to unify this into the way the `return_data()` function works.
- Changing the `fill_l1i`/`fill_l1d`/`instruction`/`is_data` flags to not need if statements to read. Like I said, I'm trying to attach `this` pointers to the packets. This is a very extensible way to do things. It can eventually support arbitrary cache hierarchies, including letting caches send requests to TLBs for virtual address translation (in VA prefetching, for example)
- Bandwidth limitations in checking the WQ for forwarding and RQ/PQ for duplicates. Currently all of these are fully associative and happen in one cycle, which probably isn't realistic.

By the way, I added one last commit to the branch that reorders the miss logic. It's more clear now and takes up less vertical space.

PR #76, opened 2020-10-24
Implemented Decoupled Front End, plus cleanup Fetch, and fix a few bugs in Decoded Instruction Buffer

sethpugsley:
Cleanup Fetch - the DIB lookup was happening in between seeing if the IFETCH_BUFFER instructions needed to be translated or fetched.  Now DIB lookup happens before checking to see if anything needs to be translated.  This will result in fewer unnecessary ITLB lookups.

Debug DIB - DIB set calculation was done with byte-address of instruction, rather than cache line address (>>6).  Also, instructions are now added to the DIB after they make it through the decode stage the first time, instead of immediately after they're fetched.

This still has the problem that once we add a cache line to the DIB that all instructions from that cache line will be magically fetched and decoded, including those we haven't actually seen yet.  I don't know how big of a problem that is, but this should still be a slight realism improvement over how it was before.

Performance should be very slightly improved by these changes.

Finally, I refactored a few of the loop conditions to make it more obvious how FETCH_WIDTH and DECODE_WIDTH influence what's happening, and I fixed the indentation of a few lines that were giving annoying compilation warnings.

### DIFF
--- a/src/ooo_cpu.cc
+++ b/src/ooo_cpu.cc
@@ -393,13 +393,25 @@ void O3_CPU::fetch_instruction()
 	  if(rq_index != -2)
 	    {
 	      // successfully sent to the ITLB, so mark all instructions in the IFETCH_BUFFER that match this ip as translated INFLIGHT
+	      uint32_t translate_inflight_index = index;
 	      for(uint32_t j=0; j<IFETCH_BUFFER.SIZE; j++)
 		{
-                    if(((IFETCH_BUFFER.entry[j].ip >> LOG2_PAGE_SIZE) == (ifb_entry.ip >> LOG2_PAGE_SIZE)) && (IFETCH_BUFFER.entry[j].translated == 0))
-		    {
-		      IFETCH_BUFFER.entry[j].translated = INFLIGHT;
-		      IFETCH_BUFFER.entry[j].fetched = 0;
-		    }
+		    if(IFETCH_BUFFER.entry[translate_inflight_index].ip == 0)
+		      {
+			break;
+		      }
+                    if(((IFETCH_BUFFER.entry[translate_inflight_index].ip >> LOG2_PAGE_SIZE) == (ifb_entry.ip >> LOG2_PAGE_SIZE))
+		       && (IFETCH_BUFFER.entry[translate_inflight_index].translated == 0))
+		      {
+			IFETCH_BUFFER.entry[translate_inflight_index].translated = INFLIGHT;
+			IFETCH_BUFFER.entry[translate_inflight_index].fetched = 0;
+		      }
+
+		    translate_inflight_index++;
+		    if(translate_inflight_index >= IFETCH_BUFFER.SIZE)
+		      {
+			translate_inflight_index = 0;
+		      }
 		}
 	    }
 	}
@@ -444,13 +456,23 @@ void O3_CPU::fetch_instruction()
 	  if(rq_index != -2)
 	    {
 	      // mark all instructions from this cache line as having been fetched
+	      uint32_t fetch_inflight_index = index;
 	      for(uint32_t j=0; j<IFETCH_BUFFER.SIZE; j++)
 		{
-		  if((IFETCH_BUFFER.entry[j].ip >> LOG2_BLOCK_SIZE) == (ifb_entry.ip >> LOG2_BLOCK_SIZE)
-		     && (IFETCH_BUFFER.entry[j].translated == COMPLETED) && (IFETCH_BUFFER.entry[j].fetched == 0))
+		  if(IFETCH_BUFFER.entry[fetch_inflight_index].ip == 0)
+                      {
+                        break;
+                      }
+		  if(((IFETCH_BUFFER.entry[fetch_inflight_index].ip >> LOG2_BLOCK_SIZE) == (ifb_entry.ip >> LOG2_BLOCK_SIZE))
+		     && (IFETCH_BUFFER.entry[fetch_inflight_index].translated == COMPLETED) && (IFETCH_BUFFER.entry[fetch_inflight_index].fetched == 0))
 		    {
-		      //IFETCH_BUFFER.entry[j].translated = COMPLETED;
-		      IFETCH_BUFFER.entry[j].fetched = INFLIGHT;
+		      IFETCH_BUFFER.entry[fetch_inflight_index].fetched = INFLIGHT;
+		    }
+
+		  fetch_inflight_index++;
+		  if(fetch_inflight_index >= IFETCH_BUFFER.SIZE)
+		    {
+		      fetch_inflight_index = 0;
 		    }
 		}
 	    }
@@ -1615,6 +1637,10 @@ void O3_CPU::update_rob()
 	uint32_t index = IFETCH_BUFFER.head;
         for(uint32_t j=0; j<IFETCH_BUFFER.SIZE; j++)
         {
+	    if(IFETCH_BUFFER.entry[index].ip == 0)
+	      {
+		break;
+	      }
             if((IFETCH_BUFFER.entry[index].ip >> LOG2_PAGE_SIZE) == (itlb_entry.ip >> LOG2_PAGE_SIZE))
             {
 	      if(IFETCH_BUFFER.entry[index].translated == INFLIGHT)
@@ -1658,6 +1684,10 @@ void O3_CPU::update_rob()
 	uint32_t index = IFETCH_BUFFER.head;
         for(uint32_t j=0; j<IFETCH_BUFFER.SIZE; j++)
         {
+	    if(IFETCH_BUFFER.entry[index].ip == 0)
+              {
+                break;
+              }
             if((IFETCH_BUFFER.entry[index].ip >> LOG2_BLOCK_SIZE) == (l1i_entry.ip >> LOG2_BLOCK_SIZE))
             {
 	      if((IFETCH_BUFFER.entry[index].translated == COMPLETED) && (IFETCH_BUFFER.entry[index].fetched == INFLIGHT))
@@ -1681,6 +1711,11 @@ void O3_CPU::update_rob()
               {
                 index = 0;
               }
+
+	    if(j >= IFETCH_BUFFER.occupancy)
+	      {
+		break;
+	      }
         }
 
         // remove this entry
@@ -1745,8 +1780,9 @@ void O3_CPU::update_rob()
 
     // update ROB entries with completed executions
     if ((inflight_reg_executions > 0) || (inflight_mem_executions > 0)) {
-      uint32_t instrs_executed = 0;
-        if (ROB.head < ROB.tail) {
+        uint32_t instrs_executed = 0;
+        if (ROB.head < ROB.tail)
+	  {
             for (uint32_t i=ROB.head; i<ROB.tail; i++)
 	      {
 		if(instrs_executed >= EXEC_WIDTH)
@@ -1755,9 +1791,9 @@ void O3_CPU::update_rob()
 		  }
 		instrs_executed += complete_execution(i);
 	      }
-	    
-        }
-        else {
+	  }
+        else
+	  {
             for (uint32_t i=ROB.head; i<ROB.SIZE; i++)
 	      {
 		if(instrs_executed >= EXEC_WIDTH)
--- a/src/ooo_cpu.cc
+++ b/src/ooo_cpu.cc
@@ -446,9 +446,10 @@ void O3_CPU::fetch_instruction()
 	      // mark all instructions from this cache line as having been fetched
 	      for(uint32_t j=0; j<IFETCH_BUFFER.SIZE; j++)
 		{
-            if((IFETCH_BUFFER.entry[j].ip >> LOG2_BLOCK_SIZE) == (ifb_entry.ip >> LOG2_BLOCK_SIZE) && (IFETCH_BUFFER.entry[j].fetched == 0))
+		  if((IFETCH_BUFFER.entry[j].ip >> LOG2_BLOCK_SIZE) == (ifb_entry.ip >> LOG2_BLOCK_SIZE)
+		     && (IFETCH_BUFFER.entry[j].translated == COMPLETED) && (IFETCH_BUFFER.entry[j].fetched == 0))
 		    {
-		      IFETCH_BUFFER.entry[j].translated = COMPLETED;
+		      //IFETCH_BUFFER.entry[j].translated = COMPLETED;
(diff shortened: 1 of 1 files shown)


sethpugsley:
I should also note that I created this branch, decoupled_frontend, in order to actually implement a decoupled frontend, but I decided to checkpoint the progress I made while laying the groundwork for that.  That's why this pull request doesn't actually include a decoupled frontend yet.

sethpugsley:
The first commit laid the groundwork for the decoupled front end, and fixed a few bugs I found along the way.

The second commit implements the actual decoupled front end.

Decoupled front end means the IFETCH_BUFFER can continue to be filled with instructions, and get further and further ahead of decode and the ROB if they're giving backpressure, for as long as the branch predictor keeps making accurate predictions.

This change makes it so that the entire IFETCH_BUFFER isn't magically marked as translated and fetched as soon as an L1I read completes.  Instead, ITLB and L1I reads can only mark up to FETCH_WIDTH instructions as translated/fetched, which are marked in program order, and the rest of the instructions from that page/cache line in the IFETCH_BUFFER have to start over with their ITLB/L1I access.  Note that if an instruction has completed its ITLB access and is waiting for its L1I access to complete, it only starts over with the L1I access in this situation.  It doesn't get its translation taken away.

Because I'm not changing the size of the IFETCH_BUFFER in this PR, this will lose performance across the board, because there's less magic going on.  However, it opens the door for IPC to now scale with larger IFETCH_BUFFER sizes.  For large enough IFETCH_BUFFER sizes, performance can now also scale dramatically with branch predictor performance without messing around with the various latencies modeled in the core.  The downside is that using a larger IFETCH_BUFFER significantly reduces simulator performance.

With a decoupled front end, the IPC gained from a larger IFETCH_BUFFER and better branch prediction is the same performance that a better instruction prefetcher can provide.  The overlap between a decoupled front end and a good instruction prefetcher isn't 100%, but it is significant.  I expect, but haven't done the study yet, that instruction prefetching's benefit will go way down in the presence of this decoupled front end with a larger IFETCH_BUFFER and better branch predictor.

The changes in these commits should definitely be added to ChampSim, but someone will need to go and figure out what the new correct default IFETCH_BUFFER and DECODE_BUFFER sizes are, and we should probably add those knobs to the configuration json file.  They don't necessarily need to be multiples of FETCH_WIDTH or DECODE_WIDTH.  My initial guess is that IFETCH_BUFFER should be somewhere in the range of 64-128, and DECODE_BUFFER should be about 32.  I don't want to unilaterally make those decisions, though.

src/ooo_cpu.cc
```diff
@@ -304,11 +304,58 @@ void O3_CPU::fetch_instruction()
...
+
+	  available_fetch_bandwidth--;
+      }
+
+      if(available_fetch_bandwidth == 0)
+	{
+	  break;
+	}
+
+      index++;
```
ngober, line 342: This early advance is potentially pretty dangerous. It means that we will not allow two consecutive entries in the IFETCH_BUFFER to be checked in the DIB. The way this appears to work is 

1. IFETCH_BUFFER[head] is checked in DIB.
    - index is incremented
2. IFETCH_BUFFER[head+1] is checked for translation.
    - IFETCH_BUFFER[head+1] was never checked in DIB.
    - index is incremented on L481
3. IFETCH_BUFFER[head+2] is checked in DIB
... and so on

This increment and check should be removed and we should rely on L481 to increment the index.

src/ooo_cpu.cc
```diff
@@ -441,8 +491,8 @@ void O3_CPU::fetch_instruction()
     }
 
     // send to DECODE stage
-    std::size_t checked_for_decode = 0;
-    while (checked_for_decode < DECODE_WIDTH && IFETCH_BUFFER.occupancy > 0 && DECODE_BUFFER.occupancy < DECODE_BUFFER.SIZE &&
+    available_fetch_bandwidth = FETCH_WIDTH;
+    while (available_fetch_bandwidth > 0 && IFETCH_BUFFER.occupancy > 0 && DECODE_BUFFER.occupancy < DECODE_BUFFER.SIZE &&
```
ngober, line 495: This change appears to have no effect except to make a counter that goes down instead of a counter that goes up. Why is that?

src/ooo_cpu.cc
```diff
@@ -304,11 +304,58 @@ void O3_CPU::fetch_instruction()
...
+  // scan through IFETCH_BUFFER to find instructions that hit in the decoded instruction buffer
   uint32_t index = IFETCH_BUFFER.head;
   for(uint32_t i=0; i<IFETCH_BUFFER.SIZE; i++)
     {
-        ooo_model_instr &ifb_entry = IFETCH_BUFFER.entry[index];
+      ooo_model_instr &ifb_entry = IFETCH_BUFFER.entry[index];
+
+      // Check DIB to see if we recently fetched this line
+      dib_t::value_type &dib_set = DIB[(ifb_entry.ip >> LOG2_BLOCK_SIZE) % DIB_SET];
+      auto way = std::find_if(dib_set.begin(), dib_set.end(), [ifb_entry](dib_entry_t x){ return x.valid && ((x.addr >> LOG2_BLOCK_SIZE) == (ifb_entry.ip >> LOG2_BLOCK_SIZE));});
```
ngober, line 317: I don't think I agree that we ought to be searching for the block address instead of the byte address. Since the DIB "stores micro-ops" (it doesn't, but...), it feels to me like an entry should correspond to a single instruction, not a whole line that may contain many instructions.

There might be a case to be made that if the block address hits, we can skip translation, or possibly translation+fetch, but I don't think a block address hit should cause us to skip decode.

src/ooo_cpu.cc
```diff
@@ -475,13 +525,39 @@ void O3_CPU::decode_and_dispatch()
...
+
+	// Search DIB to see if we need to add this instruction
+	dib_t::value_type &dib_hit_check_set = DIB[(db_entry.ip >> LOG2_BLOCK_SIZE) % DIB_SET];
+	auto way = std::find_if(dib_hit_check_set.begin(), dib_hit_check_set.end(), [db_entry](dib_entry_t x){ return x.valid && ((x.addr >> LOG2_BLOCK_SIZE) == (db_entry.ip >> LOG2_BLOCK_SIZE));});
+	if (way == dib_hit_check_set.end())
+	  {
+	    // we have a miss in the DIB, so we need to replace something
+	    // find victim in DIB
+	    dib_t::value_type &dib_set = DIB[db_entry.ip % DIB_SET];
+	    auto way = std::find_if_not(dib_set.begin(), dib_set.end(), [](dib_entry_t x){ return x.valid; }); // search for invalid
```
ngober, line 544: This looks like a bug. The set we are searching here is different than the set searched on L538. Is this intentional?

src/ooo_cpu.cc
```diff
@@ -475,13 +525,39 @@ void O3_CPU::decode_and_dispatch()
...
+	    if (way == dib_set.end())
+	      way = std::max_element(dib_set.begin(), dib_set.end(), [](dib_entry_t x, dib_entry_t y){ return x.lru < y.lru;}); // search for LRU
```
ngober, line 546: These statements could all be combined into
```suggestion
	    dib_t::value_type &dib_set = DIB[db_entry.ip % DIB_SET];
	    auto way = std::max_element(dib_set.begin(), dib_set.end(), [](dib_entry_t x, dib_entry_t y){ return !y.valid || (x.valid && x.lru < y.lru;}); // search for LRU
```

This would remove an if-statement and be more concise.

src/ooo_cpu.cc
```diff
@@ -475,13 +525,39 @@ void O3_CPU::decode_and_dispatch()
     if (DECODE_BUFFER.occupancy == 0)
         return;
 
+    std::size_t available_decode_bandwidth = DECODE_WIDTH;
+
     // dispatch DECODE_WIDTH instructions that have decoded into the ROB
-    uint32_t count_dispatches = 0;
-    while (count_dispatches < DECODE_WIDTH && DECODE_BUFFER.occupancy > 0 && ROB.occupancy < ROB.SIZE &&
+    while (available_decode_bandwidth > 0 && DECODE_BUFFER.occupancy > 0 && ROB.occupancy < ROB.SIZE &&
```
ngober, line 531: I have the same comment here as on L495. This appears to only change the direction of the counter.

src/ooo_cpu.cc
```diff
@@ -346,37 +393,29 @@ void O3_CPU::fetch_instruction()
...
 		{
-                    if(((IFETCH_BUFFER.entry[j].ip >> LOG2_PAGE_SIZE) == (ifb_entry.ip >> LOG2_PAGE_SIZE)) && (IFETCH_BUFFER.entry[j].translated == 0))
-		    {
-		      IFETCH_BUFFER.entry[j].translated = INFLIGHT;
-		      IFETCH_BUFFER.entry[j].fetched = 0;
-		    }
+		    if(IFETCH_BUFFER.entry[translate_inflight_index].ip == 0)
+		      {
+			break;
+		      }
```
ngober, line 402: This check can safely be removed, since the if-statement on the next line will filter it out.

ngober (changes requested):
I think this is a good change. It is a simple tweak that enables us to scale up the IFETCH_BUFFER size enough to show effects of branch prediction.

I have noted some places where your changes confuse me a bit, and at least one place where there might be a bug.

One change that I don't see here that I am surprised not to see is any effect of branch misprediction on the IFETCH_BUFFER. In a real system, the fetch buffer would be flushed on a branch direction or branch target misprediction. Since we do not have any wrong-path execution, we will not have any instructions that need to be flushed. How are you attempting to model that here?

src/ooo_cpu.cc
sethpugsley: The idea of the DIB is that it provides a similar benefit to a modern uop cache, which in the case of the latest Intel core is reported to cache ~2.25k instructions.  A real uop cache stores decoded instructions on the granularity of chunks of several instructions, and not individual instructions or whole cache lines.

If each entry of the DIB stores only a single instruction, then we're only covering 64 instructions.  This is off of our ~2.25k instruction target by a factor of 32.  As a compromise, in our previous conversation on the topic we had decided that the DIB would cache entire cache lines, which can be assumed to hold on average 32 instructions each, getting us all the way to our ~2.25k target.

src/ooo_cpu.cc
```diff
@@ -304,11 +304,58 @@ void O3_CPU::fetch_instruction()
...
-        ooo_model_instr &ifb_entry = IFETCH_BUFFER.entry[index];
+      ooo_model_instr &ifb_entry = IFETCH_BUFFER.entry[index];
+
+      // Check DIB to see if we recently fetched this line
+      dib_t::value_type &dib_set = DIB[(ifb_entry.ip >> LOG2_BLOCK_SIZE) % DIB_SET];
+      auto way = std::find_if(dib_set.begin(), dib_set.end(), [ifb_entry](dib_entry_t x){ return x.valid && ((x.addr >> LOG2_BLOCK_SIZE) == (ifb_entry.ip >> LOG2_BLOCK_SIZE));});
+      if (way != dib_set.end())
+      {
+          // The cache line is in the L0, so we can mark this as complete
+          ifb_entry.fetched = COMPLETED;
```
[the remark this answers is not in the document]
sethpugsley: Good point, I'll update that.

src/ooo_cpu.cc
sethpugsley: This is an entirely separate for-loop.  I broke out the DIB-related stuff into its own for-loop that runs before the translate/fetch loop.  This was done to make it clear that this is an alternative path that instructions can go down that is independent of the main translate/fetch path.

src/ooo_cpu.cc
ngober: I see. That makes a lot of sense. Since we aren't attempting any kind of uop decoding, cache lines are a fair analog for instruction windows.

src/ooo_cpu.cc
sethpugsley: I added it as a performance optimization to actually get out of the for-loop, instead of scanning the entire IFETCH_BUFFER every time.  This loop and ones like it take up more and more time for larger IFETCH_BUFFER sizes.  I put in similar code in other for-loops that scan the IFETCH_BUFFER.

src/ooo_cpu.cc
sethpugsley: Correct, that is the only functional difference, but I believe it improves readability as it makes it more clear that the following loop is going to consume the FETCH_WIDTH resource.  We start with a full stock of the resource, and  as we go through the loop, the resource is drained, and when we run out, the loop is exited.  This matches the code pattern of how cache bandwidth is shared between loads and prefetches in cache.cc.

src/ooo_cpu.cc
sethpugsley: Same as above.

src/ooo_cpu.cc
sethpugsley: You're correct, this is wrong.  Good bug catch.

src/ooo_cpu.cc
sethpugsley: Honestly, I don't exactly understand this code.  I just copied/pasted it from elsewhere and I assume you wrote it correctly in the first instance.  After my changes are merged into develop, will you please take a closer look at this?

src/ooo_cpu.cc
ngober: Now that I opened it in a proper editor, I can see that. It wasn't clear from the git diff. Just to put the thought out there, I think before too long I want to break `O3_CPU::fetch_instruction()` up into three functions: `O3_CPU::check_dib()`, `O3::CPU::translate_instruction()`, and `O3_CPU::fetch_instruction()`. This would increase readability, since it better describes what's going on. That is beyond the scope of this patch, though.

src/ooo_cpu.cc
ngober: I added a commit that does this. I admit, the syntax is a bit clunky, but the code just searches through to find a victim in the DIB, where invalid ways are always the most eligible candidates.

ngober (approved):

ngober:
I think these changes are mergeable, but I want to give @GinoAC some time to look at it first.

GinoAC (approved):
I have no significant comments to the changes other than it looks clear for merging.

ngober:
@sethpugsley are you planning to continue developing on this branch, or should I delete it?

sethpugsley:
I think you can delete it now.  The next thing to do along this line of development is to decide the right default sizes for the IFETCH_BUFFER and DECODE_BUFFER, but that doesn't need to be done on this branch.

PR #79, opened 2020-11-07
DIB instruction window size

sethpugsley:
This change is based on discussion in https://github.com/ChampSim/ChampSim/pull/76#discussion_r512135701

The previous version of the DIB would perform some unrealistic prefetching, and this PR aims to slightly increase the amount of realism.  Before, when an instruction was fetched and decoded, its cache line address would be placed in the DIB, and any other instructions from that cache line would hit in the DIB and be magically fetched and decoded, even if that particular instruction had never been fetched or decoded before (and therefore shouldn't logically be found in the DIB).

This PR reduces the instruction window size of DIB entries from 64B to 16B, and allows this value to be changed via the config file.  In this context, an "instruction window" doesn't refer to the ROB, but rather the instructions that are covered by a single DIB entry.  It has been reported that an Intel Sandy Bridge uop cache entry, for example, covers 6 uops, and the entire uop cache is organized into 32 sets of 8 ways each.  This gives the Sandy Bridge uop cache coverage of up to 1536 uops.  The uop cache is reported to have grown in Sunny Cove to 2.25k entries.

In this PR, each DIB entry covers ~8 instructions, given the rule of thumb of ~2B/x86 instruction on average, and is aligned on 16B boundaries.  I believe real uop cache entry addresses are byte-aligned, but I don't really know.  This means there will still be some effect from prefetching, but it should be reduced by a factor of 4.  To compensate for reducing the coverage of the DIB by reducing each entry's instruction window size, I increased the number of sets by a factor of 4.  This version of the DIB still covers ~2K instructions, just like it did before.

I do not expect there to be much of a performance impact from this, but it should slightly improve the accuracy of the simulation.


ngober (approved):
I like this change quite a bit. It is a good use of the configuration file and shows how we can continue to extend ChampSim in the future.

ngober:
I want to give @GinoAC an opportunity to look at this before merging. I know in particular he keeps an eye on how performance changes.

GinoAC:
For a configuration with no prefetching I'm seeing up to 11% performance loss on SPEC17, with 3% loss on average. The gains from prefetching are consistent with the previous commit despite the lower baseline performance. 

It looks like the performance change comes from a fairly large increase in the number of instructions accessing the L1-I which, from my understanding, is what we would expect from the removal of magic hits.

ngober:
I like that ChampSim is getting to a place where we can make these kinds of changes easily. Merging.

PR #85, opened 2020-11-13
Speed up warmup execution by ignoring register dependencies

sethpugsley:
This commit eliminates register dependencies until the warmup phase is complete.  This causes the warmup phase's IPC to be much higher, leading to faster warmup times.  Warmup phase IPC is already meaningless, so I imagine making it even more meaningless isn't a bad thing.  Branch predictors, cache replacement algorithms, cache contents, and prefetchers are all warmed up as normal.  The final reported IPC of the simulation phase, as well as all other reported metrics, should be very similar to they were before this commit, but not identical.

Please experiment with this and see how much it changes the metrics you care about.  If this doesn't harm anyone else's work, then it might as well be merged in, but it's not a huge deal either way.

### DIFF
--- a/src/ooo_cpu.cc
+++ b/src/ooo_cpu.cc
@@ -230,6 +230,21 @@ uint32_t O3_CPU::init_instruction(ooo_model_instr arch_instr)
 
     arch_instr.event_cycle = current_core_cycle[cpu];
 
+    // fast warmup eliminates register dependencies between instructions
+    // branch predictor, cache contents, and prefetchers are still warmed up
+    if(!warmup_complete[cpu])
+      {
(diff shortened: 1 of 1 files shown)


ngober:
This is an interesting idea to remove all registers from the instructions during warmup. I like it as it is. It may lead to some inflight instructions being in a different order when the simulation phase begins, but that may not be a problem.

sethpugsley:
I agree that it will undoubtedly change things up a little bit at the boundary between warmup and simulation phases, but we were already dealing with that situation anyway.  The current master & develop branch version of warmup is also artificially sped up, which undoubtedly also jumbles things a little bit.  This commit just changes the nature of that jumbling a little bit.

ngober (approved):

PR #86, opened 2020-11-15
Added support for user-defined BTBs, and provided a sample Basic BTB

sethpugsley:
This is not ready for prime time yet, but I wanted to get some eyeballs on it early as I'm expecting it to go through plenty of iteration.  I would appreciate help not only with code, style, and simulator architecture, but also branch target prediction architectures in general.  I'm a BTB novice.

This PR adds non-perfect BTBs to ChampSim.  How this fits in ChampSim is that the branch predictor is now expected to return either 0 for non-taken branches, or the specific byte address of the branch target for taken branches.  The branch predictor can get help coming up with the specific target address from the BTB.  Basically how I'm imagining this will work is that the first line of the branch predictor will call the BTB to get its target prediction, and then the branch predictor can decide whether or not it wants to use that target prediction.  If the branch predictor predicts a branch as not taken, then it can just discard the output of the BTB.

The included basic_btb implements a fairly straightforward 8K entry BTB + 32-entry return address stack. From my testing, it appears to work correctly for direct and conditional branches and direct calls and returns, but it doesn't do anything special for indirect branches or calls, so it mostly gets those wrong.  I have lots of ideas about how indirect branches and calls could be predicted more effectively, but one of my goals for basic_btb is for it to be ... well, basic.  I don't want to invent anything new, and I don't know what the basic version of indirect branch prediction is.  That's something I would appreciate help deciding.

As far as performance goes, basic_btb is so much worse than the previous perfect BTB ChampSim had that it almost makes it irrelevant which branch predictor you use alongside it.  The hashed_perceptron branch predictor is still a little bit better than gshare, but not by nearly as much as it was with a perfect BTB.  I'm imagining that once basic_btb starts to do something about indirect branches that a lot of the performance difference will return.

I'm looking for suggestions about how BTBs should be integrated with ChampSim in general, as well as suggestions for improvements to the particular sample basic_btb I included, while still making sure it remains basic.

Finally, I ran basic_btb.cc through clang-format using llvm style before I commited it, so you can see what that looks like in action if you're unfamiliar with it.

### DIFF
--- a/btb/basic_btb/basic_btb.cc
+++ b/btb/basic_btb/basic_btb.cc
@@ -8,8 +8,9 @@
 
 #include "ooo_cpu.h"
 
-#define BASIC_BTB_SETS 1024
+#define BASIC_BTB_SETS 256
 #define BASIC_BTB_WAYS 8
+#define BASIC_BTB_INDIRECT_SIZE 1024
 #define BASIC_BTB_RAS_SIZE 32
 
 struct BASIC_BTB_ENTRY {
@@ -21,6 +22,9 @@ struct BASIC_BTB_ENTRY {
 
 BASIC_BTB_ENTRY basic_btb[NUM_CPUS][BASIC_BTB_SETS][BASIC_BTB_WAYS];
 uint64_t basic_btb_lru_counter[NUM_CPUS];
+
+uint64_t basic_btb_indirect[NUM_CPUS][BASIC_BTB_INDIRECT_SIZE];
+uint64_t basic_btb_conditional_history[NUM_CPUS];
 
 uint64_t basic_btb_ras[NUM_CPUS][BASIC_BTB_RAS_SIZE];
 int basic_btb_ras_index[NUM_CPUS];
@@ -68,6 +72,11 @@ void basic_btb_update_lru(uint8_t cpu, BASIC_BTB_ENTRY *btb_entry) {
   basic_btb_lru_counter[cpu]++;
 }
 
+uint64_t basic_btb_indirect_hash(uint8_t cpu, uint64_t ip) {
+  uint64_t hash = (ip >> 2) ^ (basic_btb_conditional_history[cpu]);
+  return hash % BASIC_BTB_INDIRECT_SIZE;
+}
+
 void push_basic_btb_ras(uint8_t cpu, uint64_t ip) {
   basic_btb_ras[cpu][basic_btb_ras_index[cpu]] =
       ip + basic_btb_ras_call_return_offset[cpu];
@@ -92,6 +101,7 @@ uint64_t pop_basic_btb_ras(uint8_t cpu) {
 void O3_CPU::initialize_btb() {
   std::cout << "Basic BTB sets: " << BASIC_BTB_SETS
             << " ways: " << BASIC_BTB_WAYS
+            << " indirect buffer size: " << BASIC_BTB_INDIRECT_SIZE
             << " RAS size: " << BASIC_BTB_RAS_SIZE << std::endl;
 
   for (uint32_t i = 0; i < BASIC_BTB_SETS; i++) {
@@ -103,6 +113,11 @@ void O3_CPU::initialize_btb() {
     }
   }
   basic_btb_lru_counter[cpu] = 0;
+
+  for (uint32_t i = 0; i < BASIC_BTB_INDIRECT_SIZE; i++) {
+    basic_btb_indirect[cpu][i] = 0;
+  }
+  basic_btb_conditional_history[cpu] = 0;
 
   for (uint32_t i = 0; i < BASIC_BTB_RAS_SIZE; i++) {
     basic_btb_ras[cpu][i] = 0;
@@ -131,8 +146,11 @@ uint64_t O3_CPU::btb_prediction(uint64_t ip, uint8_t branch_type,
     basic_btb_last_return_target[cpu] = target;
 
     return target;
+  } else if ((branch_type == BRANCH_INDIRECT) ||
+             (branch_type == BRANCH_INDIRECT_CALL)) {
+    return basic_btb_indirect[cpu][basic_btb_indirect_hash(cpu, ip)];
   } else {
-    // use BTB for all branches + calls
+    // use BTB for all other branches + direct calls
     auto btb_entry = basic_btb_find_entry(cpu, ip);
 
     if (btb_entry == NULL) {
@@ -152,6 +170,18 @@ uint64_t O3_CPU::btb_prediction(uint64_t ip, uint8_t branch_type,
 
 void O3_CPU::update_btb(uint64_t ip, uint64_t branch_target, uint8_t taken,
                         uint8_t branch_type) {
+  // updates for indirect branches
+  if ((branch_type == BRANCH_INDIRECT) ||
+      (branch_type == BRANCH_INDIRECT_CALL)) {
+    basic_btb_indirect[cpu][basic_btb_indirect_hash(cpu, ip)] = branch_target;
+  }
+  if (branch_type == BRANCH_CONDITIONAL) {
+    basic_btb_conditional_history[cpu] <<= 1;
+    if (taken) {
+      basic_btb_conditional_history[cpu] |= 1;
+    }
+  }
+
   if (branch_type == BRANCH_RETURN) {
     // recalibrate call-return offset
     // if our return prediction got us into the right cache line, but not the
--- a/btb/basic_btb/basic_btb.cc
+++ b/btb/basic_btb/basic_btb.cc
@@ -1,0 +1,191 @@
+
+/*
+ * This file implements a basic Branch Target Buffer (BTB) structure.
+ * It uses a set-associative BTB to predict the targets of non-return branches,
+ * and it uses a small Return Address Stack (RAS) to predict the target of
+ * returns.
+ */
+
+#include "ooo_cpu.h"
+
+#define BASIC_BTB_SETS 1024
+#define BASIC_BTB_WAYS 8
+#define BASIC_BTB_RAS_SIZE 32
+
+struct BASIC_BTB_ENTRY {
+  uint64_t ip_tag;
+  uint64_t target;
+  uint8_t always_taken;
+  uint64_t lru;
+};
+
+BASIC_BTB_ENTRY basic_btb[NUM_CPUS][BASIC_BTB_SETS][BASIC_BTB_WAYS];
+uint64_t basic_btb_lru_counter[NUM_CPUS];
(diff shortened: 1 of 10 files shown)


alberto-ros:
Regarding a basic indirect branch target predictor, I would suggest something like the Target Cache: http://meseec.ce.rit.edu/eecc722-fall2001/papers/branch-prediction/4/indir_isca24.pdf
It is like the gshare of the indirect branch prediction. To keep it even more basic, the tagless version could be an option.

Although I think that current systems get the target address before the direction of the branch, I do not see a major problem doing it at the same time in ChampSim. However, calling the BTB inside the branch prediction code seems an extra complexity. Maybe calling it at the OoO core code and passing the target as a parameter to the branch prediction (in case the branch wants to use it) could be a better option, if anyway any branch needs to call it.

sethpugsley:
I like the suggestion about moving the calls to the BTB to the OoO core code, and then just passing the results to the branch predictor.  Also, I'll look into the Target Cache paper.

After I added stats that break down mispredicts by branch type, I found that I had more misses for returns than even indirect calls.  I implemented the RAS the first way that popped into my head, but I don't really know how close that is to a realistic RAS.  Could someone please point me toward a resource that teaches about how a RAS is supposed to work?

ngober:
Let me see if I understand you: You propose that the module that has previously served as the branch direction predictor be generalized to serve as the branch target and direction predictor. This is okay to me, but I wonder if the way you're doing it is the right ideal. From a research perspective, you would like to isolate the element that you are researching. If I am a researcher working on branch direction prediction, I probably don't want to concern myself greatly with the branch target prediction and would prefer it to be in a separate module. Does your change allow for that?

alberto-ros:
I think the RAS is well implemented but small. A 64-entry RAS as in today systems may reduce a lot the return misses.

sethpugsley:
I took the comments from @alberto-ros and @ngober and made it so now the OoO core itself calls the BTB functions, and not the branch predictors.  This should make things more modular.  I also reverted the branch predictors back to the way they were before this PR, with the exception of changing the function arguments to include the information from the BTB.  Now the BTB predicts a target, the BPU predicts taken/not taken, and then the OoO core makes the final decision about the branch target.

@alberto-ros I tried to make the RAS bigger (I went all the way up to 256 entries), but the return branch MPKI didn't go down.  I think there's something wrong with the way I did it.

sethpugsley:
I just implemented a basic target cache, similar to the tagless one in the paper @alberto-ros linked above.  It greatly improves the accuracy of indirect branch and call predictions.

I still have not looked into what's wrong with the RAS, so I'm still interested in suggestions about that.

alberto-ros:
I took a look at the RAS. Moving from 32-entries to 64-entries actually improves for me in server-039.

The code is correct as far as I can tell. However, I have the feeling that the trace is not correct. That is, the RAS predict the correct access -- according to the pattern of calls-returns, but the trace says that the predicted address is wrong. So it may be a problem with the trace.

ngober:
I don't think that's likely. The trace reader uses the next IP as the branch target, so it's always correct. If the code predicts a different IP than the trace does, it's the RAS that is incorrect.

https://github.com/ChampSim/ChampSim/blob/51d0dfca45fc06d67d67995fca619d0b6a884a09/src/tracereader.cc#L111

sethpugsley:
I did some work on the RAS, and it turns out that the problem I was seeing was just caused by the trace I was using to debug my code.  The trace I used was originally an x86 trace, and they have different instruction sizes for different versions of "call."  @alberto-ros if you're using one of the IPC-1 traces, then it was originally ARM, and it would not have variable size call instructions.

The commit I just made tries to track the size of each individual call instruction, so that the RAS knows how many bytes to skip when it pops something off the stack.  I refactored the code a little bit, but fundamentally there was nothing wrong with the logic of the RAS I wrote, it's just that the call instructions were not always the same size.  Now that that's fixed, the RAS is extremely accurate, at least for the traces I've looked at.

A few days ago, I said this would need a lot of iteration.  Unless someone can find more problems with it through testing, then I think it's now good enough to merge.  I made a lot of progress really quickly, and learned a lot about branch target prediction in the process.  So please, everyone, test it out, and look it over, and see if you can find any more problems.  If nobody can find any problems, then I propose it be merged into develop in 1 week from now.

alberto-ros:
Let me explain why I mentioned that perhaps the trace (server_039) may be wrong.

I am running an experiment with the current configuration of the branch basic_btb.

At some point in execution there is a call with ip ffff000008111dec, which is stored in the RAS at index 9. After a number of calls and returns reaching up to index 16 in the RAS, a return at index 9 takes place. The predicted address is ffff000008111df0, which seems correct (ras[9]+4), but the actual target address in the trace is ffff000008d57128. So the prediction was incorrect.

This happens again for indexes 8, 7, 6 ... which makes me think that there is a "cut" in the trace, as the return address do not match expected targets.

As far as I know, the if the RAS predicts a target address (in our case different than 0 -- plus an offset), the prediction must be always correct, unless I am missing something.

Said that, this is not very important as it does not happen frequently, but I wanted to explain what I meant that the traces may be "wrong".

djimeneth:
That trace is from the CVP-1 traces generated by Qualcomm and translated to
ChampSim format by me. The addresses that begin with 0xffff.... I believe
are kernel code, i.e. there was an interrupt that took us into the kernel.
I have no problem believing that something weird happened in that part of
the trace. I don't know how they collected the traces. It would be
interesting to see if the return-stack behavior returns to normal if we
filter out the kernel addresses.

Daniel

sethpugsley:
When we are finally popping off a call from the RAS and comparing it to the true value of the branch target, and there's a mismatch, it might be worth it to see if that call appears anywhere on the RAS.  If it's found deeper in the RAS, then it might be the right thing to pop off a few extra entries from the RAS to get things synched back up.  Does that sound like it would fix this problem? or are you seeing returns to addresses that aren't found anywhere on the RAS?

vishalgupta97:
In the implementation of basic BTB, if there is a branch target miss, the fetch stage is stalled until that branch instruction is executed. But for Direct jump/call instructions, don't we know the target of the branch instruction at the decode stage itself ? So, if the branch predictor was correct and there was a target mismatch, can the fetch be resumed after direct jump/call  instruction is decoded ?

sethpugsley:
@vishalgupta97 that sounds reasonable to me.  Does anyone have any other input on that point?  I'm not able to code it up tonight, but I should be able to get to it in the next few days.

djimeneth:
Yeah, he's right. The target of a direct branch or call is known after
decode, because it's just the result of adding an offset from the opcode to
the current PC. Except if it's a conditional branch; then we don't really
know the target until the branch is executed, because it might not be
taken, but you shouldn't stall fetch for predicted not-taken conditional
branches because they don't need a target prediction since they just fall
through to the next PC which you already know from fetch (well you might
not know immediately with x86 if you're not reading from the uop cache or
don't have predecode bits in the icache but that's below the level of
detail ChampSim should model). Branches that are never taken shouldn't be
allocated BTB entries so they would always miss in the BTB but never stall
fetch. A never-taken branch should appear to the front-end to be a NOP that
never consults the BTB nor the direction predictor.

Also, don't stall fetch if you get a BTB miss for a return if the RAS comes
back with a target. But you should also insert return addresses into the
BTB though; the RAS might underflow and then you'll go to the BTB to get
the possible target. If the underflow is due to deep recursion, you might
even get the right target.

Daniel

alberto-ros:
Hi, regarding the RAS misses:
@djimeneth Yes, as far as I checked it only happens for 0xffff... instructions.
@sethpugsley No, the actual target is not in the RAS stack.

btb/basic_btb/basic_btb.cc
```diff
@@ -0,0 +1,239 @@
...
+  } else if ((branch_type != BRANCH_INDIRECT) &&
+	     (branch_type != BRANCH_INDIRECT_CALL)) {
+    // use BTB
+    auto btb_entry = basic_btb_find_entry(cpu, ip);
+
+    if (btb_entry == NULL) {
+      if ((branch_target != 0) && taken) {
+        // no prediction for this entry so far, so allocate one
+        uint64_t set = basic_btb_set_index(ip);
+        auto repl_entry = basic_btb_get_lru_entry(cpu, set);
```
ngober, line 224: I have an easy way to compare and find LRU, but it's currently limited to src/base_replacement.cc. Would you want me to insert it in here? I'm a big fan of code reuse.

ngober (approved):

GinoAC (approved):

sethpugsley:
I implemented @vishalgupta97's suggestion about resuming fetch after direct jumps/calls that have their targets mispredicted are decoded.

I don't think that I want to figure out and implement a strategy to deal with the situation @alberto-ros described for this basic_btb.  I want to keep it basic, and it seems like the solution to that situation might be more complicated.  If anyone can come up with something super simple that fixes it, please create a pull request of your own.

Unless something surprising happens in the next couple days, I think I'm ready for this code to be merged into develop on Monday.

alberto-ros:
If you refer to the RAS weird behaviour in some cases, I agree that it is not necessary. These misses are really infrequent and we do not know if there is something wrong with the trace.

sethpugsley:
If nobody voices any objection in the next few hours, I'm planning on merging this into develop tomorrow morning.

alberto-ros:
I tested it heavily and I have no objections with merging.

inc/ooo_cpu.h
```diff
@@ -219,10 +220,15 @@ class O3_CPU {
...
-    uint8_t predict_branch(uint64_t ip);
-    void    initialize_branch_predictor(),
-            last_branch_result(uint64_t ip, uint8_t taken);
+  // branch predictor
+  uint8_t predict_branch(uint64_t ip, uint64_t predicted_target, uint8_t always_taken, uint8_t branch_type);
+  void initialize_branch_predictor(),
+    last_branch_result(uint64_t ip, uint64_t branch_target, uint8_t taken, uint8_t branch_type);
+
+  // btb
+  uint64_t btb_prediction(uint64_t ip, uint8_t branch_type, uint8_t &always_taken);
```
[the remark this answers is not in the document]
sethpugsley: I'll create a separate PR for a version of this that returns the 2 values using a std::pair<>, and we can see which one we like more.

btb/basic_btb/basic_btb.cc
sethpugsley: I actually think that these user-definable modules should be totally self-contained, so that it will be easier to maintain backward and forward compatibility in the future.  I think that's what we've done so far with all the others, like prefetchers and branch predictors.

PR #87, opened 2020-11-20
Cache refactor, Part 2

ngober:
This is the second and probably final part of my work on refactoring the cache model. It is a follow-up to #75.

Here are the major changes made:

- The hit and miss logic is now unified between the queues in the cache. This means that a hit from the prefetch queue now runs the same code as a hit from the read queue, and a (non-L1D) write miss performs the same code as a MSHR fill. This alone eliminates hundreds of lines of code and makes the cache a lot more comprehensible.
- The cache blocks are now part of a single `std::vector`. This might be an inconsequential change, but in the best case it may improve locality and help the simulation run faster.
- All memory elements now use a producer/consumer model for memory requests. The DRAM is a request consumer (since it does not issue requests itself), and the O3_CPU model is has cache busses that are request producers (since they do not consume requests). The cache, however, is both, since it may issue requests as a result of consuming them. This allows me to introduce the `PACKET::to_return` member, so that the return logic is unified. No more if-statements that do special things to the PROCESSED queue for the L1 caches.

These changes are not entirely performance-neutral for two reasons:

- When merging the miss logic, I had to reorder some calls to the prefetcher interface, which may result in changes in prefetcher behavior.
- I changed the `PACKET::lq_index_depend_on_me` member family to a ~`std::set`~ `std::list` because iterating through that is easier, but the iteration order appears to be different than `fastset`. This means that packets are issued to the caches in a different order (occasionally), but it does not affect correctness.

Some future work that results from this:

- Introducing exclusivity is now trivially easy. All we need to do is add an invalidation to the hit logic and make the if-statement on L378 of cache.cc always true.
- Creating a hardware page walker is easier. It can use the `MemoryRequestConsumer` interface and integrate cleanly with existing logic. Furthermore, it can issue page walk requests to the L1D directly and get returns, now that the return logic is unified.
- The next step in virtual-address prefetching is easier. Data caches can, in theory, issue translation requests directly to the TLBs. We would need to add some disambiguation logic to `CACHE::return_data()`, but this shouldn't be too difficult. This will help us remove the fixed latency.

I am more than happy to talk about why I made any decisions I made here. I think these are good changes for the future of Champsim because they make it more accessible and easier to extend.


sethpugsley:
I just tested this out versus the develop branch.  I ran ./config.sh with the default champsim_config.json for both branches, and ran for 50M warmup and 50M simulation on the 600.perlbench_s-570B.champsimtrace.gz trace.  I found that run time for cache_refactor is about 50% longer than develop.

Is this just due to fancy run-time C++ stuff, or did you change anything algorithmically that has worse O() behavior?  Do you think you'll be able to get that performance back while still achieving the goals you had for this code?

ngober:
I've resolved the performance losses. They were the result of bad algorithmic decision-making on my part. I think this is ready for re-review.

sethpugsley:
I now observe only a 5% run time increase on the same trace where I noticed a 50% run time increase before.  I'll take a closer look at the code now and approve it if everything looks good.  Thanks for fixing the performance.

sethpugsley (approved):
There's a lot here.  This is a simplification of a lot of code that was complex and daunting to work with mostly because of its many redundancies.  This goes a long way to simplifying everything through C++-ification.

I'm sure more changes will be required as people work with the new way the cache is implemented, and discover things they don't like about it, but for now, it looks good to me.  I'm also looking forward to building out the page translation system on top of these changes, but that's a discussion for another time.

GinoAC (approved):
I agree with Seth, these changes are a solid step forward in modernizing the code base. I'm sure we'll hear some issues with it once the changes get some mileage, but we can address them as they come in.

ngober:
I agree. I'm not saying this is a perfect refactor, but it's not significantly different than the previous version. Between this and part 1 I've removed more than a thousand lines, and the whole of the cache model is now below a thousand lines in its entirety. If there are bugs, they're now easier to find. If people want to make changes, there are now fewer places they need to look.

If everything is good, I'll merge this over the weekend.

sethpugsley:
I was just about to merge it when I saw your message.  I think you can merge it immediately, unless you think there's some other reason not to.  I'd prefer for it to be merged tonight so that I can use it as a base for other things this weekend.

PR #88, opened 2020-11-20
Removing some redundant code, making the code more homogeneous (and t…

alberto-ros:
…rying my first pull-request)


ngober (approved):
For context, L451 of ooo_cpu.cc also performs this check. https://github.com/ChampSim/ChampSim/blob/341cebd59b9908ca971998d162c1ac2faa3d02d0/src/ooo_cpu.cc#L541

This change looks good and I see no issue with merging it.

PR #89, opened 2020-11-20
Changing the way the code iterates de ROB, in order to have a single …

alberto-ros:
…body of the loop instead of three identical ones

### DIFF
--- a/src/ooo_cpu.cc
+++ b/src/ooo_cpu.cc
@@ -266,7 +266,7 @@ uint32_t O3_CPU::check_rob(uint64_t instr_id)
     if ((ROB.head == ROB.tail) && ROB.occupancy == 0)
         return ROB.SIZE;
 
-    for (uint32_t i=ROB.head, count=0; count<ROB.occupancy; i=(i+1)%ROB.SIZE, count++) {
+    for (uint32_t i=ROB.head, count=0; count<ROB.occupancy; i=(i+1==ROB.SIZE) ? 0 : i+1, count++) {
         if (ROB.entry[i].instr_id == instr_id) {
             DP ( if (warmup_complete[ROB.cpu]) {
             cout << "[ROB] " << __func__ << " same instr_id: " << ROB.entry[i].instr_id;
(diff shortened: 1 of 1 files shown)


ngober (approved):
I'm always a fan of changes that make the code simpler. This is a good use of the comma syntax in for-loops.

src/ooo_cpu.cc
```diff
@@ -266,32 +266,12 @@ uint32_t O3_CPU::check_rob(uint64_t instr_id)
...
-            }
-        }
-        for (uint32_t i=0; i<ROB.tail; i++) {
-            if (ROB.entry[i].instr_id == instr_id) {
-                DP ( if (warmup_complete[cpu]) {
-                cout << "[ROB] " << __func__ << " same instr_id: " << ROB.entry[i].instr_id;
-                cout << " rob_index: " << i << endl; });
-                return i;
-            }
+    for (uint32_t i=ROB.head, count=0; count<ROB.occupancy; i=(i+1)%ROB.SIZE, count++) {
```
sethpugsley, line 269: I'm pretty sure this function is never called from anywhere.  Am I crazy?  What was it originally intended to do?

sethpugsley:
I'm seeing the same thing that I saw in PR #87, which is that this new commit increases simulator execution time by 50%.  I can understand this code a lot more easily than the code in #87, and I don't see anything fishy going on here.  I double-checked and it really does reproducibly take 50% more time to run with this commit versus develop.  I feel a bit like I'm going crazy, because what are the odds that two PRs would decrease simulator performance by the same amount?  Can anyone else verify this?

sethpugsley:
I just did an experiment where I applied #89 and #87 on top of develop, and the problem does indeed compound with both of those PRs in place.  With both of them, run time increases by almost 100%, versus just 50% with only one of them applied on top of develop.  Can someone else please try this and report your results?

alberto-ros:
Hi @sethpugsley, I think that you are right and using the modulo operator to iterate the large and not-power-of-two ROB ChampSim simulates now, it is too much time consuming. In my laptop is not so much increase but noticeable.

We can have a clean code and performance back to previous numbers by replacing:

i=(i+1)%ROB.SIZE

with:

i=(i+1==ROB.SIZE) ? 0 : i+1

Please, fell free to try it and update it.

sethpugsley:
I just implemented, briefly tested, and pushed the change @alberto-ros recommended, and it seems to restore all of the performance that the % operator lost.  I'm going to approve this PR now, and leave it to @ngober to actually do the merge after he verifies I didn't break anything.

sethpugsley (approved):

PR #91, opened 2020-11-24
btb_prediction() now returns a std::pair<> to return target and alway…

sethpugsley:
This PR replaces pass-by-reference of the always_taken variable in the BTB interface with instead just returning a std::pair<>.  This is probably the more C++-like way of doing things.  Reported IPC is identical, and simulator performance is basically the same, so it's really just a style choice.

It's totally fine if we don't go this route, and this PR is rejected.  Please share your opinions on the style choice, and if everyone likes it, it's fine to merge it.

### DIFF
--- a/btb/basic_btb/basic_btb.cc
+++ b/btb/basic_btb/basic_btb.cc
@@ -148,8 +148,8 @@ void O3_CPU::initialize_btb() {
   }
 }
 
-uint64_t O3_CPU::btb_prediction(uint64_t ip, uint8_t branch_type,
-                                uint8_t &always_taken) {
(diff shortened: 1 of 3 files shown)


ngober (approved):
This is a great way to handle it! I was hesitant to suggest it earlier because some people are hesitant to accept the "return a pair" method of returning two values.

GinoAC:
This looks good to me. I'm always in favor of moving us closer to the world of modern C++ if it doesn't kill IPC performance or 
 simulator performance.

sethpugsley:
Let's see if anyone else has thoughts about this, and if nobody objects, then we can merge it in a couple days.

sethpugsley:
I'll merge this tomorrow if nobody speaks up with objections before then.

PR #92, opened 2020-11-27
Declutter source code directories after compilation

sethpugsley:
I noticed that the Makefile that config.sh generates dumps all of the .o and .d files back into the source directories of the corresponding .cc files, and I thought this made the directories too messy.  In this commit, I changed the config.sh script to generate a Makefile that puts all of the .d and .o files from src/ into obj/ instead.

I know almost nothing about make files, and I'm not great at scripting either, and this was the best I could do, but the real goal that I'm hoping someone else will help me to achieve is to declutter ALL of the source code directories in this same manner.  Currently, all of the user-definable modules like prefetchers, btb, replacement, etc., still have their .d and .o files dumped into the same directories as the .cc files, and I want help changing this.

Could someone who knows about make files and scripting please push another commit to this branch to make it so that all .d and .o files end up in the same place, out of the source code directories?

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -253,7 +253,7 @@ with open('Makefile', 'wt') as wfp:
     wfp.write('.phony: all clean\n\n')
     wfp.write('all: ' + config_file['executable_name'] + '\n\n')
     wfp.write('clean: \n\t find . -name \*.o -delete\n\t find . -name \*.d -delete\n\t $(RM) -r obj\n\n')
-    wfp.write(config_file['executable_name'] + ': $(patsubst %.cc,%.o,$(wildcard src/*.cc)) ' + ' '.join('obj/' + k for k in libfilenames) + '\n')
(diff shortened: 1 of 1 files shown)


ngober:
I was thinking about this when I wrote the current build system. It isn't the most ideal thing to have intermediate files in the source directories, but I don't think it's bad. In particular, it makes recompiling the modules a bit simpler, since we don't have to purge any out-of-date `*.o` files when we reconfigure. If we put the `.o` files somewhere else, our makefile gets more complex pretty quickly. I tried to make the simplest possible makefile, since it was being auto-generated by the configure script. Now, of course we can change it, I'm just not convinced it's necessary.

The intermediate files can be removed by running `make clean`. Would that solve the problem you're having?

sethpugsley:
make clean always works.  I was just annoyed that sometimes when I'm trying to edit a file, the source code file name won't tab-complete on the command line.  That, and aesthetics, are the two problems I have with the way it currently works.

ngober:
I'm closing this pull request because it has become obsolete. My recent changes in #259 put all object and dependency files in `.csconfig/`. If the object files are not desired, the directory can be removed and will be rebuilt when the simulator is reconfigured.

PR #93, opened 2020-11-30
Created main.yml to automatically lint commits

### LINKED ISSUE #84: Code Formatting
There has been considerable talk on #38 about what code format style to adopt and how to implement it. I'm creating this issue to collect all of the discussion in a more topical place.

First, why ChampSim needs to care about code formatting: its purpose is to be an accessible tool for rapid innovations in computer architecture research, particularly in memory prefetching and replacement. In order to be accessible, it has to be readable and comprehensible. Therefore, any style that ChampSim adopts must prioritize comprehensibility.

Any style tool that ChampSim uses must also not be a burden to a novice programmer. We see many people using ChampSim that have never worked on a C++ project before. ChampSim is designed with modularity, that is, you can add your own custom prefetchers, branch predictors, and replacement policies. These modules, being primarily used in-house, need not respect any ChampSim format guidelines and should not be required to. Yet, contributors to the codebase should be required to maintain proper style (for self-evident reasons). Therefore, any style tool that Champsim uses must be as optional as possible for researchers while being as required as possible for maintainers.

In the other thread, two toolchains have been suggested:
 - [editorconfig](https://editorconfig.org/)
 - [ClangFormat](https://clang.llvm.org/docs/ClangFormat.html)

Let's use this thread to discuss two things: What style to adopt and what toolchains to use to implement it.

ngober:
Resolves #84.

This workflow is triggered on pushes and PRs to the master and develop branches. It checks out the code, runs clang-format with llvm formatting (this can be later changed to a .clang-format file in the repository), then commits the resulting formatted code.

I have made this PR to master, not to develop, so that the changes will take effect immediately.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,25 @@
+name: lint
+  
+on:
+  # Triggers the workflow on push or pull request events for master and develop
+  push:
+    branches: [ master,develop ]
+  pull_request:
+    branches: [ master,develop ]
+
+jobs:
+  build: 
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: DoozyX/clang-format-lint-action@v0.11
+        with:
+          source: 'src inc prefetcher branch replacement btb'
+          style: llvm
+          inplace: True
+      - uses: EndBug/add-and-commit@v4
+        with:
+          author_name: Auto-Formatter
+          author_email: format@example.com
+          message: 'Formatted with clang-format'
+          add: 'src inc prefetcher branch replacement btb'


ngober:
**NOTICE**

Apparently this PR will perform all formatting immediately. 

We may not want to add this to master, since it will probably cause significant diffs with develop. I'm not immediately sure how to integrate this. We may want to merge into develop instead.

sethpugsley:
I think that this should be applied to develop, and it should just get merged into master when we merge the two in the hopefully-not-too-distant future.  Is it easier on you if this is merged into develop before or after your latest cache_refactor work?

I don't really understand the main.yaml file, but is it going to try to run clang-format on things that aren't C++ files?  Can we specify that it should only run on .h and .cc files?  Also, there's a single .cc file in scripts/ that we might as well hit while we're at it.

ngober:
This is probably the kind of merge we should do right before merging develop into master. It's too bulky to merge easily when there are multiple projects in flight.

I don't totally understand the YAML file either, but I experimented a few times until it worked. By default, it runs on files in the given directories with extension 'c,h,C,H,cpp,hpp,cc,hh,c++,h++,cxx,hxx'. We can tighten that list if we want, but the default is fine. I can add scripts/ to the list of directories.

PR #94, opened 2020-12-05
Implemented the stack pointer folding technique to boost IPC

sethpugsley:
This PR implements a ChampSim version of stack pointer folding, which is an old technique that breaks dependencies on the stack pointer.  The idea is that the true value of the stack pointer can be calculated as soon as an instruction is decoded, because pushes, pops, calls, and returns all have behavior that is fixed at decode time.  In the ChampSim version, we're really just saying that stack pointer reads don't have to wait for the latest write to the stack pointer for the dependency to be satisfied.

This PR greatly improves performance of the DPC3 traces, or anything else that was generated with the Intel Pin tool.  It does not, however, improve the performance of the IPC-1 traces at all.  I suspect that this is because the IPC-1 traces were converted from ARM traces, and the hacks that were used to generate calls & returns in the IPC-1 traces were not universally applied to all instructions that read and write the stack pointer.  I'm hoping @djimeneth can weigh in on that.

### DIFF
--- a/src/ooo_cpu.cc
+++ b/src/ooo_cpu.cc
@@ -552,9 +552,25 @@ void O3_CPU::decode_and_dispatch()
 	    way->lru = 0;
 	    way->addr = db_entry.ip;
 
-        // Add to ROB
-        ROB.entry[ROB.tail] = db_entry;
-        ROB.entry[ROB.tail].event_cycle = current_core_cycle[cpu];
+	    // Stack Pointer Folding
+	    if(db_entry.is_branch == 0)
+	      {
+		// the exact, true value of the stack pointer for any given instruction can
+		// be determined immediately after the instruction is decoded without
+		// waiting for the stack pointer's dependency chain to be resolved
+		for (int i=0; i<NUM_INSTR_SOURCES; i++)
+		  {
+		    if(db_entry.source_registers[i] == REG_STACK_POINTER)
+		      {
+			db_entry.source_registers[i] = 0;
+			db_entry.num_reg_ops--;
+		      }
+		  }
+	      }
+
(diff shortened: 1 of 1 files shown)


djimeneth:
What if there is an instruction like subq %rax,%rsp e.g. in the C++
implementation of allocating an array whose size is a variable? You
wouldn't know the value of %rax until its producing instruction had
executed, well beyond the decode time for the subq. Does your code verify
that the stack pointer is computed correctly, and roll back if it isn't?

As I recall, the translated CVP traces still use the original ARM names for
the stack pointer, except for branch instructions. If the pull request is
approved I can go in and see what it would take to translate the ARM stack
pointer name to the correct Pin name. Probably not too hard, but hopefully
it wouldn't have other weird side effects.

Daniel

sethpugsley:
@djimeneth that's a good point about your subq example.  No, the code never rolls anything back, it needs to just try to get it right the first time.  Some inaccuracies will undoubtedly persist, but I think I can do a little bit better than this current version.  Let me iterate on this a little bit over the weekend, and hopefully by Monday I'll have something that works correctly even in situations like that.  It will probably come down to actually identifying pops & pushes, and then only applying SP folding to those and calls & returns.  I'll have to think about it.

ngober:
As you're looking into it, take a look at #32. In that, I added a way to remap the register indices at runtime. If the problem is that the register indices in the IPC1 traces are ARM-like, then we can simply create a new ARM-to-PIN mapping and the problem might be solved outright.

sethpugsley:
Thanks for the suggestions, everyone.  My git branches got tangled up, and I need to try this again.  Sorry this is messy, and expect another PR soon.

PR #95, opened 2020-12-05
Implements a ChampSim-friendly version of stack pointer folding

sethpugsley:
This is a do-over of #94.  Sorry.

This PR implements a ChampSim-friendly version of the concept of stack pointer folding.  The idea is that for most instructions that manipulate the stack pointer, the new value of the stack pointer can be accurately determined before the instruction is executed.  For example, pushes, pops, calls, and returns all move the stack pointer in predictable ways, and we don't need to execute one of these instructions before knowing what the stack pointer will be as input to the next.  This lets us break the dependency chain created by the stack pointer.

However, it is possible to directly modify the stack pointer by a variable amount without using one of the above instructions.  In this case, you do need to wait on the execution of this instruction before subsequent instructions can read the stack pointer.  This PR covers both cases.

This PR improves the performance of some traces by a lot, others by a little, and some not at all.  The more register spilling and function calls that a program has, the more this will help.  However, the IPC-1 traces, which were originally converted from ARM traces, used a hack to generate calls and returns in the ChampSim traces, so other references to the stack pointer don't use the same register number.  This could be fixed by either re-generating the IPC-1 traces, or perhaps by doing something similar to #32.

### DIFF
--- a/src/ooo_cpu.cc
+++ b/src/ooo_cpu.cc
@@ -186,6 +186,31 @@ uint32_t O3_CPU::init_instruction(ooo_model_instr arch_instr)
 	arch_instr.branch_target = 0;
       }
 
+    // Stack Pointer Folding
+    // The exact, true value of the stack pointer for any given instruction can
+    // usually be determined immediately after the instruction is decoded without
+    // waiting for the stack pointer's dependency chain to be resolved.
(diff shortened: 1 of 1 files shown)


ngober (approved):

sethpugsley:
I'm going to merge this into develop in the morning, unless there's a big objection before then.

Issue #90, opened 2020-11-20
Two questions about the OoO code


alberto-ros:
Hi,

I have two questions regarding the OoO core code. 

The first one is why in main.cc, the schedule and execute calls for normal instructions and memory instructions are alternated:

             ooo_cpu[i].schedule_instruction();
	      // execute
	      ooo_cpu[i].execute_instruction();

	      ooo_cpu[i].update_rob();

	      // memory operation
	      ooo_cpu[i].schedule_memory_instruction();
	      ooo_cpu[i].execute_memory_instruction();

In the other stages, the calls go from later to earlier stage, which makes more sense to me.

The second question is why the LQ or SQ are not filled along with the ROB at dispatch, but at schedule. It seems to me that these queues can be filled out-of-order even if, as commented in the code, the schedule stage is in-order. My understanding was the ROB, LQ, and SQ are filled at the same stage, and usually at the last in-order stage at the front-end.

Thanks a lot,
Alberto

sethpugsley:
These are both good points.  Eventually someone should do something about both of them.

For the first point--I agree that we should be more consistent in the last-to-first ordering of things within the main simulation loop.  Also, the redundant calls to update_rob() should probably be eliminated.  This fix should be fairly straightforward.

To answer the second point--I didn't know what I was doing when I first coded up all of that stuff years ago.  That's why the LQ and SQ entries are allocated when they are, and why they are separate from the L1D read and write queues.  This is clearly not correct, but it is functional for now.  I'm scared of touching all the LQ and SQ stuff though, because it's pretty complicated.  Anyone else is welcome to give it a try, but it's not high on my priority list right now.

alberto-ros:
Thanks, Seth. I will take a deeper look at the OoO and see if I can do some modifications in this regard.

Issue #63, opened 2020-08-31
Handling of event_cycle is inconsistent


ngober:
The event_cycle member (of ooo_model_instr, PACKET, etc.) is handled inconsistently throughout ChampSim. It is supposed to be used to model delays. An object is not ready until its event_cycle has passed. Here are two examples of how it is handled.

https://github.com/ChampSim/ChampSim/blob/3064f5c3dca1961596fdc831f6029f9af9a17a04/src/ooo_cpu.cc#L832
https://github.com/ChampSim/ChampSim/blob/7867885f52686f46432fa8173400f6a5ce4a5da5/src/cache.cc#L236

In one, it is compared to the current cycle with `<`, in the other, with `<=`. I think this would ideally be handled with a function like

    template <typename T>
    bool is_ready(T item, uint64_t cpu)
    {
        return item.event_cycle < current_core_cycle[cpu];
    }

which we can then use consistently throughout. Such a function can, because of templating, be applied to any struct or class that has an event_cycle member. 

This change is likely not going to have significant performance impact, but it serves to clean up the codebase.

ngober:
Closing following the merge of #124.

PR #96, opened 2020-12-07
Refactors and reorders some main simulation loop functions

sethpugsley:
This PR addresses some of the issues raised in #90 by @alberto-ros.

The goal of these changes was primarily to have the code make more intuitive sense for people who are familiar with CPU architecture and simulation.  Along the way, I discovered the true impact of that multiple calls to update_rob() was having (which was a horrible, and inaccurate function name, by the way), and so I refactored the code a bit to add a considerable amount of additional realism into the simulation.  Or, at least now the simulator does more of what I thought it did all along, but didn't actually do.

update_rob() served basically 2 functions.  It dealt with all of the returning memory operations from the PROCESSED queues, and it also checked for instructions that had .executed==INFLIGHT, and made them COMPLETED.  The queue processing was limited to 1/cycle, and the execution finalizing was limited to EXEC_WIDTH/cycle, but since we were calling it multiple times per cycle, it was doing both of these things way more than that.

The new code separates out these two memory and execution behaviors into separate functions.  The execution-related function now only advances EXEC_WIDTH instructions from executed==INFLIGHT to COMPLETED, and a separate memory-related function now processes a number of items from each PROCESSED queue equal to the L1I and L1D per-cycle bandwidth.  These are each only called once per cycle.  The net effect is that performance is waaaay down with this patch compared to before, but it is now more sensitive to L1I/D bandwidth and EXEC_WIDTH.  This is a big win for simulation accuracy.

I also think this lays the groundwork to figure out other things that are bugging me, like if the load-to-use latency of L1D hits is really what we think it is.  The more realistic the simulation is, the easier it should be to find the true bottlenecks in the model.


ngober:
I like the idea of this change, but I have a couple questions.

First, in `handle_memory_return()`, you use the L1I bandwidth for the ITLB loop (and similar for data). Is this intentional? Why not use the TLB bandwidth?

Also, due to other work I've done, I've noticed that the order of the functions in `O3_CPU::operate_cache()` causes shootthrough during warmup. It doesn't happen as badly in the ROI because the latency is applied when the packet is added to the cache queue. This might not be a big deal, since it is during warmup, but would it be a good idea to fix that in this patch?

Finally, the `O3_CPU::fetch_instruction()` function has gotten very large and unwieldy. Would it be wise to break it up into parts?

alberto-ros:
To me the code seems more clear now and it is a clear step forward in clarity and accuracy. Still, it has not addressed my two questions in #90 . For example, it is necessary to interleave the schedule and execution phases for memory and other instructions? Do the LQ and SQ should be allocated in order and along with the ROB?

Also, regarding this part of the code, I have a local modification where I split the decode and dispatch stages (and add a dispatch queue), to make the code clearer. If you think that it would be good that those changes are incorporated to the develop (or this) branch I can do the pull-request.

sethpugsley:
This latest push hopefully prevents the shoot through you're seeing, @ngober.  I'll have to think about the MAX_READ thing a little more, but the real answer is probably that all of the L1s (including caches and TLBs) should use LOAD_WIDTH as their MAX_READ.

@alberto-ros, this PR does not attempt to do anything about the LQ and SQ being allocated too late, because that fix is likely to be quite big, and I'm scared of it.  As for the other issue in #90, I guess I don't understand what you  mean.  Is there something bad about ordering these as execute--schedule--execute_mem--schedule_mem?  Instructions need to have their loads complete before they can execute, and scheduling needs to happen first, which means they should appear later in the main simulation loop, right?

As for the dispatch code, please push to this branch, and incorporate it into this PR.

alberto-ros:
I did a (direct) push to this PR of my version where I split decode and dispatch stages. I set the latencies to 1 cycle for each stage, and the number of entries of each buffer to 32. The default width is 6 for both stages.

Regarding the schedule/execute order, my only concern is having schedule before execute_mem. As you say, if schedule needs to happen first, it should appear later. However, if the pipeline paths are independent for memory and other instructions (I have not checked this) then it should be equivalent than having: schedule, schedule_mem, execute, execute_mem. Ideally, I think that both schedule functions and both execute functions should be merged into one, but I guess, this is a very minor thing.

src/ooo_cpu.cc
```diff
@@ -1563,178 +1601,204 @@ void O3_CPU::reg_RAW_release(uint32_t rob_index)
...
+    L1I.operate_writes();
+    L1D.operate_writes();
+    L2C.operate_writes();
+
+    L2C.operate_reads();
+    L1D.operate_reads();
+    L1I.operate_reads();
+    STLB.operate_reads();
+    DTLB.operate_reads();
+    ITLB.operate_reads();
```
ngober, line 1616: I'm not certain what the purpose here is to have the writes propagate in one direction and the reads propagate in another. Why will writes be able to shoot through, but not reads?

ngober (approved):

src/ooo_cpu.cc
sethpugsley: It's more important for performance realism reasons to prevent fills from shooting through on their way from DRAM to the L1 caches.  This makes it so the transition between each level in both directions takes at least 1 cycle.

sethpugsley:
@alberto-ros let me explain the life cycle of an instruction in the ROB, and I think it will become more clear why I think this is okay the way it's written.

When an instruction is first allocated into the ROB in-order, it can't do anything until its register dependencies are satisfied.  So it waits around until other instructions execute, and make its source registers available.  After all source registers are available, then it can begin issuing loads, if it has any, by first "scheduling_memory_instruction" and then "executing_mememory_instruction."  After all of its loads have returned data, then the instruction can finally "schedule_instruction" and "execute_instruction."  After an instruction has executed, it will then eligible to retire in-order.

So if we want to model a pipeline, we need to order the functions last-to-first like this: execute_instruction(), schedule_instruction(), execute_memory_instruction(), and finally schedule_memory_instruction().  If there are any problems with this, it might be in a poor choice of function names, especially "execute_memory_instruction()."  Any suggestions on what to rename any of these functions?

src/ooo_cpu.cc
ngober: This makes me wonder if it's better to fix shootthrough using the `event_cycle` members, setting it to a later cycle. That seems like it could fix the problem. I noted in #63 that this is already not handled well.

GinoAC (approved):

alberto-ros (approved):

alberto-ros:
Thanks, Seth, for the explanation. Now, it is more clear to me the way the pipeline is implemented. I think that this pull request is ready for merging it into the develop branch.

sethpugsley:
I'm going to merge this into develop in the morning, unless there's a big objection before then.

PR #98, opened 2020-12-07
Added cvp to champsim trace converter

GinoAC:
This is the trace converter for CVP to ChampSim friendly format. I've included a short README warning that this does not cover every conversion case. In the future we may want to do a bit of refactoring if it generates enough interest, but until then it'll be out there for anyone that needs it.


sethpugsley (approved):
I just tested this out with the traces available on the CVP2 website, and it does appear to work without any problems.  It produced a viable ChampSim trace, and at least the branch statistics that the converter reports matched ChampSim's final output almost perfectly.

My only suggestion is that the README should include a URL pointing to where the CVP and/or CVP2 traces can be downloaded from.

Issue #100, opened 2020-12-10
[Help Wanted / Question] Running ChampSim Tracer


yogeshmo:
I'm trying to create a trace, but I'm not exactly sure what steps I should be following.

I ran `./make_tracer.sh`, but I keep getting the following error:
`makefile:15: /Config/makefile.default.rules: No such file or directory
make: *** No rule to make target '/Config/makefile.default.rules'.  Stop.`

Can someone help me out?

GinoAC:
In make_tracer.sh, change the PIN_TOOL path to match where your pintool is located.

yogeshmo:
This worked, thank you!

Issue #50, opened 2020-06-26
Missing namespaces in tracer/champsim_tracer.cpp


wilsonwang881:
Could you please attach ```std::``` in front of every i.e. ```cerr```, ```endl```?

The champsim_tracer pintool is not compiling with Pin 3.14 (64-bit) with those namespaces missing from the source code.

PR #101, opened 2020-12-11
Small update to tracer and REAME

GinoAC:
This updates the README to direct trace-users to change the tracer building script to point to the Pintool directory (#100). I've also addressed the namespace error brought up in (#50). This can probably be pushed directly into master since it doesn't directly affect ChampSim's functionality.

### DIFF
--- a/tracer/champsim_tracer.cpp
+++ b/tracer/champsim_tracer.cpp
@@ -14,6 +14,8 @@
 #define NUM_INSTR_DESTINATIONS 2
 #define NUM_INSTR_SOURCES 4
 
+using namespace std;
+
 typedef struct trace_instr_format {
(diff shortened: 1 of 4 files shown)


tracer/champsim_tracer.cpp
```diff
@@ -14,6 +14,8 @@
 #define NUM_INSTR_DESTINATIONS 2
 #define NUM_INSTR_SOURCES 4
 
+using namespace std;
```
rajesh-s, line 17: This is also useful in getting tracer working on newer PIN tool versions!

GinoAC: Great! Thank you for confirming this.

ngober (approved):

PR #103, opened 2020-12-11
Added note in README on ChampSim tracer & PIN tool compatibility

rajesh-s:
Newer PIN tool versions can be made use of, since 3.2 is dated. Using a newer PIN tool version is necessary when working on newer linux kernels.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,9 @@ Get this version of Pin:
 wget http://software.intel.com/sites/landingpage/pintool/downloads/pin-3.2-81205-gcc-linux.tar.gz
 ```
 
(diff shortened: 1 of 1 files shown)


ngober (approved):
This is a good note to follow up on #101. Looks good for a merge.

rajesh-s:
Authors, if you feel like this might make it easier to distinguish between PIN tool issues and ChampSim tracer, please include this note:

You can test Pin tool installation/compatibility with your OS by using the following before using it with ChampSim tracer:

wget https://software.intel.com/sites/landingpage/pintool/downloads/pin-3.17-98314-g0c048d619-gcc-linux.tar.gz
tar -xzvf pin-3.17-98314-g0c048d619-gcc-linux.tar.gz
cd pin-3.17-98314-g0c048d619-gcc-linux/source/tools/ManualExamples
make
cd ../../../
./pin -t source/tools/ManualExamples/obj-intel64/inscount0.so -- ls # Testing with LS command

PR #106, opened 2020-12-21
Load to use latency printout

sethpugsley:
This commit is not ready to merge, but I would like some other people to take a look at it.

All I've added so far is the ability to visualize the load-to-use latency of load instructions, from the time that the instruction's register dependencies are met to when the instruction executes.  This enables us to pin down the actual latency of an uncongested L1D, L2C, or LLC  load hit.  The actual load-to-use latency is much higher than I had previously assumed, due to loads and fills traversing all of the buffers between cache levels.

In the current version of the develop branch, configuring the DTLB to have 1 cycle of latency and the L1D to have 5 cycles of latency actually yields a load-to-use latency of 10 cycles.

Let's say that we wanted our cache to have load-to-use latencies similar to an Intel Sunny Cove core, as given here:
https://www.anandtech.com/show/14664/testing-intel-ice-lake-10nm/2
L1D - 5 cycles
L2C - 13 cycles
LLC - 30 cycles

We would have to configure ChampSim's caches with the following latencies in order to observe actual load-to-use latencies that match Sunny Cove:
DTLB - 0 cycles
L1D - 1 cycle
L2C - 7 cycles
LLC - 10 cycles

The reason I started this PR with only the ability to visualize load-to-use latency, and without actually changing the cache latencies, is because I wanted to get peoples' opinions on the right way to proceed.  Should we make people do the math themselves when they're filling out a configuration file and enter the unintuitive cache latencies I listed above that will lead to the correct result they actually want, or should we let people fill out the cache latencies they think they want, and then change them behind the scenes to the correct values?  If we go with the second one, what's the best place to do this?

2 more things.  1) In the version of this PR that should ultimately be merged, the load-to-use latency printing will be off by default (it's currently on). 2) develop can no longer automatically merge into master.  What should we do about this?


alberto-ros:
I think that it would be interesting to know where are the extra cycles are added and if it makes sense to add them or not. 
Also, could it be that getting the latency from the time the register dependencies are ready is not what we want here? I mean, the numbers reported by Anandtech correspond exactly to that time? 
Finally, I would opt for keeping the latency as it is now and subtract it in the simulator, but with care. For example in my simulator, when I specify 5 cycles L1 hit, I know that the first cycle is for the address calculation, which is done in a "execute" stage in the pipeline, so then, in the next stage ("memory"), which is when I actually access the L1 cache, it would just take 4 extra cycles. Since the TLB access is done in parallel to the cache access, the TLB latency does matters there (if lower than 4).

ngober:
Alberto, for the case you mention with address calculation being performed in the execute stage, is that handled already in our execute stage, with the execute latency provided? If that's the case, should we subtract the execute latency out from the L1 latency behind the scenes (in your example, the user specifies an execute latency of 1 and an L1 latency of 5), or should we ask the user to make the distinction (an execute latency of 1 and an L1 latency of 4)?

I suppose my question is this: is it more intuitive to think of the L1 as happening in parallel with the execution, or in series?

alberto-ros:
It is hard to answer what is more intuitive.

If cache latencies are usually mentioned in reports from issue to the completion of the memory instruction, then I would suggest to indicate overall latency and let the simulator to subtract latencies (that is give an input to the simulator of 1 for execution and 5 for L1 as it is now).

However, in all simulators that I know when a cache latency is specified it is usually added to the cache access part. So this would be more intuitive for most researchers. And also if someone takes CACTI or McPAT and gets the access latency for ChampSim default L2C would get definitely less than 13 cycles.

So a solution could be to specify as LATENCY of a cache level just the latency of accessing the structure, and making the simulator to print also for each cache level the minimal ISSUE_TO_COMPLETED latency.

This is a very good question indeed.

sethpugsley:
Real CPUs are implemented as a collection of pipelines, which ChampSim really doesn't do.  My understanding is that in a real CPU, the L1D load-to-use latency is really just the length of the L1D pipeline.  The L2C load-to-use latency would be the length of the L1D pipeline, plus the L2C pipeline, plus the latency of any buffering between them.

In a real CPU, the L1D pipeline does many different tasks, some of which can be done in parallel.  These tasks include such things as VA->PA translation, memory disambiguation, tag array access, MSHR lookup and allocation, data array access, data rotation, and probably many others I'm not remembering.  The actual data array access is probably not where the majority of the latency is coming from in the L1D pipeline.  The L2C pipeline is probably a little simpler, because it doesn't have to do VA->PA translation or memory disambiguation, but it also has a lot going on.

For a real CPU, I would define load-to-use latency as the time between when a ready Load Buffer entry is selected and sent down the L1D pipe, and when the result of that load is consumed by the execution of an instruction and/or written to the register file.

In ChampSim it's trickier, because we don't actually have an L1D pipe, and we don't have a register file.  We do have a Load Buffer, but its behavior does not really line up with a real CPU (@alberto-ros has mentioned before that the LB, along with the SB, really should be allocated by instructions when they first enter the ROB).  Loads selected out of ChampSim's current LB do serial lookup of the TLB and then the L1, which is also not right.

The complete solution would be to do all of the following (and maybe more):
Change it so that LB and SB are allocated when instructions enter the ROB
Do TLB lookups in parallel with L1 caches
Have a single "L1D pipeline latency" parameter that represents the entire L1D pipeline, rather than it being broken out between TLB latency and L1 latency

On the other hand, this PR is about working within the inaccurate way we're currently doing things to make the observed load-to-use latencies match more closely to what they would be if we had the above changes.

ngober:
I'm closing this because it is stale. If there is further interest, we can reopen it.

PR #107, opened 2020-12-23
Added configuration dump to replace example configuration

ngober:
Previously, there was an example configuration file in the root of the repository. This was fine as a temporary measure, but it isn't extensible, as new features would need to be added manually. This patch adds a `--dump-config` option that prints the current configuration to the command line (and exits without configuring).

As a bonus, the configuration script now accepts `-`, that is, the standard input, as a valid configuration file which... might be useful to someone... I suppose.


sethpugsley:
I have a couple of things to say.  First, when you do --dump-config, it prints out an extra line of text before the dump, so you can't just redirect the output > to a new config file and immediately use it before going in and editing the new config file.  Even if that were fixed, without more documentation, some new users might assume that the right thing to do is to directly edit the defaults in config.sh, and not build and edit their own config files.

Second, I don't really like eliminating the example config in the default ChampSim package.  Not including any example config files could make it sound like we're advocating for the ONE TRUE CONFIGURATION that's hard-coded in config.sh.  I like the idea of there being an example_config.json that's included, and off of which people can build.  I would like to take the idea of example configs even further and include a client_config.json and server_config.json to show people how to change the knobs to change things like memory bandwidth, and maybe an "in-order" config (or at least as close as ChampSim can get ...), and an ultra-wide config.

ngober:
I think the idea about shipping ChampSim with multiple example config files is great, but at that point, you could make an argument for no defaults at all, and that the configuration should be fully-specified, minus any that could be inferred.

I drew inspiration here from the way clang-fomat does configuration. Maybe the solution is to have a few hard-coded defaults, have the user select one, then override the defaults with the user's provided configuration, like clang-format does?

ngober:
I'm considering this patch again given #243 and some anecdotes I've been hearing from others. I think that the inclusion of the file `champsim_config.json` is causing people to think that the only way to specify a configuration is to completely specify the configuration, and this is not the intention. The intent of the configuration system is to allow the user to rely on defaults, but this is not coming across with the large configuration file visible in the root. I think a better solution is an `examples/` directory, plus this patch.

ngober:
After another long time, I'm revisiting this patch. After 811aeea, there is a collection of configurations that are used for testing. A [wiki page](https://github.com/ChampSim/ChampSim/wiki/Creating-a-configuration-file) describes how to create a good configuration file. Still, without fail whenever I am interacting with a user, they are editing a clunky frankenstein of `champsim_config.json` that doesn't make a lot of sense. I don't know if this patch is the correct way to resolve the issue, but the issue certainly exists.

PR #113, opened 2021-01-07
Added cache bandwidths to constructors

ngober:
This patch moves the number of reads and writes per cycle into the constructor of the caches and adds support for changing them to the configuration file. Previously, these values were hard-coded in the constructor of the core model.


sethpugsley:
Can you please include an updated champsim_config.json with the new parameters filled in as part of this PR?

ngober:
Updated champsim_config.json.

PR #114, opened 2021-01-08
Moved DRAM configuration to config file

ngober:
This patch moves six more DRAM configuration items to the config file.

In the future, I want to make these members of the `MEMORY_CONTROLLER` class, but leaving them as preprocessor defines will work for now.


sethpugsley:
This one also now needs some conflict resolution.

Issue #97, opened 2020-12-07
Packet Queue removal is non-contiguous


ngober:
When a packet is removed from a `PACKET_QUEUE`, it is done in a very unsafe way.

The removal occurs on L111 of src/block.cc.

https://github.com/ChampSim/ChampSim/blob/7867885f52686f46432fa8173400f6a5ce4a5da5/src/block.cc#L111

In it, the passed pointed-to value is replaced with an empty packet. Then, the head is incremented a few lines later. However, there is no guarantee that the passed pointer is at the head of the queue or even is a member of the current queue. In theory, you could do something particularly stupid like

    RQ.remove_queue(WQ.entry[PQ.SIZE])

Which, while syntactically valid and compilable, is a huge bug. Now, the only place where a non-head element is removed from the queue is in the DRAM model.

https://github.com/ChampSim/ChampSim/blob/e367f9c79aef714528bec795bb817c0669c297d9/src/dram_controller.cc#L357

The `request_index` here is the `next_process_index`, which is set on L589 of the same file, where it is set to a scheduled packet with the earliest event cycle (with ties broken by index closest to zero? Is this an unrelated bug?). 

Are we getting lucky that there are no holes appearing in the queues and that packets are not getting popped from the queue in an untimely way?

I ran a quick experiment to verify what was going on. I added an assertion to `PACKET_QUEUE::remove_queue()` that checked the passed pointer to determine if it is inside the bounds of the queue, and nothing immediately jumped out as wrong. I also added the following code following L600 of src/main.cc, after the call to `DRAM.operate()`. It examines the read and write queues for each channel in the DRAM. If the element falls between the head and the tail, it should have a nonzero address, that is, it shouldn't be empty. Otherwise, it should have a zero address. 

        bool in_bounds = true;
        bool elem_present = true;
        std::size_t queue_idx;
        for (std::size_t chan = 0; chan < DRAM_CHANNELS; ++chan)
        {
            queue_idx = DRAM.WQ[chan].head;
            do
            {
                if (queue_idx == DRAM.WQ[chan].tail)
                    in_bounds = false;
                elem_present = (DRAM.WQ[chan].entry[queue_idx].address != 0);
                assert((elem_present && in_bounds) || (!elem_present && !in_bounds));

                ++queue_idx;
                if (queue_idx == DRAM.WQ[chan].SIZE)
                    queue_idx = 0;
            } while (queue_idx != DRAM.WQ[chan].head);

[...]

PR #115, opened 2021-01-08
Added delay queue to simplify timing specifications

ngober:
This patch introduces two classes: `champsim::delay_queue`, a sequential container in which items must sit for a fixed time before being marked ready, and `champsim::circular_buffer`, which supports it as the underlying structure.

The delay queue is a natural replacement for the decode and dispatch buffers, as well as the channels of the caches, since these queues only exist to order elements and hold them for a fixed time.

While I was in there, I converted some structures in the core model to use the new circular buffer class, since they fit. The next most obvious place to use this would be the ROB, but changing that would be extensive and probably is worth its own PR. Furthermore, this patch does not completely replace `CORE_BUFFER` and `PACKET_QUEUE`. `CORE_BUFFER` is still used for the ROB, LQ, and SQ. `PACKET_QUEUE` is still used in the DRAM model, because some elements are handled non-sequentially (See #97). Again, this is probably worth its own PR.

These changes are not performance-neutral. I think this comes from the timing at which elements are aged towards becoming ready. If there's serious concern about this, I'll look deeper into the details.


sethpugsley:
I like all of this.  I think there is value in figuring out exactly why performance is so much higher with this PR than in develop, but I think this PR puts us in a much better position to be confident that all buffers throughout the system are handled in the same way.  There's are fewer places that need to be debugged in this PR than in develop.

ngober:
Are there other places you think I should put a `circular_buffer` or a `delay_queue`?

sethpugsley:
I think this should be merged as soon as the conflicts I just created by merging another PR into develop are resolved.

ngober:
Ready for merge.

PR #117, opened 2021-01-19
Allowed CACHE::WQ to merge entries if the queue is full

ngober:
In the cache hierarchy, the write queue was the only queue that could not check itself for duplicates when it was full. This patch allows it to do that. It has some effect on write-heavy programs, since some store instructions that stalled at the head of the ROB due to a full L1D write queue are now able to merge their stores.

Note: This technically violates in-order stores. However, since the L1D read queue checks the write queue for forwarding on insert, I think the conditions for this to be problematic are extremely narrow. If that is a concern, let me know.

### DIFF
--- a/src/cache.cc
+++ b/src/cache.cc
@@ -358,7 +358,9 @@ bool CACHE::filllike_miss(std::size_t set, std::size_t way, PACKET &handle_pkt)
             writeback_packet.type = WRITEBACK;
             writeback_packet.event_cycle = current_core_cycle[handle_pkt.cpu];
 
-            lower_level->add_wq(&writeback_packet);
+            auto result = lower_level->add_wq(&writeback_packet);
+            if (result == -2)
+                return false;
         }
 
         assert(cache_type != IS_ITLB || handle_pkt.data != 0);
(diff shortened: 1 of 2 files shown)


sethpugsley:
Technically, you can only legally do this kind of merging if the two stores you're merging in this way are next to each other in program order, with no other stores to other cache lines in between.

You could achieve this by giving each store a unique, incrementing store_id, and then each WQ entry could keep track of the store_id of the youngest store that has already merged into it (or the one that allocated that WQ entry).  Only allow merging into a WQ entry if the store_id of the write you want to merge in is exactly 1 more than the youngest store_id that's already merged into it (or that allocated it).

ngober:
Ah, I apologize, I was unclear. I worked through a few iterations of this change before settling on this, so I was confused which one I pushed.

The stores may be out-of-order in the cache write queue, but this was already the case due to merging. The retirement from the core's store queue is still strictly in order.

In this patch, if the queue is full and the merge fails, the core gives up and tries again next cycle. Previously, it would give up if the queue could not accept all stores in the instruction, regardless of if any of them could be merged (or if there were space for some of them in the queue).

PR #123, opened 2021-02-02
Separated hit and fill latencies

ngober:
I accidentally introduced a bug in #115 where, due to initialization order, all of the queues in the caches would have latency 0. This patch fixes that by initializing the latencies first. Also, while I was in there, I separated the hit and fill latencies out so that they can be applied separately. By default, they are approximately half of the total latency, with the fill taking longer if the sum is odd.


sethpugsley:
This is good flexibility to have.  Could you please include an updated champsim_config.json to show where these new parameters would fit into the config file?  Other than that, I think it looks good to go.

ngober:
I can, but the way I designed this change, you don't need to specify the hit and miss latencies in the configuration file. You could if you want to, and it would override the defaults, but by default they is calculated from the total latency.

sethpugsley:
Okay, I guess we don't need to modify the config file, but I do want to add that I think the fill_latency should start out as 1 for each level.  We should be able to move the cache up the hierarchy pretty swiftly, in only a few cycles.  1 is maybe a little too low, but it should definitely be in the low single digit range for the L1 & L2.  The LLC could maybe be longer, but I don't have a good number for this yet.

I never really thought about it before this PR, but the fill latency should really only represent how long it takes to move the cache line from one level to the next up the chain, and not how long it takes to actually fill into the data arrays.  Filling the data arrays can be done off the critical path, and filling the data arrays should not slow down the load that originally caused the cache miss.

ngober:
This makes me think of a discussion we had in https://github.com/ChampSim/ChampSim/pull/96#discussion_r541257263. This PR gives us the opportunity to use the fill latency to prevent upward shoot-through. Is there a way these ideas can work together? 

My motivation is that I eventually want to have a single call to `CACHE::operate()`, rather than `CACHE::operate_writes()` and `CACHE::operate_reads()`. I'm working on following through on a desire from when I created the config file to permit heterogeneous cores, and a single operate call makes that easier.

sethpugsley:
In light of this PR, I'm coming around to the idea of reuniting operate_writes() & operate_reads() into a single operate() function.  If you push another commit that combines those, and make the default fill_latency be 1, then I think it'll be ready to merge.

ngober:
I made these changes. I'm happy with this flexibility. When this is merged, it would be worth looking at #106 to see if it affects the results there.

PR #124, opened 2021-02-03
Heterogeneous cores

ngober:
This patch is a huge leap forward towards an idea I've had since introducing the build system last year. Cores are now not required to be identical. For example, a configuration

     {
        "num_cores": 2,
        "ooo_cpu": [
            {
                "rob_size": 100
            },
            {
                "rob_size": 200
            }
        ]
    }

will produce a system with two cores, one of which has a ROB of size 100, the other with a ROB of size 200. This extends to any core dimensions or latencies, as well as cache dimensions and latencies. As before, any unspecified parameter is defaulted. If a different number of cores are specified in "ooo_cpu" than is given in "num_cores", cores will be duplicated (or truncated) to fill out the remaining cores. Thus, the above configuration with "num_cores" equal to 4 would produce 2 of each core.

The cores and LLC both now take a parameter "frequency". They are now permitted to operate in separate frequency domains. The frequency parameter is ignored for private caches, which operate at the same frequency as their core. As a consequence, the global `current_core_cycle` has been removed. The caches and cores now have a local `current_cycle` member that performs the same function.

One exception is that modules cannot yet be heterogeneous. Therefore, the branch predictor, BTB, prefetchers, and replacement policies must be identical across all cores (the prefetchers and replacement policies may still differ by level). Resolving this is beyond the scope of this PR, but can be the subject of a future one.

The model performance is largely unchanged. Small differences do exist, since this does change the sequencing of operations in the main loop. However, I have not seen that the difference is highly significant.


ngober:
As a further note, I believe this change would resolve #63, due to the cleanup of `event_cycle`.

ngober:
I changed my mind and I removed the restriction that private caches operate at the same frequency as the core. The frequency will still default to the frequency of the core. If someone is doing some arcane trickery with clocks out there, they might find this useful. Otherwise, it costs me nothing to include it here.

PR #129, opened 2021-02-17
Increased Flexibility in Cache Configurations

ngober:
_**This patch depends on #124 and should not be merged before it.**_

This patch increases the flexibility of the cache hierarchy. Previously, the cache hierarchy was limited to split I- and D-caches, a shared private L2 cache, and a global LLC. After this patch, caches can be specified arbitrarily under the `cache` key in the configuration file, then referenced from the core configuration. This permits sharing of caches and opens the door for more advanced configurations in the future.

As an example, here are some example configurations:

    {
        "ooo_cpu": [
            {
                "L1I": {
                    "lower_level": "LLC"
                }
            }
        ]
    }

The above configuration bypasses the L2 cache for instructions. Misses from the L1I are sent to the LLC.

    {
        "num_cores": 2,
        "ooo_cpu": [
            {
                "L2C": "shared_l2c"
            }
        ],

        "cache": [
            {
                "name": "shared_l2c"
            }
        ]
    }

The above configuration creates a shared L2 cache between the cores. The dimensions of the L2 cache are defaulted, but could be specified.

There are still limitations to this:

1. The cache hierarchy must still be exactly 3 levels deep at its deepest. This limitation exists because some cache initialization still occurs in the core constructor.
2. The L1I and L1D must still be distinct. This is because the prefetching mechanisms are still hard-coded.
3. The L2C is still assumed to be lower level of the L1D. This is because some initialization occurs externally.
4. Prefetching and replacement policies are determined by core 0. This is, again, because these mechanisms are hard-coded. If we can define a unified interface for prefetchers and replacement, then nearly all of these limitations are alleviated.

This is not a final patch to enable this. It is held back by not having unified interfaces for prefetchers. However, this is a big step forward and a strong improvement to ChampSim.


ngober:
I've just pushed what I hope is the last step in a fully-arbitrary cache configuration. However, this is now a huge patch that changes the back end of ChampSim somewhat considerably.

Notably, I have virtualized the hooks to all modules (branch predictors, BTBs, prefetchers, and replacement policies). These hooks are now passed as parameters in object constructors. See `inc/cache.h` and `inc/ooo_cpu.h` to see what's going on here. I use the `-D` preprocessor flag to change the names of the hooks that users write at compile time, then I include all of the new names in `inc/cache_modules.inc` and `inc/ooo_cpu_modules.inc`. This is the biggest and most complicated change.

This enables a lot of new things. The memory hierarchy is not restricted to be exactly 3 levels deep. The hooks `CACHE::l2c_prefetcher_*` and similar were the biggest culprits of this. I have removed the `if (cache_level == IS_L1D)` and similar statements. Now, a cache doesn't know its level, it just takes the actions we tell it in the constructor to take. As a result, caches can be connected together in arbitrary configurations.

This change does make the configuration a bit more complicated. The configure script attempts to assume as much as it can, but since it cannot assume 3 levels of cache with a shared LLC, it fails to make good assumptions in certain cases. If you are creating arbitrary configurations, you should be prepared to have a more robust configuration file.

ngober:
I apologize for the force-push, but this had diverged a lot, and it ended up being simpler to rebase than to merge.

I have removed the virtualization of the module functions. It turns out that `std::function` generates a whole lot of code, even when it is declared `const`. Furthermore, debugging issues when the function signatures were wrong was nearly impossible and took forever. This is likely to occur when new users are using ChampSim. These two reasons together were enough to make this decision.

Now, the modules are identified by a member of a generated enum (whose names are mangled and use of their names should be considered undefined behavior). The configuration script generates a discriminator function with cascading if-statements to check the module identifier and delegate to the appropriate module function. With `-O3`, this seems to generate pretty small code.

GinoAC:
The new virtualized functions are a solid improvement for further modularizing the prefetcher/replacement policies. It might be challenging to extend modules' arguments, and we should include a short description of how to do so in any future documentation. 

When you're able, please include the updated JSON configuration file. This PR currently does not compile with the current develop branch's configuration file and stalls out if the prefetchers are changed to the correct subdir names.

GinoAC (approved):

PR #144, opened 2021-03-19
Fixing a bug in the computation of the DRAM_DBUS_RETURN_TIME value.

itisntalex:
Hello everyone,

As I am currently doing research on the memory hierarchy and playing a bit with the DRAM, I came across a tiny bug that incurs an underestimation of the data-bus return time (basically the latency from DRAM to LLC on an LLC read/prefetch miss).

The `DRAM_DBUS_RETURN_TIME` variable used to be computed that way:
```cpp
DRAM_DBUS_RETURN_TIME = (BLOCK_SIZE / DRAM_CHANNEL_WIDTH) * (CPU_FREQ / DRAM_MTPS);
```
It turns out, for some combination of values, roundings are happening. For instance: with a CPU at 4 GHz and a DRAM 3.2 GT/s this computation gives you a return time of 8 cycles, which is wrong. The right value is actually 16 as the rightmost term of the product should be rounded to 2.

Thus, this pull request introduces a function dedicated to the computation of the return time and takes care of the rounding. Documentation and comments are also provided.

Hope you'll find this useful. Tell me about your opinions.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -5,6 +5,8 @@
 #include <signal.h>
 #include <string.h>
 #include <vector>
+#
+#include <cmath>
 
 #include "champsim_constants.h"
 #include "dram_controller.h"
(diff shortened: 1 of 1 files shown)


ngober:
This strikes me as overkill. It seems that the issue is a rounding caused by the integer-division operator. Would it work instead just to edit the line to

    DRAM_DBUS_RETURN_TIME = std::ceil(((double) BLOCK_SIZE * CPU_FREQ) / (DRAM_CHANNEL_WIDTH * DRAM_MTPS));

I suspect at that point, the ceiling function and conversion to double may be unneccesary (but they cost nothing, so it's worth it to leave them in.

sethpugsley:
This is how I would usually solve integer division truncation errors:

`float dbus_ret_time = ((1.0 * BLOCK_SIZE ) / DRAM_CHANNEL_WIDTH) * ((1.0 * CPU_FREQ ) / DRAM_MTPS);
DRAM_DBUS_RETURN_TIME  = (uint32_t)dbus_ret_time;`

I think we should not add a new function that's only called one time.

ngober:
Closing after the merge of #147.

PR #145, opened 2021-03-26
Fixed widths of bitmasks

ngober:
Gino brought this bug to my attention today. There was an instance in cache.cc where the upper bits were getting chopped because the bitmask was only 32 bits long. This patch addresses that issue and also fixes an instance in vmem.cc where it might have also been occurring.

### DIFF
--- a/inc/util.h
+++ b/inc/util.h
@@ -10,7 +10,7 @@ constexpr unsigned lg2(uint64_t n)
 
 constexpr uint64_t bitmask(std::size_t begin, std::size_t end = 0)
 {
-    return ((1 << (begin - end))-1) << end;
+    return ((1ull << (begin - end))-1) << end;
 }
 
 constexpr uint64_t splice_bits(uint64_t upper, uint64_t lower, std::size_t bits)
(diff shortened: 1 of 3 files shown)


GinoAC:
Hey Nathan, looks like this is causing a segfault for gcc 8.3.0 on CentOS 7.9.2009 in the bitmask function while running IPC-1's server_003 trace. Appears to be due to a 0 being passed in somewhere.

ngober:
Apologies for the force-push. I pushed untested code and it broke. This should work.

sethpugsley:
I just tested this, and I can't find any problems, but I have a question about the way that bitmask() works.  I see that the inputs to it are of type std::size_t.  How many bits is this, precisely?  Because I don't understand that part, I would be more comfortable if the body of the function said something like this
`return ((1ULL << (begin - end))-1) << end;`
or this:
`return ((((uint64_t)1) << (begin - end))-1) << end;`

Are my fears totally based on nothing?

ngober:
The `std::size_t` type is an arbitrary unsigned integral type that is supposed to represent counting things. It is the type of array indices. I use it when I need to represent the size of something or its position in a sequence.

I pushed a commit that addresses your concern. Better to be safe than squash more bugs later.

PR #146, opened 2021-03-30
Restructured virtual address prefetching

ngober:
I want to submit this change for consideration. It may be a better way to express what is intended by virtual-address prefetching.

This patch reverts the signatures of prefetchers to match the master branch, but adds `"virtual_prefetch"` as a configuration option for caches. When this is set to true, all addresses sent to the prefetcher hooks will be virtual addresses, and all issued prefetches are assumed to be in the virtual space.

With this change, a configuration file that used to say

    {
        "L1D": {
            "prefetcher": "va_next_line_l1d"
        }
    }

will now say

    {
        "L1D": {
            "prefetcher": "next_line_l1d",
            "virtual_prefetch", true
        }
    }

I think this change makes sense because a prefetcher probably should not be mixing virtual and physical addresses. One feature that is lost here is that prefetches within the same page are still penalized the translation time like any other prefetch. If we want to resolve this, I think we need to be sending requests to TLBs.

What do you think? Is this a better way to do VA prefetching?


sethpugsley:
Sorry I haven't made any progress with this stuff in a while, but I think I'm ready to merge this.  One thing I noticed, is that spp_dev's functions have the wrong signatures in this version, so that part will be left hanging, and needs to be fixed with #148.  Please get that done, and I'll merge #148 too ASAP.

PR #147, opened 2021-04-02
Refactor of DRAM model

### LINKED ISSUE #97: Packet Queue removal is non-contiguous

ngober:
I wanted to make a change to subclasses of `MemoryRequestConsumer`: If a packet is passed to one of its member functions, the packet producer is finished working on it and control is completely delegated to the consumer. This was violated notably in the way that the DRAM returned to the LLC. However, this was more difficult than I assumed it would be and I had to restructure the DRAM state machine. Since I was refactoring a large amount of the DRAM already, I went ahead and rewrote the whole thing to resolve #97 and integrate #144. I think this patch solves a bunch of issues and also eliminates a few hundred lines of code.


champsim_config.json
```diff
@@ -158,15 +158,16 @@
         "channels": 1,
         "ranks": 1,
         "banks": 8,
-        "rows": 65536,
+        "rows": 32768,
```
GinoAC, line 161: This should stay at 64K rows.

ngober: Fixed.

GinoAC (approved):
No major functional changes compared to what is currently in the develop branch, so there's no change in performance.

PR #148, opened 2021-04-05
Updating SPP to the upstream repository

ngober:
I've done a lot of parallel work to update SPP. This fixes a whole lot of bugs that we've discovered over the years and updates it to use more contemporary C++. It should be a lot more readable now, plus it showcases the way to split a design across multiple files.


ngober:
@sethpugsley Pinging you to make sure you see this update.

sethpugsley:
I'm running a test with all of the IPC-1 public traces, which will finish in a while, but I did immediately notice that for bandwidth-constrained traces, the new SPP is much worse than the old one.  For example, on some mcf traces I'm seeing performance losses of almost 30%.  Is this a surprise, or did you and @GinoAC already know about this with this version of SPP?

sethpugsley:
I just compared spp_dev to spp_upstream in a configuration using the develop branch with hashed_perceptron branch predictor, and next_line_l1i and next_line_l1d prefetchers, on the IPC-1 public traces, running for 50M warmup, and I'm seeing a geomean "speedup" of .834x, meaning a 17% reduction in IPC.  Can anyone else confirm?

Issue #162, opened 2021-05-11
Develop Branch - Simulating 2MB pages


gvavou5:
I want to simulate a baseline system with 2MB pages and compare it against a system with 4KB pages. To do so, I create a new configuration which is the same as the provided one except that the page size is 2097152, instead of 4096. 

I have some questions regarding the virtual memory implementation.

1. in the function PageTableWalker::get_offset (ptw.cc file) you declare shift = 12; I believe that it should be declared as LOG2_PAGE_SIZE. The same applies for the function PagingStructureCache::get_index (ptw.cc file).

2.  in the function PageTableWalker::get_offset (ptw.cc file)  you declare uint64_t offset = (full_virtual_addr >> shift) & 0x1ff; I believe that the mask should take into account the page size. Thus, 0x1ff if page size = 4kb and 0xfff if page size = 2mb.

Is it correct what I am saying or do I miss something in the implementation?

ngober:
I've proposed #172 to address this. In it, it properly applies `LOG2_PAGE_SIZE` and the masking bits.

ngober:
Resolving after the merge of #172.

Issue #131, opened 2021-02-18
Simulating different page sizes 


gvavou5:
Hello, 

I would like to see the impact of different page sizes. To do so, I did the following experiments: (i) page size is 4KB, (ii) page size is 2MB. Note that **both experiments don't use prefetching at any cache level** and when I consider 2MB pages I properly changed both PAGE_SIZE and LOG2_PAGE_SIZE in inc/champsim.h. 

Intuitively, 2MB pages would only improve performance by saving TLB misses (i.e., page walks) because the current version of ChampSim **does not have** a page table walker that stores entries from all page table levels into the caches. Hence, the cache MPKI rates should be the same for both experiments explained above.

My evaluation verified that there is no change in the cache MPKI rates when using different page sizes (intuitively correct) but I observed that when using 2MB pages the average cache latencies are much higher than the ones reported when 4KB pages are used. 
For instance, the L2C average miss latency of trace 410.bwaves_1963 is 296cc and 565cc when using 4KB and 2MB pages, respectively.
I don't see why such behavior is normal. The average latencies should be the same. Am I wrong? Does anyone know anything about it?

** I am referring to the master branch **

ngober:
This appears to be resolved in #172.

I ran a similar test to you. I ran 410.bwaves-1963B for 10M/100M and I see very similar latencies and MPKIs for all of the data caches.

ngober:
Resolving after the merge of #172.

PR #172, opened 2021-08-11
PTE/VMEM refactor to support large pages

ngober:
This patch modifies the page table walker and the virtual memory. The primary intent is to remove the assumption that pages are 2^12 bytes in size and that there are 512 PTE entries per PTE page. Now, these two classes work more closely together.

This probably addresses #131 and #162. I'm curious if it does fix them.

There is a change in performance that results from this. It seems to come from a bug that I fixed where the PTW appeared to make one more step down the walk than it was supposed to.

### DIFF
--- a/inc/vmem.h
+++ b/inc/vmem.h
@@ -1,7 +1,7 @@
 #ifndef VMEM_H
 #define VMEM_H
 
-#include <iostream>
+#include <cstdint>
 #include <deque>
 #include <map>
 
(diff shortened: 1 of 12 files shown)


inc/vmem.h
```diff
@@ -1,37 +1,40 @@
...
 #define VMEM_RAND_FACTOR 91827349653
 // reserve 1MB of space
 #define VMEM_RESERVE_CAPACITY 1048576
 
+#define PTE_BYTES 8
+
+//Virtual Address: 57 bit (9+9+9+9+9+12), rest MSB bits will be used to generate a unique VA per CPU.
+//PTL5->PTL4->PTL3->PTL2->PTL1->PFN
+
+constexpr uint64_t MINOR_FAULT_PENALTY = 200;
```
GinoAC, line 17: This seems like another knob to add to the configuration file later on.

ngober: I chose not to add it because I considered it an estimate used for testing purposes, but it just takes 4 additional lines to make it configurable. Should I do that now, or leave it until later?

GinoAC: That's valid. Does this value change in any other PRs to be based on the memory system latencies or is it staying static?

I mainly left this to remind myself to add it post-review, but you can add it now if it doesn't take too much effort.

ngober: This is a static penalty. If we want to make it more dynamic or more realistic, we would need to actually build out some more proper support.

ngober: I went ahead and made it configurable. I figure, why not?

GinoAC: Thanks! Yeah, I don't see how it could hurt.

GinoAC (approved):

PR #184, opened 2021-12-11
Do not allow secondary miss to go to prefetcher or add appropriate statistics

AbhishekUoR:
This miss information is already sent to the prefetcher in the past; as Champsim does not have an MSHR structure that tracks down secondary misses for the same cache line, we should not train prefetchers on this miss. If we want to train prefetchers on this, we need to add the miss and access stats that is "sim_access" and "sim_miss".


ngober:
Thanks for the pull request. Could you please base it on the develop branch?

ngober:
I'm closing this PR since it has fallen significantly behind development.

PR #190, opened 2022-01-14
Update IP-stride prefetcher

ngober:
This patch updates the included IP-stride prefetcher to leverage and demonstrate good usage of updated prefetcher interfaces.

### DIFF
--- a/prefetcher/ip_stride/ip_stride.cc
+++ b/prefetcher/ip_stride/ip_stride.cc
@@ -1,5 +1,6 @@
 #include <algorithm>
 #include <array>
+#include <map>
 
 #include "cache.h"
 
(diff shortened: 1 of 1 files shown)


GinoAC (approved):
The major change to this, aside refactoring, is the move from a direct-mapped metadata structure to a 4-way associative structure. The updated version experiences slightly lower performance improvements. This is most likely due to the refactor issuing prefetches on a per-cycle basis instead of issuing multiple prefetches per cache access. I would argue this degradation is a movement to a more realistic model.

ngober:
This comment doesn't matter, but I'm putting it here for public record. The change in associativity you mentioned was actually a move from a fully-associative structure to a 4-way associative structure. The performance dip is probably due to a combination of the associativity change, which deflates performance in highly collision-prone workloads (a hashed index might help here), and the per-cycle issue, which likely only deflates performance in situations where the path is terminated sooner than it would have been.

PR #193, opened 2022-01-17
Refactored perceptron branch predictor

ngober:
This patch refactors the perceptron branch predictor. I removed pointer arithmetic in favor of index arithmetic, and used the opportunity to demonstrate how users might create a class to encapsulate module functionality.

### DIFF
--- a/branch/perceptron/perceptron.cc
+++ b/branch/perceptron/perceptron.cc
@@ -34,280 +34,126 @@
  * a hardware budget of (24+1)*8*163 = 32600 bits, or about 4K bytes,
  * which is comparable to the hardware budget of the Alpha 21264 hybrid
  * branch predictor.
- *
- * There are three important functions defined in this file:
- * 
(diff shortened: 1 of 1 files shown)


alberto-ros:
I noticed in the branch predictor codes that always_taken is not used at all. But the branch predictor is called for all branches and therefore the predictor can predict not_taken for an inconditional branch, and overriding the BTB decision. Is this an intended behavior?

ngober:
That's a good point. The value isn't ever used. But, the branch is considered taken if either the branch predictor or the BTB predicts the branch is always taken.
https://github.com/ChampSim/ChampSim/blob/fb87c5daaf3b637363c9e95483999852cf291edd/src/ooo_cpu.cc#L235

Perhaps for performance, it makes sense to prevent updates based on whether the branch is always taken. I'm not immediately sure how we would accomplish that without changing the interface, but it sounds like a good idea.

alberto-ros:
I see. You are right :-)

PR #195, opened 2022-01-18
Added deprecation warnings

ngober:
This patch adds back support for `current_core_cycle`, the former style of determining which cycle the simulator was at. It does this by adding an array-like class that emits a warning and redirects to the core's `current_cycle` member.

This also deprecates `CACHE::prefetch_line(ip, base_addr, addr, fetch_level, metadata)` in favor of `CACHE::prefetch_line(addr, fetch_this_level, metadata)`. The first now redirects to the second, and emits a warning.

There is also a lot of cleaning up of vestigial code and `#includes` that weren't required.


GinoAC (approved):
I was able to get the deprecation warning to pop up and I don't see any side effects from the changes. Looks good to go!

PR #199, opened 2022-01-24
If RAS is empty return 0 as predicted address and do not add instruction size

alberto-ros:
Minor update not to believe that the RAS returns a valid target.


alberto-ros:
This push requests becomes done after #225

PR #203, opened 2022-02-01
Moved collision detection into the operation step of MemoryRequestConsumers

ngober:
This patch pulls the collision detection of `PACKET`s in `MemoryRequestConsumers` into the operation step. Now, packets which are added are not immediately scanned for their collisions, and so can never be immediately (before the `add_*q()` function returns) returned. All packets, provided the underlying allocation succeeds, will spend at least some time in the queue.

This has some advantages. We can now avoid an antipattern in `CACHE` ([here](https://github.com/ChampSim/ChampSim/compare/develop...ngober:virtual_prefetch?expand=1#diff-f12a80a9b06758711decb7a6dadca0e5d80e28206c03953ca5e64551130254f0R297)), where we check the size of the underlying consumer's queue, allocate an MSHR if there is space in the underlying cache, and only then sends the forwarding request. Now, the underlying consumer does not need to advertise its queue capacities. Instead, it returns a boolean concerning whether or not any allocation actually occurred.

This arose out of a larger work that I'm working on, concerning translation in caches, but the change I'm making started getting systemic, so I'm sending it in multiple parts.


ngober:
I've discovered some issues with this patch, so I'm breaking it into smaller parts and submitting them separately.

PR #208, opened 2022-02-10
Remove lint trigger for pull requests.

ngober:
The `add-and-commit` action that we have been using to automatically format commands doesn't work when the remote branch is a fork. To accommodate this, I've removed the triggering condition for pull requests. Instead, it will run when the pull request is merged.

### DIFF
--- a/.github/workflows/buildcheck.yml
+++ b/.github/workflows/buildcheck.yml
@@ -11,7 +11,9 @@ on:
       - '.github/workflows/*.yml'
       # re-include current file to not be excluded
       - '!.github/workflows/buildcheck.yml'
+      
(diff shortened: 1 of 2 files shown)


ngober:
I added a new workflow that builds a default configuration and runs a single trace on it.

There are definitely some problems with this that should be resolved:
  - Only one trace is run. We should decide on a representative sample of traces to run
  - The trace is remotely accessed at the pleasure of Dr. Jimenez. This is not sustainable long-term.
  - Because the output of ChampSim remains difficult to parse. Until it becomes easier to parse, regression checks will be difficult.

Nevertheless, I think that this is a fine start.

PR #209, opened 2022-02-14
Change core, cache, and PTW to use builders for construction.

ngober:
This patch doesn't change any of the functionality of the model. It is more of a provision for future development. There are no user-visible changes.

The constructors for the core model, cache model, and PTW model were bulky and difficult to debug when something changed. This change makes the arguments to the constructor reorderable, defaultable, and documented. It also moves a considerable amount of the processing of default parameters into the C++ compilation step, where previously it was in the Python configuration step.


ngober:
I've returned to this in order to update it. Certain parts of it had become broken and hard to repair. However, after rebasing atop #303, it has become a lot more possible to achieve this patch. I will leave it as a draft until its dependents get merged.

PR #210, opened 2022-02-16
Move instruction prefetcher to the cache model

### LINKED ISSUE #196: Use of NUM_SET and NUM_WAY in instruction prefetchers
https://github.com/ChampSim/ChampSim/commit/fb87c5daaf3b637363c9e95483999852cf291edd#commitcomment-64121863

@alberto-ros I'm migrating the discussion here so that it has more visibility. CC: @sethpugsley @GinoAC 

The instruction prefetchers are members of O3_CPU and have been since the initial designs. However, when IPC1 was run, the prefetchers had access to the global constants `L1I_NUM_SET`, `L1I_MSHR_SIZE`, etc. Those constants were removed in favor of `CACHE::` member variables. However, this makes `CACHE::NUM_SET` is inaccessible from the instruction prefetchers.

How can we support instruction prefetchers like this? A short-term solution could be to add `O3_CPU::L1I_MSHR_SIZE`, but this isn't a good idea for the future. Alberto suggests making the instruction prefetchers members of `CACHE` instead, but this would break compatibility with old code, and also I'm not sure what to do then about `O3_CPU::prefetcher_branch_operate`.

ngober:
Resolves #196.

This patch moves the instruction prefetcher to become a member of the cache model, rather than the core model. This allows the instruction prefetcher to access information about the cache, such as the number of sets and the number of ways.

Instruction prefetchers must still specify a branch hook, which still distinguishes the instruction prefetchers from the data prefetchers. I think that this constraint can eventually removed, but that is beyond the scope of this patch. One thing I do like about this patch is that it removes a whole lot of special-case code from the configuration script.


sethpugsley (approved):
I've already looked at this before, and I just tested it out, and it seems to work well.

GinoAC (approved):
The code looks good, and there aren't any performance side-effects, so it looks good to go to me.

ngober:
I'm merging this into the development branch and it will go out with the next release. That may be a few months or a year, but this way it can be used.

PR #212, opened 2022-02-28
Zero-cycle elements in the delay_queue get ready instantly.

alberto-ros:
Right now, the delay queue is updated with 1-cycle delay.


ngober:
I think this change looks good. I'm a bit hesitant to push this to master immediately, since I know there are students using this for class projects and I don't want to accidentally ruin their baseline performance numbers. Is it better to push this to develop and merge it in sometime this summer? I'm not sure.

alberto-ros:
Pushing it to develop for now should be okay.

PR #214, opened 2022-03-03
Used std::vector in instruction model

ngober:
This began as a necromancy of #46, but the work that has been done since then has unlocked a lot more that it was able to do.

This patch converts the register and memory usage information to `std::vector`s. This is an important change, since any future trace format might have arbitrary numbers of instruction sources or destinations. Knowing how many destinations and sources an instruction might have is no longer necessary in the simulator. This also converts many instances that performed repeated zero checks (where zero represented, say, no register access), and so is less branchy.

This patch also simplifies the register dependency tracking. We no longer iterate through the ROB for each instruction looking for data dependencies. Now, they are maintained in a separate structure. This gets a moderate speed benefit because the ROB is the largest structure in the core model.

As I was doing this work for registers, I tried to do the same thing for memory. In particular, I wanted to remove `ooo_model_instr::lq_index` and `ooo_model_instr::sq_index`, since they contained a lot of pointers to nowhere. Now, if the source or destination memory vectors contain an entry, it contains a pointer to an actual entry in the load or store queues (sort of, see my next PR).

This also enormously simplifies the memory scheduling logic, since now we can just use all of C++'s iteration logic instead of wrangling indices. All in all, this removes 200 lines of code, which is a much simpler codebase. The core model is almost less than 1000 lines long!


GinoAC:
I ran this on a subset of CVP traces, and it appears that RFO misses from the L1D are not propagating to the L2 and higher. Loads from the L1D and L1I appear unaffected by the bug.

ngober:
Wow, that's a sneaky error, but thankfully an easy fix. That problem came from a copy-paste issue. Thanks for catching that.

GinoAC:
Looks good now! I'll go ahead and merge it.

PR #215, opened 2022-03-03
Decoupled store completion

ngober:
This patch depends on #214. It should be merged first.

This patch decouples store completion from instruction retirement. Now, the load queue, store queue, and ROB are allocated in the dispatch step. Allocated load/store entries are marked ready at instruction execution (address calculation), when they may begin translation and the loads are also fetched. Stores are marked to be issued whenever their corresponding instruction is retired, and they can be sent whenever the queue has bandwidth.

This is supposed to make ChampSim more like real machines, but it has a big change in performance. Hopefully, that's a good thing.


ngober:
This PR is ready to go. I found a couple bugs in #214 and rolled fixes for them in here. It produces a geomean 1.34x speedup on the IPC1 traces.

GinoAC:
I like these changes overall. My only suggestion would be to add some comments to the dependency merging in src/cache.cc so its clearer what's going on. Otherwise, it looks good to go to me.

GinoAC (approved):
I like these changes overall. My only suggestion would be to add some comments to the dependency merging in src/cache.cc so its clearer what's going on. Otherwise, it looks good to go to me.

Issue #213, opened 2022-03-02
ITLB and L1I access


alberto-ros:
¡Hi everyone! 

I noticed two features in the implementation of the fetch stage which raised some questions to me.

The first one is that the L1I is accessed after the ITLB offers the translation. This way the L1I is modelled as a physically indexed, physically tagged (PIPT) cache. Would not be more realistic to model a virtually indexed, physically tagged (VIPT) cache? That is, to overlap the ITLB access with the L1I access.

The second one is that the implementation assumes that the ITLB and the L1I have only one read port. My question here is if this is the standard in current processor, e.g., Sunny Cove (as the default parameters try to mimic that processor).

Thanks always for your time!
Alberto

ngober:
This is something that I'm working on right now. It has been a problem for some time that the TLBs and the caches operate in series. It's taking a bit of work to separate the two. Hopefully I will have something to PR in the next couple weeks.

Your second point about there only being one read port is a good point, and I hadn't considered it. I'll make that a next step after the parallel TLB access.

ngober:
The parallel TLB access is enabled with #227.

GinoAC:
I'm going to go ahead and close this since it looks like the issues are resolved in #227 and #217.

PR #217, opened 2022-03-07
Allow multi-ported ITLB and L1I cache

alberto-ros:
Right now only one access to ITLB and L1I are performed per cycle. However, the default config file indicates that the ITLB and the L1I have two read ports. This pull request, related to issue #213, resolves the issue.


ngober (approved):
This looks good. It's a minimal set of changes that does the job. There's probably a more elegant way to do this, but it might require something like C++20 views or whatever.

GinoAC (approved):

PR #224, opened 2022-03-11
Reset taken for non-branch instructions

alberto-ros:
For some reason, some non-branch instructions have the branch_taken field set. They should reset it.

This does not affect performance now.


ngober (approved):
This is probably indicative of an underlying problem with the tracer, but after looking through it, it doesn't look like the value is uninitialized. We'll have to make a note to check this out later.

PR #225, opened 2022-03-11
BTB and branch predictors are called for all instructions and just us…

alberto-ros:
Right now the BTB and the branch predictor are called only for branches. Also they pass information it is unknown that the point they are called. 

This push request tries to have a more realistic prediction, but not using the information that it is not known, for example, if the instruction is a branch or not.


branch/bimodal/bimodal.cc
```diff
@@ -14,7 +14,7 @@ void O3_CPU::initialize_branch_predictor()
   bimodal_table[this] = {};
 }
 
-uint8_t O3_CPU::predict_branch(uint64_t ip, uint64_t predicted_target, uint8_t always_taken, uint8_t branch_type)
```
ngober, line 17: I'm not certain that removing this information from the module hook is a good thing. Sure, it may not be the case that all branch predictors know things like the predicted target, but I don't think it's a far-fetched idea that someone might want to do that. If a researcher doesn't want their branch predictor to be aware of the information, they can always just ignore it.

ngober (changes requested):
I like the changes made to `src/ooo_cpu.cc`. Running the branch logic on each instruction rather than on only branches is a good and realistic change. However, I have some reservations about changing the function signatures.

I'd be curious to hear others weigh in on this. I think this is a change worth considering.

branch/bimodal/bimodal.cc
alberto-ros: I am good with keeping those parameters in the call. No predictor uses that info now, but I am OK with keeping them.

For predicted_target and always_taken, I would expect the branch_predictor to be called a cycle after BTB access, but this is not the case.  Maybe we can add a comment there :-)

djimeneth: I haven't read through the code but rather the comments. In my opinion running the branch predictor on every instruction is maybe a little more intense than it needs to be. In a real system, it's true that when an instruction is fetched, we don't know whether it's a branch so we have to assume it is. But the situation is a little more complex. We would fetch N instructions per cycle (as a "fetch group" corresponding to some fraction or multiple of icache blocks) and then start a branch direction lookup for M instructions, where M could be less than N. We would have a bookkeeping structure somewhere that keeps track of which instructions in a fetch group are branches and have M "buckets" per  fetch group. For example, the Samsung Exynos M family of SoCs (for which I developed the branch predictor) could predict up to 8 branches per cache line of 32 instructions; if there were more branches than that in a cache line we did something special. On the other hand, the proposed EV8 branch predictor allowed every branch in a group of (as I recall) 16 instructions to be predicted using a clever organization of the prediction tables, but that would not have been possible without that organization, and certainly would not be possible with modern TAGE or perceptron predictors because of the complexity. There are two extremes: predicting only branches and predicting all instructions, and industry practice today is somewhere in the middle. I don't think we want to go as far as dictating a mechanism for identifying branches (e.g. the Samsung idea) within fetch groups for ChampSim. Thinking about it some more, I think the best thing is to just invoke the predictor on known branches, even though that is unrealistic. It should not have a significant impact on performance modeling; after all, the predictor will only be updated with real decoded branches so the accuracy should be the same. We have to think, what is the goal of this thing? We have said that ChampSim is a tool that should be easy to get started using, good for education and contests. That goal is at odds with having it be accurate as we can see in this case. Frankly, having the branch predictor invoked on every instruction would get in the way of, say, a branch prediction contest, necessitating competitors to build structures for disambiguating branches from their limited hardware budget when they should be innovating new algorithms.

djimeneth: Also, the branch direction predictor should be accessed in parallel with the BTB, not a cycle after. Delay in looking up the direction predictor is intolerable; in some processors the access to the SRAM arrays begins speculatively before the block address is even known; my HPCA 2003 paper proposes this idea (also see André's "ahead pipelining" idea) and after my presentation someone from Sun came up to me and said "we're already doing that." 

Now again, it's more complicated than just that. The BTB also knows whether a branch has ever been taken or not taken. If there is no BTB entry, we should override the direction predictor and predict "not taken." If there is a BTB entry and the entry shows no evidence that the branch has ever *not* been taken (through a special metadata bit in the BTB entry) then we override the direction prediction and predict taken. When we get to update the direction predictor, we can again use these facts to determine whether we really want to update the direction prediction tables for a branch that has ever only shown behavior in a single sense. This has the effect of filtering a lot of updates out of the direction predictor. I don't know if we're doing that in ChampSim but we should.

alberto-ros: Thanks Daniel for your detailed comments. In fact this PR goes in that direction. Important to say also that it does not change IPC of traces that do not require too much BTB capacity. 

A major problem that I see now in the code is that the BTB uses the branch type to go to the BTB, Indirect target predictor or RAS. 

I will summarize here the changes:

The BTB stores now information for any (taken at least once) branch (including indirect and returns, even if the target  prediction is provided by the Indirect target predictor or the RAS, as they may be not tagged). This incurs just a small increase in BTB pressure. The BTB is accessed for any instruction and the info stored in the BTB is used for finding the target, either from the BTB, Indirect or RAS. 

The branch predictor is accessed in parallel to the BTB (for any instruction, but this does not affect IPC). If no target is found in the BTB, the final branch decision will be not taken, anyway. And if the branch is recognized by the BTB as always taken (e.g., inconditional), the final branch decision will be taken, no matter what the predictor says.

Both the BTB and the branch predictor are updated as done now, that is, only for branches.

ngober (approved):

ngober:
I've finally gotten around to thinking about this change and I think I'm in favor now, particularly after the comments on #256. The change to the module interfaces is probably not very significant, I suppose it represents a pretty small fraction of our userbase. Furthermore, all of the information that was removed from the prediction hooks is still present in the update hooks.

Issue #112, opened 2021-01-02
CVP Trace converter


aperais:
Hello,

This is not a ChampSim issue per se, but I wanted to start a discussion regarding the use of CVP traces in ChampSim. That's definitely encouraged, but we stumbled upon a fairly big caveat when developing new features for CVP2 and so I wanted to know if it would be possible for me to PR a disclaimer to cvp_trace/README.md with that caveat. I am also planning to have a README with the same caveat where the traces are hosted and on the CVP website.

Long story short, because we do not have the addressing mode in the (originally Aarch64) traces, we treat the pre/post increment mode base register update as a "memory" destination register and not an "ALU" destination register. For instance

`ldr x1, [x2, #16]!`

Should load 8 bytes from memory with whatever the memory latency is, and should put in x2 the previous value of x2 plus 16. If the infrastructure is not aware of that, both destination registers are generated at the same cycle, and notably, if there is a miss say at LLC, then x2 will become ready quite late, when in reality it should have been ready soon after the effective address was computed. Moreover, if the infrastructure is not aware of it, it might assume that 2 x 8 bytes have to be loaded, making the data footprint incorrect. This is another drawback of the CVP trace format where the "memory size" field is per item, and it is assumed that there are as many items as destination registers. Unfortunately, for Aarch64, that is not the case depending on the addressing mode. This issue compounds with the fact that the trace does not immediately allows to differentiate between a load pair and a load single with base update (both have two destination registers). 

We have added some code in the CVP2 to deal with that issue, but it is unclear that ChampSim or the trace converter perform this kind of analysis, and it seems important that everyone using the CVP traces in their current form with ChampSim is bound to see incorrect performance (and data side memory footprint).

aperais:
For all intent and purpose this is somewhat similar to x86 pop (load + update base register). I don't think it would be as much of an issue (doubt pop is used within a loop much, but I may be wrong) with x86 but it may be worth checking that this is handled correctly for x86 as well.

aperais:
Hello,

For your reference, I have been overhauling the CVP trace reader ease reusing CVP traces in other frameworks. The reader is currently located at : [https://gricad-gitlab.univ-grenoble-alpes.fr/tima/sls/projects/cvptracereader](https://gricad-gitlab.univ-grenoble-alpes.fr/tima/sls/projects/cvptracereader)

[...]

PR #227, opened 2022-03-16
Parallel TLB lookup

ngober:
This patch performs TLB lookup in parallel with the cache tag lookup, which has been requested often. It also adds full TLB lookup support for virtual-address prefetching.

The queues in the cache model are now themselves capable of generating requests to the TLBs. The TLBs have been removed from the core model. There are two kinds of cache queues: a `CACHE::TranslatingQueues` and a `CACHE::NonTranslatingQueues`, which are instantiated at configure time.

A consequence of this is that virtual-address prefetches can now be passed to the TLBs. There is now an additional configuration option: `"lower_translate"`.

    {
        "STLB" : {
            "name": "my_stlb"
        },
        "L2C" : {
            "prefetcher": "va_ampm_lite",
            "virtual_prefetch": true,
            "lower_translate": "my_stlb"
        }
    }

Most of the time, the default `lower_translate` specification should be sane, but you should pay attention to this if you are making your own designs.

This results in a pretty considerable performance increase.


alberto-ros:
Does the commit misses the update to champsim_config.json?

Should it be:

 "lower_translate": "cpu0_STLB"

Or:

 "lower_translate": "STLB"

?

ngober:
I think the example I gave wasn't great. I've edited the example configuration above. In general, the default names for TLBs and caches aren't something that the user knows or should try to name. If you want to pick a particular `lower_translate`, you should have already given the cache a name.

I added a commit that will make a core's STLB the default translation supplier for the L2C. I'm not 100% sure this is the best default. I've heard arguments that the L2C should have its own private TLB, I've heard some people say that it should go to the DTLB. In my view, if you want to make changes, those changes are available to you, but here are the defaults:

 - L1I -> ITLB
 - L1D -> DTLB
 - L2C -> STLB
 - LLC -> ??? (If you specify this, I hope you know what you're doing)

alberto-ros:
Not I see how it works and I like it. Defaults seem okay to me. Even having a virtual prefetcher at L2C may be non-conventional.

alberto-ros:
I found out a problem with this patch. Prefetchers do not work because of the following code at prefetch_line:

  if (fill_this_level)
    pf_packet.to_return = {this};

Prefetches should never be returned to the cache that generates them since and MSHR has not been allocated for them at the time of inserting them in the PQ.

So commenting those lines of code fixes the issue, but it neglects the fill_this_level feature.

alberto-ros:
The other checks that I did look good :-)

ngober:
I managed to write a test that isolated the issue and then fixed it. Thanks for bringing it to my attention.

alberto-ros:
It seems that still there are some bugs, as I encountered a deadlock.

I just compile the code with the default options and run:

./bin/champsim -warmup_instructions 20000000 -simulation_instructions 20000000 -traces secret_int_118.champsimtrace.xz 

The trace is obtained from: https://hpca23.cse.tamu.edu/CVP-1/secret_traces/index.html

If you cannot replicate it and you want the output of the deadlock let me know it.

ngober:
I'd be curious about this deadlock, but I'm a bit hesitant to pursue it differently since we've had some trouble with the CVP traces in the past (#112). I suspect this is a problem that would actually require us to make a new trace format to solve. I am actively working on that, and I'd be open to suggestions if you have any, but that's for another thread.

alberto-ros:
This is a new deadlock and not related to #112, since #112 is a performance bug. I would not relate both things, as this is I guess some bug introduced in this PR (appeared in a CVP trace because I am testing them, but it could appear in another trace).

PR #237, opened 2022-04-06
Fix a minor bug where tracereader reads twice the same instruction on…

alberto-ros:
… initialization


ngober:
Could you look at #185? I think I may have fixed this in that patch.

alberto-ros:
I couldn't compile it and run it, but looking at the code it seems it should be fixed.

ngober:
This appears to have been resolved in #185.

PR #238, opened 2022-04-06
Updated PIN tracer

ngober:
Intel recently moved the PIN documentation and broke the link that was in the README. I used the opportunity to test the tracer with the latest version (3.22), update the documentation, and reorganize the tracer directory.

I moved the PIN tracer under `tracer/pin/`, with the intention that we might build out other example tracers in the future. I'm looking at DynamoRIO, but who knows?


GinoAC:
I have to generate some new traces so I'll test this out sometime in the next week.

PR #240, opened 2022-04-13
Added an option to run trace(s) once

cpepi001:
This may be helpful to run tiny/small trace(s) assuming low/infinite warmup/simulation instructions

### DIFF
--- a/inc/instruction.h
+++ b/inc/instruction.h
@@ -67,6 +67,8 @@ struct cloudsuite_instr {
 };
 
 struct ooo_model_instr {
+  bool drained = false;
+
   uint64_t instr_id = 0;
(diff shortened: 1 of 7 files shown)


ngober:
Is this similar to [this commit](https://github.com/ChampSim/ChampSim/pull/185/commits/5c2f7435829c16c9d6b4f76cbaf54f6a0f176b31)? I recently found a similarity between #185 and #237 that made me consider merging #185 first. If you think these are similar, then I'll merge #185 and we can work this change into it.

cpepi001:
I am not really sure, but also this commit needs to handle the inflight instructions as for now when the trace ends it returns

PR #252, opened 2022-07-18
Bug Fix in cache.cc

zeal4u:
THANKS a lot for this project that helps me a lot in researching methods for cache performance improvement.

Recently, I found the program can cause a deadlock in a corner case.

When there are an instruction prefetch `P` in L1I and a load `L` in L1D with the same physical address, `L` is issued later. 
Their requests should be merged in L2C MSHR.
Due to the following codes in `cache.cc`:
``` 
    if (mshr_entry->type == PREFETCH && handle_pkt.type != PREFETCH) {
      // Mark the prefetch as useful
      if (mshr_entry->pf_origin_level == fill_level)
        pf_useful++;

      uint64_t prior_event_cycle = mshr_entry->event_cycle;
      *mshr_entry = handle_pkt;

      // in case request is already returned, we should keep event_cycle
      mshr_entry->event_cycle = prior_event_cycle;
    }
```
the request in L2C MSHR of `P` will be REPLACED by `L`'s request, not merged as supposed.

In the corner case,  the data is returned to L1D only, and never filled to L1I, resulting in a deadlock.

For reproducing the deadlock, I also pushed the codes of prefetcher EIP that cause the deadlock.
Many traces can be used to reproduce the bug, e.g., `srv_408.champsimtrace.xz` provided by [https://zenodo.org/record/5517143#.YtVa4ptByV4](url)

BTW, I modified `config.sh` to name a meaningful binary file, :).

### DIFF
--- a/prefetcher/EIP/EIP.cc
+++ b/prefetcher/EIP/EIP.cc
@@ -1,0 +1,871 @@
+////////////////////////////////////////////////////////////////////////
+//
+//  Code submitted for the First Instruction Prefetching Championship
+//
+//  Authors: Alberto Ros (aros@ditec.um.es)
+//           Alexandra Jimborean (alexandra.jimborean@um.es)
+//
+//  Paper #30: The Entangling Instruction Prefetcher
+//
+////////////////////////////////////////////////////////////////////////
+
+#include "ooo_cpu.h"
+
+#define L1I_PQ_SIZE 32
+#define L1I_MSHR_SIZE 8
+#define L1I_SET 64
+#define L1I_WAY 8
+
+#define DEBUG(x) 
+
+// To access cpu in my functions
+uint32_t l1i_cpu_id;
+
+uint64_t l1i_last_basic_block;
+uint32_t l1i_consecutive_count;
+uint32_t l1i_basic_block_merge_diff;
+
+// LINE AND MERGE BASIC BLOCK SIZE
+
+#define L1I_MERGE_BBSIZE_BITS 7
+#define L1I_MERGE_BBSIZE_MAX_VALUE ((1 << L1I_MERGE_BBSIZE_BITS) - 1)
+
+// TIME AND OVERFLOWS
+
+#define L1I_TIME_DIFF_BITS 20
+#define L1I_TIME_DIFF_OVERFLOW ((uint64_t)1 << L1I_TIME_DIFF_BITS)
+#define L1I_TIME_DIFF_MASK (L1I_TIME_DIFF_OVERFLOW - 1)
+
+#define L1I_TIME_BITS 12
+#define L1I_TIME_OVERFLOW ((uint64_t)1 << L1I_TIME_BITS)
+#define L1I_TIME_MASK (L1I_TIME_OVERFLOW - 1)
+
+uint64_t l1i_get_latency(uint64_t cycle, uint64_t cycle_prev)
+{
+  uint64_t cycle_masked = cycle & L1I_TIME_MASK;
+  uint64_t cycle_prev_masked = cycle_prev & L1I_TIME_MASK;
+  if (cycle_prev_masked > cycle_masked) {
+    return (cycle_masked + L1I_TIME_OVERFLOW) - cycle_prev_masked;
+  }
+  return cycle_masked - cycle_prev_masked;
+}
+
+// ENTANGLED COMPRESSION FORMAT
+
+#define L1I_ENTANGLED_MAX_FORMATS 7
+
+// HISTORY TABLE
+
+#define L1I_HIST_TABLE_ENTRIES 1072
+#define L1I_HIST_TABLE_MASK (L1I_HIST_TABLE_ENTRIES - 1)
+#define L1I_BB_MERGE_ENTRIES 4
+#define L1I_HIST_TAG_BITS 58
+#define L1I_HIST_TAG_MASK (((uint64_t)1 << L1I_HIST_TAG_BITS) - 1)
+
+typedef struct __l1i_hist_entry {
+  uint64_t tag;       // L1I_HIST_TAG_BITS bits
+  uint64_t time_diff; // L1I_TIME_DIFF_BITS bits
+  uint32_t bb_size;   // L1I_MERGE_BBSIZE_BITS bits
+} l1i_hist_entry;
+
+l1i_hist_entry l1i_hist_table[NUM_CPUS][L1I_HIST_TABLE_ENTRIES];
+uint64_t l1i_hist_table_head[NUM_CPUS];      // log_2 (L1I_HIST_TABLE_ENTRIES)
+uint64_t l1i_hist_table_head_time[NUM_CPUS]; // 64 bits
+
+void l1i_init_hist_table()
+{
+  l1i_hist_table_head[l1i_cpu_id] = 0;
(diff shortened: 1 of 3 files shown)


prefetcher/EIP/EIP.cc
```diff
@@ -0,0 +1,871 @@
...
+}
+
+uint64_t l1i_find_hist_entry(uint64_t line_addr)
+{
+  uint64_t tag = line_addr & L1I_HIST_TAG_MASK;
+  for (uint32_t count = 0, i = (l1i_hist_table_head[l1i_cpu_id] + L1I_HIST_TABLE_MASK) % L1I_HIST_TABLE_ENTRIES; count < L1I_HIST_TABLE_ENTRIES;
+       count++, i = (i + L1I_HIST_TABLE_MASK) % L1I_HIST_TABLE_ENTRIES) {
+    if (l1i_hist_table[l1i_cpu_id][i].tag == tag)
+      return i;
+  }
```
ngober, line 93: The iteration of `i` here is really strange and difficult to parse. It looks like iterating from the head backwards to the beginning, then from the end back to the head. This seems very strange. Would it be possible to use two `std::find_if`s?

```suggestion
  auto tag_check = [tag](const auto& x){ return x.tag == tag; };
  std::reverse_iterator head_rit{std::next(std::begin(l1i_hist_table[l1i_cpu_id]), l1i_hist_table_head[l1i_cpu_id])};
  auto found = std::find_if(head_rit, std::rend(l1i_hist_table[l1i_cpu_id]), tag_check);
  if (found == std::rend(l1i_hist_table[l1i_cpu_id]))
    found = std::find_if(std::rbegin(l1i_hist_table[l1i_cpu_id]), head_rit, tag_check);
  if (found == std::rend(l1i_hist_table[l1i_cpu_id]))
    return L1I_HIST_TABLE_ENTRIES;
  return std::distance(std::begin(l1i_hist_table[l1i_cpu_id]), found.base());
```

prefetcher/EIP/EIP.cc
```diff
@@ -0,0 +1,871 @@
...
+
+// compute delta
+uint32_t l1i_find_bb_merge_hist_table(uint64_t line_addr)
+{
+  uint64_t tag = line_addr & L1I_HIST_TAG_MASK;
+  for (uint32_t count = 0, i = (l1i_hist_table_head[l1i_cpu_id] + L1I_HIST_TABLE_MASK) % L1I_HIST_TABLE_ENTRIES; count < L1I_HIST_TABLE_ENTRIES;
+       count++, i = (i + L1I_HIST_TABLE_MASK) % L1I_HIST_TABLE_ENTRIES) {
+    if (count >= L1I_BB_MERGE_ENTRIES) {
+      return 0;
+    }
```
ngober, line 132: What is this doing here? Isn't this always the terminating condition for the loop, since `L1I_BB_MERGE_ENTRIES < L1I_HIST_TABLE_ENTRIES`?

prefetcher/EIP/EIP.cc
```diff
@@ -0,0 +1,871 @@
...
+  }
+}
+
+uint64_t l1i_find_timing_mshr_entry(uint64_t line_addr)
+{
+  for (uint32_t i = 0; i < L1I_TIMING_MSHR_SIZE; i++) {
+    if (l1i_timing_mshr_table[l1i_cpu_id][i].tag == (line_addr & L1I_TIMING_MSHR_TAG_MASK) && l1i_timing_mshr_table[l1i_cpu_id][i].valid)
+      return i;
+  }
+  return L1I_TIMING_MSHR_SIZE;
```
ngober, line 218: Yes, I think this is just asking for a `std::find_if`:

```suggestion
  auto it = std::find_if(std::begin(l1i_timing_mshr_table[l1i_cpu_id]), std::end(l1i_timing_mshr_table[l1i_cpu_id]), [tag=line_addr & L1I_TIMING_MSHR_TAG_MASK](const auto& x){ return x.valid && x.tag == tag; });
  return std::distance(std::begin(l1i_timing_mshr_table[l1i_cpu_id]), it);
```

prefetcher/EIP/EIP.cc
```diff
@@ -0,0 +1,871 @@
...
+#define L1I_ENTANGLED_NUM_FORMATS 6
+
+uint32_t l1i_get_format_entangled(uint64_t line_addr, uint64_t entangled_addr)
+{
+  for (uint32_t i = L1I_ENTANGLED_NUM_FORMATS; i != 0; i--) {
+    if ((line_addr >> L1I_ENTANGLED_FORMATS[i - 1]) == (entangled_addr >> L1I_ENTANGLED_FORMATS[i - 1])) {
+      return i;
+    }
+  }
+  assert(false);
```
ngober, line 373: This feels like a very dangerous assertion. There aren't guarantees that `(line_addr >> 58) == (entangled_addr >> 58)`.

prefetcher/EIP/EIP.cc
```diff
@@ -0,0 +1,871 @@
...
+typedef struct __l1i_entangled_entry {
+  uint64_t tag;                                        // L1I_TAG_BITS bits
+  uint32_t format;                                     // 3 bits
+  uint64_t entangled_addr[L1I_MAX_ENTANGLED_PER_LINE]; // keep just diff
+  uint32_t entangled_conf[L1I_MAX_ENTANGLED_PER_LINE]; // L1I_CONFIDENCE_COUNTER_BITS bits
+  uint32_t bb_size;                                    // L1I_MERGE_BBSIZE_BITS bits
+} l1i_entangled_entry;
+
+l1i_entangled_entry l1i_entangled_table[NUM_CPUS][L1I_ENTANGLED_TABLE_SETS][L1I_ENTANGLED_TABLE_WAYS];
+uint32_t l1i_entangled_fifo[NUM_CPUS][L1I_ENTANGLED_TABLE_SETS]; // log2(L1I_ENTANGLED_TABLE_WAYS) * L1I_ENTANGLED_TABLE_SETS bits
```
ngober, line 409: If we're using FIFO replacement, I think it would be better for this to be:
```suggestion
std::array<std::array<std::deque<l1i_entangled_entry>, L1I_ENTAGLED_TABLE_SETS>, NUM_CPUS> l1i_entangled_table;
```

prefetcher/EIP/EIP.cc
```diff
@@ -0,0 +1,871 @@
...
+  return pf_addr;
+}
+
+// INTERFACE
+
+void O3_CPU::prefetcher_initialize()
+{
+  cout << "CPU " << cpu << " EPI prefetcher" << endl;
+
+  l1i_cpu_id = cpu;
```
ngober, line 695: I think this could be broken on multicore simulations.

ngober (changes requested):
Thank you for this pull request! I remember that there had been discussions about this, but we reasoned that instructions and data ought to be on different blocks. I suppose that assumption isn't true.

First of all, please rebase onto `develop`. The master branch is intended to be stable and gets only critical hotfixes.

I also have a number of comments about your submitted implementation of EIP, which is in a style that is difficult to reason about and debug. Consider the ip_stride prefetcher as an example of a style that aids in debugging.

zeal4u:
Thank you for your patient code review.
The bug fix codes have been rebased onto `develop`.

PR #256, opened 2022-07-25
Fix a bug for conditional branches, which were not considering the always_taken flag provided by the BTB

alberto-ros:
Before the flag always taken was considered to update the target, but was not updating the branch_prediction field for the instruction. Then, the evaluation for conditional branches to decide if they mispredicted or not, was neglecting that flag. This request fixes this problem.


ngober:
Could you please add a test that fails prior to this change, but succeeds afterwards? This will help to verify the change and make sure it does not regress.

ngober:
I took a closer look at this change and I'm not sure I agree with it. The current policy is to allow the branch direction predictor to correct the BTB, and this change causes the BTB to override the prediction of the direction predictor. I think this is a bad behavior because it will be surprising to a user. When the user returns a prediction from their branch predictor, they expect that prediction to be used every time. If they intend to allow the BTB to override, they can, since the `always_taken` value is given as a parameter.

alberto-ros:
I see your point. Still, ideally the "always_taken" should not be passed to the branch predictor, as the BTB and the branch predictor are supposed to be accessed in parallel. Then with both outcomes, the prediction should be made.

I see two points here:

1. From the point of view of a branch predictor, predicting something and being overridden is not nice, but we are talking about a not-taken prediction for a branch that has always being taken.

2. Performance wise it is better to override. I was seeing some poor performance some times due to this feature, hence my fix. 

But I am ok with either way.

PR #258, opened 2022-07-29
Check for a potential BTB bug.

alberto-ros:
BTB target should not be updated if 0 (which seems the case for non-taken conditional branches)


ngober:
Could you please add a test that shows the non-taken conditional branches would update the BTB prior to this change?

ngober:
I did some tinkering with this. It seemed really strange to me that the target would be zero. The target is set when the trace is read: each branch's subsequent instruction is given as the target. I looked around nearby and I wonder if it would just be better to remove the if-clause at [L148 of the core model](https://github.com/ChampSim/ChampSim/blob/5e4d84f6fa152a450ac02be2c5af545659a9b43c/src/ooo_cpu.cc#L148). That seems to be what is causing the target to be zero, and it seems to be redundant with the check that happens on L198.

alberto-ros:
Thanks, Nathan, to take a look at this issue. The question here is what "branch_target" means. To me like you suggest, it should be the target encoded in the branch/jump instruction or computed by it. In the simulator, right now is 0 then the branch is not taken. So I agree with your suggestion, but taking care that the simulator does not considers in another part of the code that the target should be 0 for non-taken branches. Because I did not check fully if this is the case, I opted for my less intrusive correction. However, I like more your alternative given that the check I mention is done.

alberto-ros:
Hi Nathan. I think that the solution you propose does not resolves this problem. If I am correct, from the trace reader the target_address is filled with the IP of the next instruction in the trace. That is, for the same (direct) conditional branch, when taken the variable will have the actual target and when not taken PC+X (where X is the size of the instruction). Still, the BTB should never be updated with the PC+X (read from the trace) but instead with the actual target of the branch.

So not resetting branch_address to 0 does not resolves the problem but instead fills the BTB with wrong information.

Do you agree with this or I am missing something here?

ngober:
You make a good point. I hadn't considered that. We wouldn't want the BTB polluted with not-taken conditional branches.

I want to ask a different question now, concerning what is good for the simulator instead of what is correct given the current design... I think there are two options:

- Merge this change as-is. Document that the update function receives a 0 for the target when it is not known. The user who writes a BTB module is responsible for handling this case.
- Do not merge this change, but do not call `O3_CPU::impl_btb_update()` for conditional not-taken branches. When the update function is called, it is guaranteed to contain a correct address and the user can handle it.

From a user's perspective, would it be better to receive the update that the conditional branch was not taken or to have that guarantee of a correct target every time? I think I lean towards the first, to merge this change as-is, but I want to make sure.

ngober:
I agree that the traces need to change. I think a new trace format is going to become necessary very soon. I'll keep this on the list for features to include in it.

PR #270, opened 2022-09-09
Strengthened support for ASID checking

ngober:
This patch affirms that the ASID is actually meaningful. Everywhere we check the address field, we now also check the ASID, if that makes sense for the context.

This is actually a somewhat simple change, just adding in equality checks, but there is one big issue: This very likely breaks prefetching in shared caches. This patch makes the assumption that the ASID of a prefetch is the same as the ASID of the most recent packet. This makes sense if the cache only ever sees one ASID or if the prefetch is issued in either `CACHE::prefetcher_cache_operate()` or  `CACHE::prefetcher_cache_fill()`. However, if the prefetch is issued in `CACHE::prefetcher_cycle_operate()`, it is very likely the ASID will have changed in the meantime, so the prefetch is much less likely to hit the cache if the block is already resident, and is entirely unlikely to be useful.

This aliasing problem already existed to a certain extent with the `CACHE::cpu` member, but that problem would usually only arise when the member was used as an index into arrays or something like that. It wouldn't cause prefetches to become useless. Unfortunately, this means modules must be redesigned to save the `CACHE::asid` member or the prefetching module interfaces must be changed to take an ASID parameter. It's not clear to me which is more desirable right now.


ngober:
I'm closing this because it has fallen derelict for too long. If different work arises to fix this, we can revisit the idea.

PR #280, opened 2022-10-01
Check for the update of a zero target in the BTB

alberto-ros:
This is a test in order to demonstrate that the BTB is updated sometimes with a zero target. The added assert triggers when running for example server_039.champsimtrace.xz (IPC-1).

This pull request is a test for the pull request #258. BTB should not be updated (and an entry should not be added) on a branch with a target == 0.


ngober:
Closing this following the conversation in #258. Merging this would fail the assertion on almost every workload, but the workaround has already been merged.

PR #291, opened 2022-11-09
Update ip_stride.cc

FJShen:
Seems like an error in the calculation of the starting boundary of the tracking set.


ngober:
Thanks for the patch! It definitely looks like a bug. The PR prompted me to finally get the Module Support Library rolling out, #292. I've gone ahead and modified the ip_stride prefetcher to use these new features, since they would also serve to fix the bug.

PR #292, opened 2022-11-09
First draft of Module Support Library

ngober:
This patch introduces the Module Support Library. It contains a few functions and a class which are designed to be used by user modules. It's not a lot, but I hope that it could eventually grow.

The intention of the MSL is to provide users with a set of well-tested structures to simplify designs. Rather than writing yet another (potentially buggy) implementation of an LRU table, the user can use `champsim::msl::lru_table`, which is tested and simpler to use.

Currently, no documentation is provided for this. There probably needs to eventually be a wiki page about each of these.


ngober:
I also added a tested fixed-width counter that saturates at fixed values. There are two using-declarations exported: `champsim::msl::fwcounter<BIT_WIDTH>` and `champsim::msl::sfwcounter<BIT_WIDTH>`, implementing an unsigned and signed counter.

test/040-lru-table.cc
```diff
@@ -0,0 +1,217 @@
...
+    WHEN("We add a new element to the same set as the first element") {
+      uut.fill({data+2});
+
+      THEN("A check to the first-added element misses") {
+        auto result = uut.check_hit({data});
+
+        REQUIRE_FALSE(result.has_value());
+      }
+
+      THEN("A check to the second-added element misses") {
```
FJShen, line 135: "hits"

PR #293, opened 2022-11-11
Develop

agusnt:
This patch solve several bugs with the first levels of cache and the virtual memory

### Bug

The PTE does not remove from the free list the PA used.

### Bug

When merging two instructions in the L1D queues, the following case may occur: untranslated instruction A has physical and virtual address `0xbcde` and translated instruction B has physical address `0xbcde`. The two instructions will merge generating a deadlock since instruction B has not been fetched.

### Solution

A flag has been added to indicate that the instruction has been translated, only when the instructions are translated is the merge performed.

## PTW to the first level Bug

This is the biggest change in the patch

### Bug

If there is a burst of processor requests that are going to miss in all TLB levels. Only the first requests will be served (depending on the number of walkers). The rest remain in L1D RQ and MSHR waiting for a walker to be released. All of them remain in L1D RQ until filled. If the processor requests go to the same pages that are resolving the walkers remain in L1D RQ until the translation is resolved. The walkers have not yet finished going through all the translation levels.  When the walker tries to access the next level of the translation table they cannot insert the request in L1D RQ because is full. 

### Solution

Add a new queue for PTW requests at L1D

### Bug

If the prefetcher (L1D) launches the address A with destination L2 and subsequently launches the same address with destination L1D, a deadlock occurs when the MSHR is merged.

### Solution

When we perform the merge of two memory instructions (one is a prefetch) the `fill_this_level` variable is updated to true if one of the two instructions has it set to true.


ngober:
Thanks for the patch. I like the simplicity of the approach. It looks like the tests here are failing because you've added a pure virtual function to a base class. Would you please repair these tests? Also, please add some tests to demonstrate that the new behavior works. I want to avoid a situation where the behavior might regress due to later changes.

agusnt:
Hi Nathan,

I fix the test and add several test to test the new patch, I try in my machine and `make test` run flawless. Also add an option in the function `get_size` (cache.cc) to return the size of PTWQ.

agusnt:
Hi Nathan,

I am just merge your latest changes. I test the changes using the following: `./config.sh` and `make test` and all test passed.

Captura de pantalla de 2022-11-22 13-16-47

Can you tell me how to reproduce the way that you test the changes?

Thank you

ngober:
Something strange is happening that I don't quite understand. I'm seeing the tests pass when I run them locally, but test 600 seems to fail whenever Github runs the tests. I've tried to figure out what's going on, but I haven't been able to devote my full attention to it.

I was wondering at first if the issue was related to the virtual memory, so I fixed a bug in 3e2b0d402b0a2fc065b013d1907d10a9fa829ccf, but it doesn't seem to have actually resolved the problem.

I like these changes and I want to merge them, but I want the tests to pass first. If you can figure out what's going on, that would be very helpful.

ngober:
Ah, I accidentally wrecked the tests by trying to use the Github text editor. I'll fix it later tonight or possibly tomorrow.

agusnt:
I think I have an idea of what is going on. Since the problem does not happen on my local machine (using gcc or clang), I think it come from how the compiler handles some stuff.

I run the 600 benchmark on GitHub and on my local machine with debug mode and discover that the `get_pte_pa` function returns a different address when running locally or online

```
My local machine
[VMEM] get_pte_pa paddr: ffff0000 vaddr: deadbeef pt_page offset: 0 translation_level: 4

GitHub
[VMEM] get_pte_pa paddr: 0 vaddr: deadbeef pt_page offset: 0 translation_level: 4
```

So I believe that the problem resides in how `ppage_free_list` constructor is handle by the compiler. 

Since the other 600 test works well, I fix the test by change the first `vmem` parameter (`paddr`) from 20 to 33 (like the other tests from 600 file)

https://github.com/agusnt/ChampSim/commit/61520d1d92bd35619c626026273b32a113d5aaae

and now the value return by `get_pte_pa` function seems right

```
[VMEM] get_pte_pa paddr: 101000 vaddr: deadbeef pt_page offset: 0 translation_level: 4
```

However, I do not know if this change is ok (even if all test pass now). If the definition of vmem in the test was right, I think that we have to fix how `vmem` reserve memory to avoid that it can give a physical address equal to 0.

What do you think?

ngober:
Looks like you're correct. The problem was that the virtual memory in those tests was too small. I don't know how that didn't get picked up in other places. I tweaked the virtual memory model and added a bunch of assertions that highlighted other tests that were sneaking by. I'll probably need to go through and fix more with the virtual memory model, but for now, this will do.

Issue #296, opened 2022-12-09
Branch MPKI bug


sawansib:
Hello,

Recently the branch MPKI has increased (in my tests and also mentioned by [Daniel A. Jiménez](https://github.com/ChampSim/ChampSim/issues/271) in the current develop branch. I found that the reason is an incorrect update of branch type information when doing the BTB update. 

Consider the [line](https://github.com/ChampSim/ChampSim/blob/7b459db03ce62d8db2c7afc7db16a4609f8b28b7/btb/basic_btb/basic_btb.cc#L124) where by default the branch type is set to ALWAYS_TAKEN. Now if we have a case where the BTB entry for a taken conditional branch already exists then the same entry's branch type will be updated from conditional (original BTB entry branch type) to always taken which is incorrect.

In my tests, I find that fixing this error will revert back the MPKI to the previous version of the develop branch. As my fixes are temporary I'm not including them here. 

Sawan

ngober:
I'm looking through the git blame for the line you linked, and it looks like a merge somewhere introduced this bug. Could you give more detail about the case you observe where the behavior is incorrect? Also, if you submit a pull request for this, I'd be eager to see your solution.

sawansib:
Hey Nathan,

Sorry for late reply!

You can see the issue in the log i am adding below. The log simply prints which IP is added to the BTB and also the result of the search in the BTB by IPs. The log is for spec_gcc_001 from the 1st IPC. 

`
btb_btbDbasic_btb_update_btb ip? 66b60c target? 66b5f0 **type? 3** taken? 1
btb_btbDbasic_btb_update_btb ip? 66b600 target? 66b60c type? 3 taken? 1
btb_btbDbasic_btb_btb_prediction ip? 66b60c target? 66b5f0 **type? 2**
`

Bug: - You can see for the IP `66b60c` the BTB is updated with branch type 3 (CONDITIONAL), but when the same IP looks in the BTB it gets the target correctly, but the branch type is 2 (ALWAYS_TAKEN) which is incorrect. 

Cause:- It happens due to the line i mentioned in the beginning of this issue. So, when the update is called in the BTB by default the branch type is set to 2 (ALWAYS_TAKEN). In the case of the taken branch there is no condition that update the default branch type of 2 set earlier. 

Solution: - The simple solution is to add a condition for taken branches. For example: -
`
 auto opt_entry = ::BTB.at(this).check_hit({ip, branch_target, type});
  
  if (taken) {
[...]

PR #297, opened 2022-12-12
Created strong type for addresses

ngober:
The correctness of address operations is critical to the correctness of ChampSim. This patch introduces a strong type for addresses to prevent bugs that would result from unchecked operations on raw integers.

The primary benefit is this: Most address operation bugs are compiler errors. Those that are not are usually runtime exceptions. If a user still needs to manipulate the bits of an address, they can, but they must explicitly enter an unsafe mode.

The need for this patch became apparent to me after 3165c461db2aa5490583cefeb901fa3ab6ff00d1, where I fixed a bug that arose from bad bit shifting in address manipulation. In this commit, the cache line number was being treated as a full-length address, and the size of the stride increased by a factor of the block size. After this patch, such a bug is quite difficult if not impossible. Now, the a cache line number is a completely separate type, `champsim::block_number`, from the full-length address, `champsim::address`, and so it is impossible to implicitly assign one to the other. The user is required to explicitly make the appropriate conversions.

### Types introduced
This patch introduces the following types, all in the `champsim` namespace:
 - `address`, a full-length, 64-bit address
 - `block_number` and `block_offset`, the bits above and below the size of a block. These types depend on the configuration.
 - `page_number` and `page_offset`, similar to the previous.
 - Three "extent" types: `static_extent<std::size_t UPPER, std::size_t LOWER>`, `dynamic_extent`, `sized_extent`. The static extent is compile-time sized, the other two are runtime-sized. The dynamic and sized extents differ in the meaning of their constructor parameters: a dynamic extent takes an upper and lower bound, while a sized extent takes a lower bound and a size.
 - `address_slice<Extent>`, a generalization of a subset of address bits. The previous types are simply specializations of this type: `block_offset == address_slice<static_extent<LOG2_BLOCK_SIZE, 0>>`.
 - `address_slice<dynamic_extent>`, a runtime-sized slice. This type sacrifices compile-time error checking for runtime checking, but provides more flexibility.

### Conversions
The address slices can be converted between one another. The bit shifting and masking will be handled by the types. For most operations, users should use the converting constructors.

    // LOG2_PAGE_SIZE = 12
    // LOG2_BLOCK_SIZE = 6
    address full_addr{0xffff'ffff};
    block_number block{full_addr}; // 0xffff'ffc0
    page_number page{full_addr}; // 0xffff'f000

If the type conversion is widening, the new bits will be 0. Importantly, the template parameters to `champsim::address_slice` are _absolute_.

All address slices can take their extent as the first constructor parameter

    address_slice st_full_addr{static_extent<64,0>{},0xffff'ffff};
    address_slice st_block{static_extent<64, LOG2_BLOCK_SIZE>{}, st_full_addr}; // 0xffff'ffc0
    address_slice st_page{static_extent<64, LOG2_PAGE_SIZE>{}, st_full_addr}; // 0xffff'f000

    address_slice dyn_full_addr{dynamic_extent{64,0},0xffff'ffff};
    address_slice dyn_block{dynamic_extent{64, LOG2_BLOCK_SIZE}, dyn_full_addr}; // 0xffff'ffc0
    address_slice dyn_page{dynamic_extent{64, LOG2_PAGE_SIZE}, dyn_full_addr}; // 0xffff'f000

    address_slice szd_full_addr{sized_extent{0, 64},0xffff'ffff};
    address_slice szd_block_offset{sized_extent{0, LOG2_BLOCK_SIZE}, szd_full_addr}; // 0x3f
    address_slice szd_page_offset{sized_extent{0, LOG2_PAGE_SIZE}, szd_full_addr}; // 0xfff

For static extents, the extent can be specified as a template parameter.

    address_slice<static_extent<64,0>> st_full_addr{0xffff'ffff};
    address_slice<static_extent<LOG2_BLOCK_SIZE, 0>> st_block{st_full_addr}; // 0x3f
    address_slice<static_extent<LOG2_PAGE_SIZE, 0>> st_page{st_full_addr}; // 0xfff

The address slices have a constructor that accepts a `uint64_t`. No bit shifting is performed in these constructors. The argument is assumed to be in the domain of the slice.

    champsim::block_number block{0xffff}; // 0x003f'ffc0
Relative slicing is possible with member functions:

    address_slice<static_extent<24,12>>::slice(static_extent<8,4>{}) -> address_slice<static_extent<20,16>>
    address_slice<static_extent<24,12>>::slice_upper<4>() -> address_slice<static_extent<24,16>>
    address_slice<static_extent<24,12>>::slice_lower<8>() -> address_slice<static_extent<20,12>>

### Arithmetic

The offset between two addresses can be found with

    auto champsim::offset(champsim::address base, champsim::address other) -> champsim::address::difference_type

The function is a template, so any address slice is accepted, but the two arguments must be of the same type. The return type is signed and the conversion is safe against overflows.

Address slices also support addition and subtraction with signed integers. The arguments to this arithmetic are in the domain of the type.

    champsim::block_number block{0xffff}; // 0x003f'ffc0
    block += 1; // 0x0040'0000

### Splicing

Two or more slices can be spliced together. Later arguments takes priority over the first, and the result type has an extent that is a superset of all slices.

    champsim::splice(champsim::page_number{0xaaa}, champsim::page_offset{0xbbb}) == champsim::address{0xaaabbb};

### Unsafe Operations
Sometimes, it is necessary to directly manipulate the bits of an address in a way that these strong types do not allow. Additionally, sometimes slices must be used as array indices. In those cases, the `address_slice<>::to<T>()` function performs a checked cast to the type.

    champsim::address addr{0xffff'ffff};
    class Foo {};
    std::array<Foo, 256> foos = {};
    auto the_foo = foos.at(addr.slice_lower<8>().to<std::size_t>());


alberto-ros:
It seems to me like a huge change for a "small" bug. I think it would affect all files (everything works with addresses) including prefetchers, replacement policies, predictors, etc. In case of going further with this change, we should be sure that performance remain exactly the same.

ngober:
The inciting bug was small, maybe, but it made the output of the prefetcher completely incorrect. It got me thinking that if I overlooked this kind of bug in a simple prefetcher, how much more do these bugs exist in more complex designs written by people who don't do this every day?

I can't say it for certain right now, but I don't see a significant opportunity for performance change. The goal of this change is to behave the same as before, but to make certain suspect actions into compiler errors. That is, if it compiles, we should be able to say that it's probably correct.

alberto-ros:
I agree with that you say regarding correctness. But mind that the results check is a necessary thing to do first. For example, a recent formatting change in the BTB introduced a performance bug regarding the always_taken flag #296. So my point would be to split (if possible) formatting changes or changes that do not change simulator behavior from changes that affect performance. And for the first ones, verify that performance is not affected (I guess for a number of predictors, replacement policies, prefetchers, etc).

ngober:
You're right. I'll do that locally. I do know that @GinoAC intends to work on a synthetic trace generator so that we can do some more robust integration testing. I'm looking forward to that.

ngober:
I ran all of the SPEC 2017 traces for 200M/1B, with default configuration, and the speedup of the model is negligible: 1.0000724125847442

There seems to be some differences of less than 1%, and only on a few traces. I could look into it, but that might be a very tedious endeavor.

alberto-ros:
Thanks a lot! I think checking those should not be necessary, although it is surprising. But I would run apart from the base, a version with more elaborated predictors: hashed-perceptron, IP-stride, some RIP replacement ... this way we also checked bugs in those codes (Ideally all, but I know it is harder). If we cover the frequently used, I am fine with it.

GinoAC (changes requested):
This looks great to me and is good to merge given the results you mentioned.

I recommend adding the description of this patch to the wiki for users to reference.

ngober:
I have merged #330 into this PR. Doing this eases the transition to the strong address types, since both the `uint64_t` signatures and the `champsim::address` signatures are recognized by the simulator. This was not possible before #330.

ngober:
@GinoAC I've added Doxygen-style comments to `champsim::address_slice`. It more or less has the contents of the top post on this thread. Rather than try and fully integrate Doxygen to document the C++ parts of ChampSim in this patch, I think I want to simply add the comments here and build out the HTML generation separately. What do you think

Also, I've rearchitected this slightly since its original post. I now template the address slice on an "extent", which can be passed to the constructor. Doing it this way eliminates cases where the user would write

    champsim::address_slice<champsim::dynamic_extent> my_addr{16, 8, 24};

In which the meaning of the constructor parameters is unclear at the call site. Instead, now it becomes:

    champsim::address_slice my_addr{champsim::dynamic_extent{16, 8}, 24};

which is less prone to confusion. It does mean that I have to sacrifice the `address_slice<16,8>` syntax, now we get `address_slice<static_extent<16,8>>`, but we sometimes must sacrifice pleasant code for safe code. :man_shrugging: 

I'm happy with where this is now and I'm ready for merge if there are no objections.

PR #298, opened 2022-12-20
Fix upper level prefetch

agusnt:
When a prefetch is issue with `fill_this_level == true`, the prefetch request never fill the lower level. This patch fix the bug, and add two test to verify the behavior.

# Why does this bug happen?

The flag `fill_this_level` packet flag is never set to true

# Fix 

After the first MSHR miss, the `fill_this_level` flag is set to true.

https://github.com/agusnt/ChampSim/blob/90c9e5b09f39c146c3cda872a7b8ad8cb0111776/src/cache.cc#L208-L211

### DIFF
--- a/src/cache.cc
+++ b/src/cache.cc
@@ -205,11 +205,10 @@ bool CACHE::handle_miss(const PACKET& handle_pkt)
 
     if (handle_pkt.fill_this_level)
       fwd_pkt.to_return = {this};
-    else {
-      fwd_pkt.fill_this_level = true; // We will fill the next level
+    else
       fwd_pkt.to_return.clear();
-    }
-
+    
+    fwd_pkt.fill_this_level = true; // We will always fill the lower level
     fwd_pkt.prefetch_from_this = false;
 
     bool success;
--- a/src/cache.cc
+++ b/src/cache.cc
@@ -207,7 +207,7 @@ bool CACHE::handle_miss(const PACKET& handle_pkt)
       fwd_pkt.to_return = {this};
     else
       fwd_pkt.to_return.clear();
-    
+
     fwd_pkt.fill_this_level = true; // We will always fill the lower level
     fwd_pkt.prefetch_from_this = false;
 
--- a/src/cache.cc
+++ b/src/cache.cc
@@ -205,8 +205,10 @@ bool CACHE::handle_miss(const PACKET& handle_pkt)
 
     if (handle_pkt.fill_this_level)
       fwd_pkt.to_return = {this};
-    else
+    else {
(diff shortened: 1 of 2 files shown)


test/421-prefetch-fill-level.cc
```diff
@@ -72,3 +72,97 @@ SCENARIO("A prefetch can be issued without creating an MSHR") {
...
+
+      for (int i = 0; i < 1000; i++) {
+        uut_queues._operate();
+        uut._operate();
+        mock_ll._operate();
+      }
+
+      // We ensure that the block fill the first level
+      bool find = false;
+      for (auto i: uut.block) if (find = (i.address == 0xdeadbeef)) break;
```
ngober, line 105: Could you please perform this check by using the cache's interfaces? See [a similar line in test 420](https://github.com/ChampSim/ChampSim/blob/2aa99f98d69f3467d765141c2bd31e2b474ca256/test/420-issue-prefetch.cc#L50).

Performing tests via the interfaces checks for correct behavior, rather than correct implementation. Specifically, we are testing that a prefetch issued with `fill_this_level == true` causes a later hit in the second level, but not in the first level. So, we want to test that the return time of a packet with the same address has the miss latency of the first level, but not the miss latency of the second level. If we do it this way, we're free to change the implementation of the cache without breaking the tests.

src/cache.cc
```diff
@@ -205,8 +205,10 @@ bool CACHE::handle_miss(const PACKET& handle_pkt)
 
     if (handle_pkt.fill_this_level)
       fwd_pkt.to_return = {this};
-    else
+    else {
+      fwd_pkt.fill_this_level = true; // We will fill the next level
       fwd_pkt.to_return.clear();
+    }
 
     fwd_pkt.prefetch_from_this = false;
```
ngober, line 213:
```suggestion
    else
      fwd_pkt.to_return.clear();
    
    fwd_pkt.fill_this_level = true; // We will always fill the lower level
    fwd_pkt.prefetch_from_this = false;
```

This is minor, but I think moving the assignment down to match the assignment of `prefetch_from_this` expresses the intent better. It does the same thing, but more clearly says "We will always fill the lower level".

ngober (commented):
Good bug catch!

This is probably good to merge as-is, but I want to request a change to the tests now rather than need to fix it later.

test/421-prefetch-fill-level.cc
agusnt: I see, I change the test to be at the same style like the 420.

ngober:
It looks great! The tests demonstrate the bug and the patch fixes it. Great job.

PR #303, opened 2023-01-10
Improvements to TLB model correctness

ngober:
This patch is a fix for #227. In that patch, my goal was to extract the TLBs from the core model, but I didn't want to disrupt the cache model. As a result, we ended up with a bunch of issues that caused deadlocks (notably, some of these were fixed in #293). In this patch, I moved the translation logic into the tag check logic of the cache model, so TLB lookups are now completely performed by the cache. I avoid deadlocks by giving each producer-consumer pair its own dedicated set of queues. 

This change really improves configuration flexibility. With the previous solution, there were many odd configurations that could produce strange deadlocks. Here, there is much less potential for that, and the model matches a real system more closely.

I will be doing performance analysis of this patch soon, but I want to get the code out there for other people's eyes while I do.


ngober:
I've been in contact with a user who has sent me a configuration for which ChampSim crashes before this patch, but does not after. I spent a while trying to isolate the behavior into a test to prevent regression, but it ended up being really complex and I'm not sure I'll be able to snapshot the state. The configuration can be found [here](https://github.com/ChampSim/ChampSim/files/11122366/crash_for_303.txt), and the crash happens for 400.perlbench-50B. I see this as a good reason to merge this patch, since it changes the TLB miss setaside to be more resilient.

PR #308, opened 2023-02-11
Added support for Apple devices

GinoAC:
This patch changes how the trace file pointer is instantiated on Mac devices and the file is read. I've confirmed it works on M1's.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -123,7 +123,11 @@ jobs:
       with:
         submodules: 'true'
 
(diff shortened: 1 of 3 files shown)


ngober:
Could you try uncommenting the lines in `.github/workflows/test.yml` that have to do with Mac? I would love for those tests to work, if possible.

GinoAC:
Looks like it fails when it tries to install vcpkg. I'll dig a little and see what if I can figure out what the root cause is.

PR #320, opened 2023-03-14
Only update btb with a valid target.

pAbellan-arch:
If you enter a target 0 entry, you may evict a valid entry with a real target.


ngober:
Are you sure this occurs after #318? I have tried modifying the BTB test that we introduced in that patch, and I cannot reproduce the bug this claims to fix.

[Here](https://github.com/ChampSim/ChampSim/files/10971038/issue320_test_patch.txt) is the patch that I applied to the test. Does this not replicate the behavior?

alberto-ros:
Hi Nathan, 

This patch is not related to the one of Sawan (#318) nor the one I submitted not long ago (#258), but somehow more related to this last one. 

The "problem" is that entries for non-taken branches that miss in the BTB are inserted there, potentially evicting a valid entry. Mind that in ChampSim non-taken branches have a target value of 0. So the entry is inserted for nothing.

I indeed would not consider it a bug but a bad design decision. However, seeing that in the master branch that decision is different, I can infer that a refactoring unintentionally introduced that decision.

If my data is correct, Sawan's fix moved MPKI from ~8 MPKI to ~6 MPKI, and this one from Pascual moves it from ~6 MPKI to ~2 MPKI, going back to the original and a few times reported MPKI for IPC-1 traces.

With this patch I think we close (hopefully) the issue of high/changes in the MPKI.

ngober:
Let me see if I understand.

1. Fill the BTB with branch Y, which is taken at least once, so has a target
2. Fill remaining ways with other branches
3. Fill the BTB with branch X, which is initially not taken, so has a target of 0
4. X is not found in the BTB, so line 141 does not apply
5. X is filled with a target of 0, evicting Y

In a real machine, the target of X would be known even if it is not taken. We must make a choice given the current trace format either to retain Y in the BTB when it would have been evicted in a real machine, or to fill X with target 0 causing later predictions to X to mispredict where they would have been correct in a real machine.

I don't really like either choice here. However, I think you're correct to say that retaining Y is preferable in an imperfect world because at least something gets predicted. The solution really is to generate new traces. I can feel it coming soon. I'm going to run tests and merge, and I'll get some things moving for a new trace format. It probably won't make the next release to master (should happen soon after #303 is merged), but it will probably make the following one.

alberto-ros:
Yes, you are right, Nathan. That's the case. Two extra comments:

1. Usually in text books, not taken branches do not fill BTB in case of a miss (so this change would be aligned with it).

2. Pascual did also a test where "magically" the correct target was inserted on non-taken branches instead of a 0. Still the MPKI was quite high. The reason may be branches that are always non taken, which occupy useless entries.

ngober:
Hmm, that's significant. In that case, I'm going to add a test to make sure this behavior is guaranteed.

pAbellan-arch:
Hi Nathan,

First of all, thank you for accepting the pull request.

Just to clarify some data about the MPKI of the BTB that Alberto mentioned. The MPKI of the BTB after the changes made by Sawan decreased to ~3.6 MPKI for the IPC-1 traces, which is an improvement compared to the ~8 MPKI we were getting before. If we subsequently add the control of not inserting entries with an invalid target, we keep decreasing until the ~1.27 MPKI that we previously mentioned.

PR #322, opened 2023-03-16
Always taken not used for conditional branches

alberto-ros:
This is a necessary fix after the last patch on the BTB. I believe it is not good to use the BTB to predict conditional branches, so always taken info is now only for inconditional branches.

### DIFF
--- a/btb/basic_btb/basic_btb.cc
+++ b/btb/basic_btb/basic_btb.cc
@@ -126,17 +126,11 @@ void O3_CPU::update_btb(uint64_t ip, uint64_t branch_target, uint8_t taken, uint
     type = ::branch_info::INDIRECT;
   else if (branch_type == BRANCH_RETURN)
     type = ::branch_info::RETURN;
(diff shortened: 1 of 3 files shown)


alberto-ros:
It should be unrolled in my opinion. I will elaborate more here, but MPKI results talk by itself. My point is that always-taken conditional branches should be detected by the branch predictor, not by the BTB, at least in the current version of ChampSim with a single BTB level. 

Consider a branch that it is first Taken, then not taken, and then not accessed in a long time (so evicted from the BTB). With the current mechanism the BTB will force the branch to be predicted as always taken. However, the branch predictor may predict it correctly. 

So, this always taken optimization for conditional branches may make sense if that information is always in the BTB, but this is not the common case.

ngober:
That makes sense. In particular, it seems reasonable that if the user wanted a BTB with that optimization, they should write one. I'll adjust the tests to demonstrate this and merge everything in.

PR #323, opened 2023-03-16
Migrated CLI parsing to CLI11

ngober:
This patch integrates the `CLI11` library as a command-line parser. It provides significant extensibility, especially as it would allow for future work to add options. It provides validation as well, helping ensure the input is valid.

This patch potentially introduces a breaking change: the options `--warmup_instructions` and `--simulation_instructions` have been replaced with `--warmup-instructions` and `--simulation-instructions`, respectively. If the old options are used, a help message will be printed, so I don't see this as a problematic change. It also breaks an undocumented feature that allowed an HTTP URL to be passed as a trace file name. If you were using it, you can pipe the file in with `<(curl -s trace.xz | xz -dc)` instead.

### DIFF
--- a/.github/workflows/buildcheck.yml
+++ b/.github/workflows/buildcheck.yml
@@ -45,7 +45,7 @@ jobs:
 
     # TODO run an actual set of checks
     - name: Run tests
-      run: ./bin/champsim -w50000000 -i50000000 --json=single_core_result.json -- https://dpc3.compas.cs.stonybrook.edu/champsim-traces/speccpu/400.perlbench-41B.champsimtrace.xz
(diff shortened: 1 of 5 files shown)


GinoAC (approved):
Looks good to me. I'm ok with this merging if there are no other changes you're wanting to make.

ngober:
Should this go into the next master merge, or wait until after? It includes a very minor breaking change.

GinoAC:
I'd be ok with this going into the next master merge since it's a v2 release, and it might be an excellent opportunity to advertise the flag changes. Worst case, it breaks some folk's scripts (mine included), but I doubt this will affect anyone's workflow beyond having to change a line or two.

ngober:
Sorry, I was unclear. Should we merge this immediately into v1.1, or merge it into v2?

PR #330, opened 2023-04-26
Added support for modules as objects

ngober:
This patch has been three years in the making, and I'm glad it's finally seeing the light of day.

This patch allows modules to be declared as proper C++ objects. This eliminates a whole class of bugs which can occur where information is accidentally shared in global state. With this change, we can provide stronger guarantees that the modules will be isolated. If you want to share global state, you must do so explicitly.

Modules should now be specified as subclasses of `champsim::modules::branch_predictor`, `champsim::modules::btb`, etc. The module must have a constructor that takes a pointer to `O3_CPU` or `CACHE`, depending on the module. This can be accomplished with `using prefetcher::prefetcher;`. See the built-in modules for examples.

Modules still interact with the simulator using the same set of module hooks, but now the hooks are members of the module, not the model. For example:

    void O3_CPU::last_branch_result(uint64_t pc, uint64_t branch_target, uint8_t taken, uint8_t branch_type)

should be converted to

    void my_branch_predictor::last_branch_result(uint64_t ip, uint64_t branch_target, bool taken, uint8_t branch_type)

This is a major improvement: not all of the module hooks must be implemented. The simulator will detect the presence (or absence) of the hooks and ignore any that are not implemented. As an example, take the interface for the LRU replacement policy:

    struct lru : champsim::modules::replacement
    { 
       std::size_t NUM_SET, NUM_WAY;
       std::vector<uint64_t> last_used_cycles;
       uint64_t cycle = 0;
     
       lru(CACHE* cache) : replacement(cache), NUM_SET(cache->NUM_SET), NUM_WAY(cache->NUM_WAY), last_used_cycles(NUM_SET*NUM_WAY, 0) {}
    
       //void initialize_replacement();
       long find_victim(uint32_t triggering_cpu, uint64_t instr_id, long set, const CACHE::BLOCK* current_set, uint64_t ip, uint64_t full_addr, uint32_t type);
       void update_replacement_state(uint32_t triggering_cpu, long set, long way, uint64_t full_addr, uint64_t ip, uint64_t victim_addr, uint32_t type, uint8_t hit);
       //void replacement_final_stats()
    };

The module does not implement the initialization of final stats hooks. The simulator will detect this and the functions will not be called. 

The old style of modules is still supported, and are now called "legacy" modules. There are two ways to specify in a configuration that a module is a legacy module. In the first, place an empty file named `__legacy__` in the module source directory. If it is present, the module will be compiled in the legacy manner. In this patch, the `spp_dev` prefetcher is the only built-in module for which this is the case. In the second, the configuration can be specified as:

    {
        "L2C": {
            "prefetcher": {
                "legacy": true,
                "path": "~/path/to/module"
            }
        }
    }

Modules can now be specified simply as the path to the sources (as before), or with the following configuration, where the sources for the module `my::prefetcher` can be found at `~/path/to/module`.

    {
        "L2C": {
            "prefetcher": {
                "path": "~/path/to/module",
                "class": "my::prefetcher"
            }
        }
    }

If the module is given as a path, in the old way, the class will be assumed to be named the same as the directory. However, it is not good to rely on this assumption. The class name should be specified.

This patch will also allow us to easily change the module interface in the future. Since now the module interfaces can be detected at compile time, it can be adapted to a new interface without the user needing to change the module.


ngober:
@GinoAC Given today's discussion in #344, would it be good for us to accelerate this patch to before the upcoming master release? It would ease the transition for people migrating IPC-1 code.

GinoAC:
I think it would be worth adding to the top of the stack. I'll get to reviewing!

ngober:
If we merge this, I could even re-enable support for the old branch and btb hook signatures from prior to #255 before we merge, which could even be better.

ngober:
I've done a little more thinking about this, and I'm not 100% sure that I want to merge this prior to the release. I think, while this patch is good and it's important, it's not mature yet. 

In order to access the internals of either `CACHE` or `O3_CPU`, you have to do so through the `intern_` pointer, which pretty clunky. I think we should establish a good const interface into these after it is merged, but it will take some time to work out all the bugs.

Per #344, without this patch, the next merge will break some legacy code, particularly from IPC-1. That may just have to be a pill we swallow.

alberto-ros:
I agree with you, Nathan. I do not see a big problem in #344. Well, instead of adding this patch, perhaps we could just add the prefetch_hit argument as it was before. Data prefetchers are free not to use it, and anyway the interface has already changed slightly both for data and instructions prefetchers (since their championships) so a minimal porting is always necessary.

GinoAC (approved):

PR #331, opened 2023-04-27
Update README.md

GinoAC:
We were missing the -traces flag in the example on the README.


ngober:
The flag isn't necessary. It just tells the parser that everything else is a trace file name.

GinoAC:
Yeah disregard this. Just realized the bug report I got was from someone using an older version of champsim.

PR #349, opened 2023-05-29
Added useful_prefetch parameter to prefetchers

### LINKED ISSUE #344: prefetch_hit parameter not passed any more to prefetcher_cache_operate
Hello, 

Until recently the prefetch_cache_operate function had a parameter called prefetch_hit. It was used by many prefetchers that now cannot be ported without modifying the prefetcher interface.

The information about prefetch_hit is still in cache.cc: (way->prefetch && !handle_pkt.prefetch_from_this)

Do you think it could be added back or otherwise why it was removed?

Thanks!

ngober:
Resolves #344 

This patch adds a parameter to the prefetcher cache_operate hook that indicates whether the hit block was a prefetch or not. This will help prefetchers get good feedback on their performance, and it also allows easier migration from the IPC-1 prefetchers. This is, however, a breaking change for all existing code in the name of bringing instruction prefetchers and data prefetchers together.

### DIFF
--- a/config/modules.py
+++ b/config/modules.py
@@ -199,7 +199,7 @@ def get_cache_module_lines(pref_data, repl_data):
 
     pref_nonbranch_variant_data = [
         ('prefetcher_initialize',),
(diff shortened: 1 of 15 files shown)


alberto-ros (approved):
It looks good to me to merge.

PR #356, opened 2023-06-08
BRANCH_OTHER are considered conditional.

alberto-ros:
This pull request includes two fixes:

The first one is that BRANCH_OTHER considered conditional branches in ChampSim. Among other effects, the most important current "bug" is that the BTB considers them as "always_taken" which is not right.

The second ones is considering that conditional branches can also read from other registers, and not just from flag registers.

In general, I think we should avoid to have a BRANCH_OTHER type as it may be confusing. But for now, this patch fixes some current problems.

### DIFF
--- a/btb/basic_btb/basic_btb.cc
+++ b/btb/basic_btb/basic_btb.cc
@@ -100,7 +100,7 @@ void O3_CPU::update_btb(uint64_t ip, uint64_t branch_target, uint8_t taken, uint
     ::INDIRECT_BTB[this][hash % std::size(::INDIRECT_BTB[this])] = branch_target;
   }
 
-  if (branch_type == BRANCH_CONDITIONAL) {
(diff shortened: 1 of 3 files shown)


ngober:
I think having a `BRANCH_OTHER` is important, because the simulator can't be exhaustive. There's bound to be edge cases where we can identify that a function is a branch, but we can't identify its type. 

Regarding whether "other" branches should be considered conditional, you're basically arguing that the default should be that a branch is conditional. I think I agree with this, because at least branches not known to be unconditional will predict correctly.

alberto-ros:
I did a commit that removes the second condition, such that the pull request is less intrusive. Since OTHER branches can be not taken according to their specification, the fix has to be correct.

ngober:
I think this looks good. Normally, I'd want to add tests before merging, but I think this is correct by inspection.

PR #357, opened 2023-06-09
Avoid counting multiple misses with MSHR full

jofepre:
Fixed cache miss count in case of an MSHR full. The miss should only be counted once rather than in every cycle until an MSHR becomes available.

### DIFF
--- a/src/cache.cc
+++ b/src/cache.cc
@@ -169,9 +169,7 @@ bool CACHE::try_hit(const tag_lookup_type& handle_pkt)
       ++sim_stats.pf_useful;
       way->prefetch = false;
     }
-  } else {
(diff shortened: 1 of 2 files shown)


ngober:
Good catch. Could you add a test for this? It looks like you might be able to modify test 406 to check the miss counts.

jofepre:
Hi Nathan! I’m sorry but I don’t really understand how the test framework works and thus I can’t generate a test easily and quickly...

ngober:
OK, I went in and added two tests: one that checks that the number of misses does not increase when the MSHR is full and one that checks that the number of misses does not increase when the lower level refuses the miss.

PR #367, opened 2023-08-02
Introduced strong types for time units

ngober:
This patch leverages the `std::chrono` facilities to provide a strong type for time units. Rather than tracking the raw integer number of cycles it has undergone, each operable now tracks the last (simulated) time that it operated. Latencies are now expressed as durations, rather than a raw integer number. While cycle counts and latencies used to be expressed using the same type, they are now expressed with different types. This is not only more expressive with regard to the source code, but it is bug-resistant: it is impossible to simply increment an event time, the time must be incremented by a duration.

This patch does deprecate the `current_cycle` member in favor of `current_time`. I converted the member into a deprecated function, which is likely to break some code. If a reviewer wants me to, I'll do some work to keep pre-existing code working, but I think the solution is just for modules to maintain their own cycle count, especially since #330 is coming eventually.


alberto-ros:
It will be important to consider that many prefetchers has access to the "current_cycle" variable. So it will require to adapt them to this new type. Not saying it is not good or bad, but just to add some info.

ngober:
Right. I think that in the long term it is better for those prefetchers to keep their own `current_cycle` variable if they need it.

PR #372, opened 2023-08-14
Modify branch type deduction.

jofepre:
Some colleagues and I are working on an improved cvp2champsim converter that better conveys some of the characteristics of the original CVP-1 traces to the ChampSim traces. A paper discussing the converter has been accepted for publication at the IISWC’23 conference. Sorry if my discussion is too verbose but I think it is important to understand the changes we propose.

One of the things that was missing in the original converter is the dependence of a branch from the previous instructions generating its source registers. When converting a branch from the CVP-1 trace, the original cvp2champsim sets the IP, SP, and flag registers as source and/or destination registers adequately so that ChampSim correctly identifies the branch type. In this process, however, the original source registers present in the CVP-1 trace are not preserved. This affects some conditional branches, indirect calls, and indirect jumps, which have general-purpose source registers in the CVP-1 trace. Consequently, the dependency between these branch instructions and producer instructions is lost. 

This is particularly important when a branch depends on a long-latency load. In this case, if the dependency is kept, a branch misprediction can only be corrected after the long latency load is resolved, exposing the branch misprediction penalty on the critical path. In addition, the negative impact on performance can be aggravated with a decoupled front-end and frequent BTB misses, as the prefetching effect of the decoupled front-end can be drastically reduced.

Our improved cvp2champsim converter preserves all the register dependencies among instructions present in the CVP-1 trace. To this end, in the case of branches, it adds the source registers present in the CVP-1 trace instruction to the required (depending on the branch type) IP, SP, and flag registers.

For conditional branches, we distinguish two cases. If the CVP-1 trace instruction has a source register, we add the source register to the converted instruction and avoid adding the flag register as the original cvp2champsim does. Otherwise, if the CVP-1 trace instruction does not have any source register, we keep adding the flag register, as done in the original converter. Unfortunately, replacing the flag register with the source register for some conditional branches, conflicts with the way ChampSim identifies them. 

The way we can address this shortcoming is by patching ChampSim to change the current condition to identify a conditional branch. In addition, we also need to modify the way ChampSim deduces indirect jumps as their check is performed before checking if a branch is conditional and we should avoid that conditional branches reading from others get classified as indirect jumps. 

The changes should be safe and keep compatibility with current traces. In these traces, conditional branches read from the flag register and thus they will still be identified as conditionals. Regarding indirect branches, in x86 they are always absolute and thus should not read the instruction pointer. We don't know traces for which the new branch type deduction breaks backward compatibility.


jofepre:
I've updated the cvp2champsim converter and its README based on our research paper "Rebasing Microarchitectural Research with Industry Traces", which will be published in the IISWC'23. This converter includes several improvements to perform a more accurate conversion from the original CVP traces to the ChampSim trace format.

ngober:
Thanks for sending this. I'm sure it will be very useful. I would actually love to see more tracing programs over time. It increases the use cases for ChampSim.

I have a few thoughts:
 - This isn't exactly your fault, but the code is written in a way that is difficult to test, and therefore difficult to guarantee. If we can factor out functional parts into smaller functions or distinct work units, we will be able to write tests and improve the tracer without sacrificing correctness.
 - You mention [a limitation in ChampSim's current trace format](https://github.com/ChampSim/ChampSim/blob/129def0f8aad10f435f501e104d7d2e1ab643280/tracer/cvp_converter/cvp2champsim.cc#L1553). Do you think this could be easily resolved by modifying the tracereader, or should this go into the list of features we want in a new trace format?
 - One of your cases has [both branches of an `if` the same](https://github.com/ChampSim/ChampSim/blob/129def0f8aad10f435f501e104d7d2e1ab643280/tracer/cvp_converter/cvp2champsim.cc#L564). Is this intentional?

I want to see this merged but I may want to work on it back and forth with you first so that we can make sure it's correct, which I suppose was your reason for writing the paper.

alberto-ros:
I will answer to point 2. In a new trace reader I wold argue that we need the type of branch, so all that guessing would not be necessary. But for now, in the current trace format it will be a good change.

We are aware of the new trace format being in the roadmap and we can "easily" adapt the converter to it.

jofepre:
Hi Nathan, let me reply to your thoughts:

- I understand your concerns about the code and its testing. I guess it happens frequently but when I wrote it, I didn’t think we would try to push it to the ChampSim repo. In addition, the code is overcomplicated and larger by the fact that it allows enabling and disabling the different conversion improvements we make. While this feature was important to check how they affected performance, I guess it is something we don’t need to have in the uploaded converter. If you agree, a simplification I can do is to clean the code and leave the converter with all improvements enabled. After that, we can discuss what else should be improved. 

- The limitation I point out in the code is related to the number of source and destination registers of instructions, which sometimes is not enough to convey all registers read or written by the CVP-1 trace instructions. It can be addressed in a new trace format but I don’t think it is a critical limitation as the number of affected instructions is small. Missing a memory access size or branch target fields are probably bigger limitations. And I completely agree with Alberto that the trace reader could be easily adapted to convert the CVP-1 traces to a different trace format.
 
- Regarding the two branches of an if statement with the same code, the short answer is yes, it is intentional. The cases are not exactly the same but the addressing mode is. We could merge them. 

My main concern about your comments is regarding the testing of correctness. I did my best to write a correct converter but I’m not sure how we can test it in a simple way.

ngober:
I'm very glad to see this work and I want to merge the best version of it that we can. Would you prune out the unnecessary options that you included for evaluation? Since the paper isn't released, you are the only one who knows what needs to stay and what needs to go.

After that, we'll work on the code quality. Here are some things that I have my eyes on:
 - The file information should not be kept in global state, but passed as parameters to the functions. This limits the scope in which the file can be changed.
 - Rather than piping the output to `xz`, `gzip`, etc, we should use the utilities in `inc/inf_stream.h`. Right now, I've only verified the decompression halves of these, but it shouldn't be too hard to verify the compression.
 - There are a lot of `for` loops that search through arrays and break on the correct index. These should be replaced with `std::find`, and other library algorithms. This is more expressive, that is, someone reading the code instantly knows what it does. Plus, it reduces the number of lines and therefore places for bugs to occur.

alberto-ros:
I just run simulations on all SPEC CPU 2017 traces with and without this fix. Branch information and other output does not change with this fix, so to me is ready to be accepted.

jofepre:
Since having the converter's code ready might take some time, I wonder if we should push the branch type deduction modification and move the converter to a new pull request. What do you think @ngober?

ngober:
I agree. It would probably have been best from the outset to approach these as two separate pull requests. I think the changes to `instruction.h` are mostly inconsequential and fine to merge. But, we should wait a bit for the converter.

ngober:
You will need to do the following

    git checkout modify-branch-type-deduction
    git branch update-cvp-converter
    git reset --hard ac6f8cf3570d4bc97ea6670f8f4d8b43524e219d
    git push -f origin modify-branch-type-deduction

This creates a new branch with both commits (`update-cvp-converter`), resets this branch to the first commit, and updates the pull request. The CVP converter changes will remain on the new branch.

ngober:
I've rebased this onto develop. I don't want to push incremental changes onto master too often. This will make its way into master before too long.

alberto-ros:
Hi Nathan, 

Researchers that use the new CVP traces with ChampSim, are mistakenly choosing the master branch (without this fix). Since this patch is backwards compatible, would not be good to have it in the master too?

PR #379, opened 2023-08-25
added google workload to champsim trace converter

hrishi-06:
In reference to [this](https://github.com/ChampSim/ChampSim/discussions/374#discussioncomment-6823770) conversation

### DIFF
--- a/google-workloads-to-champsim/README.md
+++ b/google-workloads-to-champsim/README.md
@@ -1,0 +1,95 @@
+# Google Workload Traces to Champsim Wrapper Script
+
+## Compile
+Use the following command to compile: 
+
+`g++ google-workload-to-champsim.cpp -lz` 
+
+## Run
+Use the following command to run the executable:
+
+`./a.out <trace>`
+
+`<trace>` should be a google workload trace file in compressed form with `.gz` extension. Google workload traces can be obtained [here](https://dynamorio.org/google_workload_traces.html).
+
+A new trace file compatible with champsim will be created. "champsim-" will be prefixed to the original trace file's name. This trace file will also be in compressed form with `.gz` extension.
+
+## About Wrapper Script
+
+### Google Workload Trace Format
+Google Workload traces are stored as records, each record being of type `trace_entry_t` and size 12 bytes. Following is the structure of `trace_entry_t`:
+
+```c
+struct _trace_entry_t {
+    unsigned short type;
+    unsigned short size;
+    union {
+        addr_t addr; 
+        unsigned char length[sizeof(addr_t)];
+        unsigned char encoding[sizeof(addr_t)];
+    };
+}
+```
+
+1. `type` : the type of entry. Instruction/Memory-Access/Marker/etc.
+2. `size` : memory reference size or instruction length
+3. `addr` : mem ref addr, instr pc, tid, pid, marker val
+4. `length` : not applicable in our context
+5. `encoding` : not applicable in our context
+
+Types of records can be broadly classified into following:
+
+1. Non-Branch Instructions
+2. Branch Instructions
+3. Memory Load
+4. Memory Store
+5. Marker
+6. Other
+
+> Google Workload traces guarantee that a memory access record will immediately be preceeded by its corresponding instruction
+
+### Champsim Trace Format
+Champsim traces are also stored as records, each record being of type `trace_instr_format_t` and size 64 bytes. Following is the structure of `trace_instr_format_t`:
+
+```c
+typedef struct trace_instr_format
+{
+    unsigned long long int ip; // instruction pointer (program counter) value
+
+    unsigned char is_branch;    // is this branch
+    unsigned char branch_taken; // if so, is this taken
+
+    unsigned char destination_registers[NUM_INSTR_DESTINATIONS]; // output registers
+    unsigned char source_registers[NUM_INSTR_SOURCES];           // input registers
+
+    unsigned long long int destination_memory[NUM_INSTR_DESTINATIONS]; // output memory
+    unsigned long long int source_memory[NUM_INSTR_SOURCES];           // input memory
+} trace_instr_format_t;
+```
+
+### Conversion
+
+Converting of Non-Branch instructions to champsim format is straightforward. We just copy instruction pointer value and set `is_branch=0` and `branch_taken=0`.
+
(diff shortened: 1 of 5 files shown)


google-workloads-to-champsim/README.md
```diff
@@ -0,0 +1,95 @@
...
+`./a.out <trace>`
+
+`<trace>` should be a google workload trace file in compressed form with `.gz` extension. Google workload traces can be obtained [here](https://dynamorio.org/google_workload_traces.html).
+
+A new trace file compatible with champsim will be created. "champsim-" will be prefixed to the original trace file's name. This trace file will also be in compressed form with `.gz` extension.
+
+## About Wrapper Script
+
+### Google Workload Trace Format
+Google Workload traces are stored as records, each record being of type `trace_entry_t` and size 12 bytes. Following is the structure of `trace_entry_t`:
```
derekbruening, line 20: For the conversion, we would highly recommend using our libraries which read the disk format into a record format suitable for analysis and simulation (the memref_t type).  The disk format is subject to change without considering that a compatibility change, as it is not meant to be a public relied-upon interface.  Furthermore, there are numerous complexities built into the disk format, such as including instruction encodings or physical address values only the first time observed, which are abstracted away by the libraries.  Trying to directly convert the disk format would require duplicating the handling of those complexities.  Best to work at the recommended and documented higher level for simplicity and to avoid breaking changes.

Thus, we would request to not take this approach and to instead create an analysis tool in our framework and convert from the memref_t level.  There are numerous changes at the disk level that have been made in the last year which break this converter as-is: it won't work on today's traces.

google-workloads-to-champsim/README.md
```diff
@@ -0,0 +1,95 @@
...
+    unsigned char source_registers[NUM_INSTR_SOURCES];           // input registers
+
+    unsigned long long int destination_memory[NUM_INSTR_DESTINATIONS]; // output memory
+    unsigned long long int source_memory[NUM_INSTR_SOURCES];           // input memory
+} trace_instr_format_t;
+```
+
+### Conversion
+
+Converting of Non-Branch instructions to champsim format is straightforward. We just copy instruction pointer value and set `is_branch=0` and `branch_taken=0`.
```
derekbruening, line 72: The drmemtrace records now have different types for taken or untaken conditional branches.

google-workloads-to-champsim/README.md
```diff
@@ -0,0 +1,95 @@
...
+Use the following command to compile: 
+
+`g++ google-workload-to-champsim.cpp -lz` 
+
+## Run
+Use the following command to run the executable:
+
+`./a.out <trace>`
+
+`<trace>` should be a google workload trace file in compressed form with `.gz` extension. Google workload traces can be obtained [here](https://dynamorio.org/google_workload_traces.html).
```
derekbruening, line 13: drmemtrace traces are now in .zip format.  Sample traces are in the https://github.com/DynamoRIO/drmemtrace_samples repository.

derekbruening (changes requested):

google-workloads-to-champsim/README.md
derekbruening: Going further, for our user-mode software-thread-oriented traces, we would recommend building a dynamic converter from memref_t, rather than an offline static converter, for use with our dynamic scheduler which re-schedules our traces onto the simulated virtual cores.  This solves several issues: it allows for mixing multiple workloads to model multi-tenant datacenters yet per-workload traces; it deflates the context switch inflation added due to tracing overhead; it (will) provide some functionality typically missing from trace-based simulation such as speculative path support by pulling instructions from prior points in the trace or using other schemes.

ngober: What do you suggest we do instead? Would this fall under the category of analysis tools, as described [here](https://dynamorio.org/sec_drcachesim_newtool.html)? Or is there a different way to get to a `memref_t`?

google-workloads-to-champsim/google-workload-to-champsim.cpp
```diff
@@ -0,0 +1,469 @@
...
+                    cs.destination_registers[0] = REG_STACK_POINTER;       // writes_sp
+                    cs.destination_registers[1] = REG_INSTRUCTION_POINTER; // writes_ip
+                }
+                else if (t.type == TRACE_TYPE_INSTR_INDIRECT_CALL)
+                {
+                    cs.source_registers[0] = REG_STACK_POINTER;            // reads_sp
+                    cs.source_registers[1] = REG_INSTRUCTION_POINTER;      // reads_ip
+                    cs.destination_registers[0] = REG_STACK_POINTER;       // writes_sp
+                    cs.destination_registers[1] = REG_INSTRUCTION_POINTER; // writes_ip
+                    cs.source_registers[2] = 99;                           // reads_other
```
ngober, line 233: I'm concerned that instructions do not get any register information unless they're branches. It implies that there are no data dependencies in the trace, which is almost certainly not true. Furthermore, all indirect branches have data dependencies on each other (on register 99), which could lead to some strange results.

ngober (changes requested):
This probably still needs work before it can be merged in. I think Derek's comments are important, and I also have some concerns about the ChampSim trace that is produced.

If we want to, we can merge this into its own branch and create a series of PRs against that. It would allow us to iterate on this over time.

google-workloads-to-champsim/README.md
derekbruening: Yes, an analysis tool.  If ChampSim files are per-software-thread, a parallel tool with software threads as the parallel shards would work.  You could start with an offline conversion to ChampSim files in the analysis tool.  The same analysis tool code could be run with the shards as virtual cores (possibly with a dynamic reschedule of the software threads) to produce per-hardware-thread trace files.

google-workloads-to-champsim/google-workload-to-champsim.cpp
biswabandan: I guess this is a limitation of the Google traces. Google traces do not have these info.

ngober: If we do not have this information, I'm not sure we can move forward converting the Google traces to ChampSim traces. Memory timing is extremely important, and register information is the way that ChampSim generates timing.

derekbruening: Regular drmemtrace traces have instruction encodings with this information so a converter is feasible there.  The version 1 release of the Google Workload traces had that information stripped out, and most core studies then had to perform limit studies assuming full deps vs zero deps.  We are working on a version 2 release of the Google workloads where we hope to get approval to release information on dependencies, though likely not full encodings.

PR #383, opened 2023-09-05
Add FR-FCFS policy to memory controller scheduler.

maccoymerrell:


maccoymerrell:
DRAM scheduler modified to utilize FR-FCFS scheduler, which should provide significant performance improvements on memory-intensive traces.

test/cpp/src/701-dram-scheduler.cc
```diff
@@ -0,0 +1,116 @@
...
+                }
+                completed_reqs++;
+            }
+            if(it->has_value())
+            {
+                last_scheduled[index] = it->value().scheduled;
+            }
+
+            index++;
+        }
```
ngober, line 57: I had to spend about an hour parsing this. Would it be simpler to do the following?

```suggestion
        std::vector<bool> next_scheduled{};
        std::transform(std::begin(uut->channels[0].RQ), std::end(uut->channels[0].RQ), std::back_inserter(next_scheduled), [](const auto& entry) { return entry.has_value() && entry.value().scheduled; });
        
        auto chunk_begin = std::begin(next_scheduled);
        auto chunk_end = std::end(next_scheduled);
        while (chunk_begin != chunk_end) {
            std::tie(chunk_begin, std::ignored) = std::mismatch(chunk_begin, chunk_end, std::cbegin(last_scheduled));
            if (chunk_begin != chunk_end && chunk_begin != request_locations[(*expected_order)[completed_reqs++]]) {
                return false;
            }
        }
        last_scheduled = next_scheduled;
```

test/cpp/src/701-dram-scheduler.cc
```diff
@@ -0,0 +1,116 @@
...
+            offset += champsim::lg2(DRAM_COLUMNS);
+            r.address += 0 << offset;
+            offset += champsim::lg2(DRAM_RANKS);
+            r.address += row_access[i] << offset;
+            r.instr_id = i;
+            r.response_requested = false;
+            packet_stream.push_back(r);
+        }
+        WHEN("The memory controller is operated") {
+            bool result = dram_test(&uut,&packet_stream,&expected_order,&arriv_time);
```
ngober, line 108: It may be more useful for this function to return `std::vector<uint64_t>` in the same kind as `expected_order`, then `REQUIRE_THAT(expected_order, Catch::Matchers::RangeEquals(observed_order));`. That would give more information on a failing test.

test/cpp/src/701-dram-scheduler.cc
```diff
@@ -0,0 +1,116 @@
...
+    //load requests into controller
+    for(auto it = packet_stream->begin(); it != packet_stream->end(); it++)
+    {
+        auto jt = uut->channels[0].RQ.begin() + vector_index;
+        *jt = DRAM_CHANNEL::request_type{*it};
+        jt->value().forward_checked = false;
+        jt->value().event_cycle = (*arriv_time)[vector_index];
+        request_locations.push_back(jt);
+        vector_index++;
+    }
```
ngober, line 25: It would be safer to use a transformation here rather than using pointer arithmetic directly.

```suggestion
    auto ins_begin = std::begin(uut->channels[0].RQ);
    //load requests into controller
    std::transform(std::cbegin(*packet_stream), std::cend(*packet_stream), std::cbegin(*arriv_time), ins_begin, [](auto pkt, uint64_t cycle) {
        pkt.forward_checked = false;
        pkt.event_cycle = cycle;
        return DRAM_CHANNEL::request_type{pkt};
    });

    std::vector<DRAM_CHANNEL::queue_type::iterator> request_locations(std::size(*packet_stream));
    std::iota(std::cbegin(request_locations), std::cend(request_locations), ins_begin);
```

test/cpp/src/701-dram-scheduler.cc
```diff
@@ -0,0 +1,116 @@
...
+
+    //carry out operates, make sure they complete in the expected order
+    uint64_t completed_reqs = 0;
+    std::vector<bool> last_scheduled(packet_stream->size(),false);
+    while(true)
```
ngober, line 30:
```suggestion
    while (completed_reqs < std::size(*packet_stream))
```

Put the condition in the place where the condition goes.

test/cpp/src/701-dram-scheduler.cc
```diff
@@ -0,0 +1,116 @@
...
+                last_scheduled[index] = it->value().scheduled;
+            }
+
+            index++;
+        }
+
+        //if we are done, break loop
+        if(completed_reqs == packet_stream->size())
+        break;
+    }
```
ngober, line 62:
```suggestion
```

Remove and put the condition at the top of the loop.

ngober (changes requested):
I'm glad this patch is here. We've needed someone to take a look at the DRAM model. I think the changes to the memory controller are pretty good (the use of `dram_get_*()` in the lambda is a bit weird, but I get it). I have some comments on the test code for your consideration.

test/cpp/src/701-dram-scheduler.cc
```diff
@@ -0,0 +1,109 @@
...
+        auto chunk_end = std::end(next_scheduled);
+        while (chunk_begin != chunk_end) 
+        {
+            std::tie(chunk_begin, std::ignore) = std::mismatch(chunk_begin, chunk_end, std::cbegin(last_scheduled));
+            //found newly scheduled request
+            if (chunk_begin != chunk_end)
+            {
+               scheduled_order.push_back(std::distance(std::begin(next_scheduled), chunk_begin));
+               completed_reqs++;
+               break;
```
ngober, line 50: This break is breaking out of L42, correct? I think we should allow the loop to continue. Is it possible that two packets might be serviced in the same cycle? If not, we should still probably allow the test to mark all newly scheduled packets, since if such a bug were to occur, the test would loop infinitely (on L30).

test/cpp/src/701-dram-scheduler.cc
```diff
@@ -0,0 +1,109 @@
+#include <catch.hpp>
+#include "mocks.hpp"
+#include "defaults.hpp"
+#include "dram_controller.h"    
+#include "champsim_constants.h"
+#include <algorithm>
+#include <cfenv>
+#include <cmath>
+
+#include <iostream>
```
ngober, line 10: Is this include being used for anything?

test/cpp/src/701-dram-scheduler.cc
maccoymerrell: It should not be possible for a channel to schedule multiple packets at once. Though a scheduler may be physically capable of doing such, physical DRAM would be limited to servicing one request at a time (per channel).

Removing this break caused the code to loop forever. I have tried for the past 3 hours to resolve this issue, but am unable to do so. It appears to be an issue with std::mismatch combined with std::transform and std::optional . Multiple packets are not being scheduled at once (as this would also break the current test, as we only record one schedule per operate() call, and thus we would have missing entries in our record vectors. Thankfully, we can expect this test to fail if we end up scheduling multiple packets at once.

I think additional work needs to be done in ChampSim to dissociate bank availability with data being returned. Data will return from the bank before the bank will be available again for more accesses. The turnaround time for memory accesses are fairly quick, as long as a bank is not accessed multiple times in a row, in which we will have to wait for the bank to be ready again. This is why bank interleaving is so important. Let me know if I am missing something in the code that actually respects this.
Currently, ChampSim doesn't do too bad at modeling this, except for :
1. The first access of each bank should see data return much faster than the bank becoming available again.
2. This time offset will carry forward to each data access to that bank following this (since we need to wait for bank availability).
To fix this:
1. We should try to associate an end-event-cycle with each bank which tells us at which cycle the bank will be available again. This can be tied or replace the valid signal each bank currently has.
2. The same can be done for packets coming through the memory controller, and arrive on the data bus at their end-event-cycle. My understanding of the memory controller is that it cannot issue any additional requests until the memory request (write or read) it is servicing is done and ready to be put onto the data bus. Data can be added to the read/write queues at maximum speed regardless. **This might explain the speedup we are seeing from the FR-FCFS algorithm.**
3. There might be other timings that need observed within the model that we are currently ignoring. I am fairly certain timing breakdowns differ with writes vs. reads as well.

PR #388, opened 2023-09-09
Added refresh cycles to memory controller, corrected minor timings

maccoymerrell:
Refresh cycles are now simulated within the memory controller, along with the associated performance loss. Refresh time is set to 64ms, with 8 rows per refresh command leading to refresh commands every 7.8us at default settings.
Associated performance gain from opening a row when all rows are closed is also added.

Test that asserts that refreshes are issued at the proper times added.


test/cpp/src/702-dram-refresh.cc
```diff
@@ -0,0 +1,94 @@
...
+    offset += LOG2_BLOCK_SIZE;
+    r.address += 0 << offset;
+    offset += champsim::lg2(DRAM_CHANNELS);
+    r.address += rand() % DRAM_BANKS << offset;
+    offset += champsim::lg2(DRAM_BANKS);
+    r.address += 1 << offset;
+    offset += champsim::lg2(DRAM_COLUMNS);
+    r.address += rand() % DRAM_RANKS << offset;
+    offset += champsim::lg2(DRAM_RANKS);
+    r.address += rand() % DRAM_ROWS << offset;
```
ngober, line 27: Avoid creating random packets in these tests. If you need variety, you can use Catch2's `GENERATE` directive, or create them procedurally. Random values in tests could potentially break in ways that are unrelated to a change being made.

test/cpp/src/702-dram-refresh.cc
```diff
@@ -0,0 +1,94 @@
...
+    {
+        //generate a random packet
+        generate_packet(channel_uut);
+        //operate mem controller
+        uut->_operate();
+        
+        //make sure that for every refresh cycle, each bank undergoes refresh at least once
+        std::vector<bool> bank_under_refresh;
+        std::transform(std::begin(uut->channels[0].bank_request),std::end(uut->channels[0].bank_request),std::back_inserter(bank_under_refresh),[](const auto& entry){return(entry.under_refresh);});
+        std::transform(std::begin(bank_refreshed),std::end(bank_refreshed),std::begin(bank_under_refresh),std::begin(bank_refreshed),[](const auto& lentry, const auto& rentry) { return (lentry || rentry);});
```
ngober, line 62:
```suggestion
        std::transform(std::begin(bank_refreshed),std::end(bank_refreshed),std::begin(bank_under_refresh),std::begin(bank_refreshed), std::logical_or<>{});
```

ngober (changes requested):
This looks pretty good. It changes a few defaults, and I want to make sure that those don't end up affecting other things.

test/cpp/src/702-dram-refresh.cc
maccoymerrell: I have rewritten the test to generate packets based on the current cycle.

maccoymerrell:
Not sure why these tests are failing

ngober:
Was this intentionally closed?

maccoymerrell:
Yes and no, I still have my local copy but the repo was deleted to fork from master for cache line bug fix.

ngober:
Ah, I see. You should probably use branches for that. Maintaining two histories is a nice feature of git.

PR #389, opened 2023-09-20
Fixed bug with cache line dirty flag being reset.

maccoymerrell:
Cache-lines could potentially have their dirty flag reset if a read-like access was made before write-back/eviction.
Along with this, a test was created to ensure proper cache write-back for lines that should be marked dirty.


ngober (approved):
This does appear to be a bug. Once the checks pass, it's good to merge.

PR #395, opened 2023-09-27
Updated the cvp2champsim converter and its README.

jofepre:
Some colleagues and I have developed an improved cvp2champsim converter, which better conveys some of the characteristics of the original CVP-1 traces to the ChampSim traces. A paper discussing the converter has been accepted for publication at the IISWC’23 conference. This pull request uploads our improved converter.


jofepre:
I made some improvements to the code of the converter. I eliminated all options and redundant code to enable and disable the conversion improvements we propose in our paper (now all of them are always enabled), avoid global variables, replace for loops with finds, etc. I'm sure the code can still be improved but I hope this is a step towards merging the converter in the ChampSim repo. Give it a look when you have time and let me know what you think.

Issue #343, opened 2023-05-18
Access to RQ and WQ sizes from the prefetcher


agusnt:
Hello,

I am trying to get the RQ/WQ size in the prefetcher_initialize() function. In previous version of development I done this with the get_size() function. However, in the current version I was not be able to find any way to do it.

May be possible to implement a get_[rq,wq,pq]_size()?

ngober:
That's correct. Ever since #303, `CACHE::get_size()` returned 0 for anything other than the MSHR. This is because each upper level now has its own set of queues, so it is ambiguous which queue's size would be returned. We could implement something, but what value would we expect from it?

agusnt:
The size of each queue. For example, if the configuration of L1D is the following one:
    ```
"L1D": {                                                                                                                                                                  
        "sets": 64,                                                                                                                                                           
        "ways": 12,                                                                                                                                                           
        "rq_size": 64,                                                                                                                                                        
        "wq_size": 64,                                                                                                                                                        
        "pq_size": 8,                                                                                                                                                         
        "mshr_size": 16,                                                                                                                                                      
        "latency": 5,                                                                                                                                                         
        "max_tag_check": 2,                                                                                                                                                   
        "max_fill": 2,                                                                                                                                                        
        "prefetch_as_load": false,                                                                                                                                            
        "virtual_prefetch": false,                                                                                                                                            
        "prefetch_activate": "LOAD,PREFETCH",                                                                                                                                 
        "prefetcher": "no"
    },
```

For RQ it should 64.

I do not know if this is the answer for your question.

ngober:
I see. I've been thinking about this since the patch I mentioned above. Right now, it's true that all of the queues to a given lower level from the upper levels have the same size, and it's true that the `"rq_size"` keys fix this. I could see how it is reasonable to just return that value.

But, I think eventually I want to move away from specifying the sizes directly in the configuration files. It is really easy to choose a bad configuration value here, for example a number that is less than the number of MSHRs in the upper level. I'm thinking instead to move to a calculated value that is the product of the number of MSHRs in the upper level and the hit latency of the lower level. Maybe then we can expose a `"rq_factor"` key that would then multiply this calculated value. I think it's a lot more expressive to say "I want to make the queue twice as big as it usually is" than to say "I want this queue to be 32 elements long now".

But, if you're researching this, I'll ask you: do you see benefit to configuring different queue sizes? Would it be better to specify the queue size in a different way?

[...]

PR #409, opened 2023-11-22
Fix: Varying queue sizes now varies queue sizes

GinoAC:
Fixed a bug where the queue sizes between cache levels was always the value of the LLC's queue sizes.

Reproducing the bug: Changing the L2's rq_size to 128 still resulted in it being set to 32 in the channel instantiation in .csconfig/[BUILD#]/inc/core_inst.inc. Also in core_inst.inc, without modifying the configuration file, the channels between each cache all had the same values despite differences in the default config file.

### DIFF
--- a/config/instantiation_file.py
+++ b/config/instantiation_file.py
@@ -353,8 +353,9 @@ def get_instantiation_lines(cores, caches, ptws, pmem, vmem):
     yield '#include "defaults.hpp"'
     yield '#include "vmem.h"'
     yield 'namespace champsim::configured {'
+
     struct_body = itertools.chain(
-        *((queue_fmtstr.format(name=ul_queues, **v) for ul_queues in v['upper_channels']) for v in upper_levels.values()),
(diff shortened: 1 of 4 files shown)


maccoymerrell:
Noticed some of this behavior when I was working with develop a month or so back.

Did you happen to check the other config options?

ngober (commented):
The Python construction that you chose is a bit arcane. Even if it was correct, somewhere along the line someone was going to change it. I pushed some changes of my own to the branch so that the behavior is tested now.

GinoAC:
> Noticed some of this behavior when I was working with develop a month or so back.
> 
> Did you happen to check the other config options?

Most of the other configs work through the cache_builder and appear to be functioning as expected.

GinoAC:
> The Python construction that you chose is a bit arcane. Even if it was correct, somewhere along the line someone was going to change it. I pushed some changes of my own to the branch so that the behavior is tested now.

Very cool, thanks! I got caught up in generators generating generators and thought I'd generate some generators too.

ngober:
> > Noticed some of this behavior when I was working with develop a month or so back.
> > Did you happen to check the other config options?
> 
> Most of the other configs work through the cache_builder and appear to be functioning as expected.

This is correct. The `"rq_size"`, `"wq_size"`, and `"pq_size"` keys take a different path through the file generation. (see #343)

PR #431, opened 2024-01-20
Fix PTW Deadlock Bug

maccoymerrell:
This bug resulted from the behavior of finishing translations in the cache and the inclusivity of the TLBs.
Due to pending feature requests involving inclusivity, this bug was resolved in a way that preserves the option of changing the inclusivity of the DTLB and STLB. 

When finishing translations in the cache, only packets that have matching virtual pages AND are untranslated are translated. This prevents in-flight PTW packets from having their physical addresses overwritten by finishing translations. Without this check, the MSHR will never resolve due to mismatching physical addresses, resulting in a deadlock.


ngober:
Would you please rebase this on develop? I think it's better to merge it there.

alberto-ros:
And thanks a lot for the nice fix :-)

PR #432, opened 2024-01-20
Ptw deadlock bug fix

maccoymerrell:
Modification made to finish_translate in cache to prevent PTW deadlock due to physical address overwrites during PTW walk.


maccoymerrell:
This request targets develop. Failing tests due to stat printout bug that was already in develop

agusnt:
@ngober I have been testing it for a few days, and it seems that the patch has solved the bugs. :)

PR #449, opened 2024-03-06
DIB hit buffer added

zmkzLe:
DIB hit buffer added


zmkzLe:
I used cmd+shift+L to replace the word "cycle" with "time" then I accidentally replaced all words I did not intend to replace so I did a second commit to the ooo_cpu.cc to fix that.

ngober:
This patch adds a bypass queue for instructions that hit the DIB.

It looks like the test that fails shows that instructions are getting reordered if they hit the DIB. This would need to be fixed before a merge.

zmkzLe:
I set it as 2 as default because currently, it does not take a cycle to promote and compare with the instructions in the decode buffer in the decode stage (another 1 cycle). The instructions that are hit in dib_check immediately advance to the dib-hit buffer at the promote stage.

ngober:
I was tinkering around with some tests and I stumbled on some that break this patch. I think the problem has something to do with the bandwidths in `O3_CPU::promote_to_decode()`. It looks like there could be some mismatches in there.

Issue #424, opened 2024-01-04
Runtime configuration


ngober:
Seems like this would still be tedious with Champsim's current need to recompile after altering the config file. I saw that simulators like Ramulator2 built their entire system based on the config file passed at runtime rather than hardcoding and selective compiling. The con of course is that you need to compile in all prefetching schemes, replacement policies, etc....

I have thought about something like this a lot when having to make 6 or 7 directories with different configurations of champsim. It feels wrong to just have a directory full of nothing but different champsim binaries with different configurations.

_Originally posted by @maccoymerrell in https://github.com/ChampSim/ChampSim/discussions/415#discussioncomment-7891215_

In response to this comment, I made some efforts to move the configuration to the run step, rather than the compile step. Progress can be tracked on the `feature/runtime-config` branch.

There are a few challenges that need to be overcome:

 - All constexpr references to `BLOCK_SIZE` or `PAGE_SIZE` need to be converted to object-local values. We can't have static sizes with a runtime configuration. We probably also need to be careful here to make sure that upper and lower levels either have the same block size or convert to their lower level's block size when sending a packet.
 - The DRAM model needs to take its parameters in its constructor
 - The build step needs a system that converts module names as strings to C++ object identifiers. This probably has a few surprises behind it.
 - A C++ replica of the Python configuration parser needs to be created. This will take a lot of work.
 - Generation of the files that support legacy modules must be maintained.

I have already succeeded on the last point. The makefile now generates the support files as necessary when building ChampSim.

The branch currently will probably not build a proper executable, because the instantiation file is no longer generated. The environment needs to be created by the C++ side instead. But, all the tests compile and pass so there is some good things happening.

I would appreciate any help on this. I worked on the first part out of sheer obsession today, but I probably can't afford to work on it much for a while.

ngober:
This branch has significant merge conflicts after the merge of #297. The `champsim::block_number` and similar types can no longer be statically defined, since `BLOCK_SIZE` is known only at runtime.

ngober:
The merge conflicts have been resolved. I developed a pattern where the runtime `BLOCK_SIZE` can be examined:

    struct block_offset_extent : champsim::dynamic_extent {
        block_offset_extent() : champsim::dynamic_extent(champsim::data::bits{champsim::lg2(BLOCK_SIZE)}, champsim::data::bits{}) {}
    };

[...]

PR #483, opened 2024-06-28
Event listeners

dpuckett98:
This pull request adds event listeners to ChampSim. Hooks are placed at key locations throughout ChampSim and whenever they are run, they generate events that contain relevant data and are passed to each registered listener. Listeners can be added to the new "listener" directory. Two example listeners, "debug" and "cpi_stacks_retire" are included in the pull request to illustrate both the functionality and usefulness of the proposed changes. All of the listeners in the listener directory are compiled in with ChampSim, but they are only activated when requested on the command line (for example, "--listeners debug,cpi_stacks_retire" will activate both example listeners).

These modifications do impact ChampSim's speed -- without any listeners activated, there is a slowdown of approximately 6-8% on the Olympus cluster (tested on perlbench-1273B and gcc-1850B from SPEC 2017).


ngober:
I haven't looked deeply into this, and I will later, but initially I like the design. I like the interface of `event_type+void*` as a way to get arbitrary information into arbitrary event listeners. I ought to pay homage to modern design and ask: did you look into `std::any` as an alternative to `void*`?

dpuckett98:
I hadn't heard of std::any until today, but after briefly looking at it online, it looks like it introduces some type safety, which could be useful for catching bugs. However, it looks like it may also dynamically allocate memory, and I am concerned about how that might hurt performance. My original implementation of the events dynamically allocated an object for every event, and it increased runtime by 10-15% (as compared to 6-8% in the current implementation)

src/cache.cc
```diff
@@ -71,6 +72,10 @@ CACHE::mshr_type CACHE::mshr_type::merge(mshr_type predecessor, mshr_type succes
     }
   }
```
ngober, line 73: Should these lines be moved into the debug listener?

src/event_listener.cc
```diff
@@ -0,0 +1,21 @@
+#include <fmt/chrono.h>
+#include <fmt/core.h>
+#include <fmt/ranges.h>
+
```
ngober, line 4:
```suggestion
```

These seem to not be used here.

config/parse.py
```diff
@@ -410,18 +411,24 @@ def prefetcher_parse(mod_name, cache):
             ).values()
         )
 
+        listeners = []
```
ngober, line 414: I'm not certain why the listeners are included as part of the configuration? In `main.cc`, you have given the option to the users to select which listeners to activate, so I assume that all listeners are compiled in always. That would make sense, but you add something here to the configuration as if you intend only to compile certain listeners, and not others.

ngober (commented):
I like the design of this overall. It's an excellent platform for future development. I've left some design comments about things to look at. My comment about moving print statements into the debug listener extends to most of the rest of the files in `src/`. It seems to me that the debug listener is doing the same thing as the print statements.

A note on passing `void*`. The reason that I don't like passing `void*` is that it seems to be prone to bugs. All it would take is one badly written listener and suddenly there is undefined behavior and memory corruption. Your comment about `std::any` is well-put, and your concerns about dynamic allocation is also well-founded. But, then I realized that the events are all known as compile-time, and so can be used in templates. I developed a sketch of how to utilize that fact [here](https://gcc.godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMRMAPVgB3oDpkpKqgYEAysib0QARgAsw0QQCqAZ0wAFTiADkABl2kAVnNK0mDUMgCkAJgBC9h6Q31kBPKMqN0AYVRaAFcWBhAAVjtSXwAZPAZMADkQgCNMYhB5AHZSHlQ1Qi8GAODQiKi8gs9ROITkljSM7NdMd2qGcQImYgISkLDIlraizu6CWqTU9MyctS6evrLBubGJ%2BsaZgEpXVCDiZHZdewBmeORgrABqG2O/IM9aQgBPG%2BwbfQBBd4%2BAeh/LgAimBoCUu6CYXUuBCePEwam%2Bc2IQQ8lwc2AA4gBJRIAfU0AAkPuJsDjwV1vjYsk5PpdaZdaKJgJdkE9zpgbtSPnT6YzLggmDwePE4fDjpzuWisbiCUSSWSmBAGVZmaQecqEJtLiBmaz6BBkNs%2BQKhQk1GoIBrrlTKQCKVlbWKKZ9EciCJdsIkAaSIUw7eK6ehdil6Jc8DxbI6aXSuHw8MhCO7Pd6uhBA0Fg5hQ5rtWHkBA8JrKU57XaHZzvsw2GoeEwDsz%2BSweAUWFbOX9AcDhZdMAA3RgEeGfRghZkWM3dvtiVvfWmS7F4wnE0gzxO2z42jlO37/D5m1DxiGZ%2BVQmFwy4Ad0ICAn/cHHwImEbFgf11uvf7rwRBCRKPfYmTvqRl8nwPk%2Bh6vn4n7Ot%2Bro3v%2B8o3H4f4ECAIBztKi5vMc2DTlGtJBAUyryji0Kwq%2BAKohi84ysSAGbuuJZAd8oF8OBiFQbuMG/pOBB0W%2BPGoR6Do4UWK74YRTLEaRmY3BRQl8eWjHlp87YAGIkMyoiTngjAHMxj6sS%2BiHIe6ABqHEEfETLIQBJHkSesKVpmNkIbc2DmdhqFSae9HAdulxcmo6bnEw44sARbppGCnYJOglyhZcYC6CF45BSk5xapcPDpo8yDujxcRzIw6SIWl5yvEl%2BlgUZtzSU5lwxBxKVqPl/aFQ%2BCTELhXJ0tlwZxiAYmXO2AAqewMPFmmNnQmAALSeGwcFuvERCXEiYh4ItCF4cN/zqJmVBBAwHhFJcqB9sQDJMPgyqrW4mAogQCDOTxQ0sc%2BMn8f2ZmqnVrCYPwgMBcQwCim8O09qgeBxfg1YQsgCAQB8IPwnYABs9ho4D/DxSjmxDaJO3cu2fgAEojZozKhW6T14C1K2oFCz2XGVY53ty3IrJ4yA4lIcyITEABUrwQLTaibLNryBji54kAA1hALk%2BiRiHuZ%2BVJzOgqEiMQ57dOgiHI6DIvdKDmzYza%2BNAdyG4MWWW7NZcaTAPEOK%2BDijxFZ1mV9blrViO1xXEIhzuu%2B7nsdSV2F%2BkNvsDUNo3PRolyHcd7QtbWBBBNItBPGCjNPZmsvEHLCd7cnqcnaIZ0XVdN1MndrQPW654IHG14matEeMFCqBl5cmJuiwTB51FaQEA%2BXWrTW46F8nCVoAw/Po2tELPVP/ITSkec9tIQTshDUMw6gMvy4rPG2QQxkCWhVEYbKrxW/6tKQ9D%2BenyX5/9pf1/9oJnqP18huJidstwrUuMPeIEBCzWh2gjXgzZUKhwYG7Bg6APZ00jl1WgcwGDB2tEpLc3IcEEDwfwWGNYCAI0QvAxsiCOA33QguB%2B2EIB2H0LINGqpZD6H0E/IaJCyEUPhggGhDYmybVQshf%2Bwk2H8FkPwu2uhti0D0OEIwYRdCyCMKgDg3AnwAyEChAwyjtjPWutMaBpA5YRH0OYPQ8gjAsDkLw0gmjDCkF0boIwagQB2OMYYbYcBYAwEQCgVA016BkAoPqCJQookgGALIY4shhB0Enr4iAKQ9AeJSPEboTw9DaNIHk5gxAngAHkUjGGbkUowaAWBsDEBUhgucclGCwCkIIwA/A518Vojpj5LDABkCY0g%2BBiDNzwH2fpHiuAPXuIcYpK1WjtPMHgFIxACkBCwGs78kiBmkAuikfImAgTDysI8Kw7TthUAsKDUyOlzwVMcnUpQYhJDSHYAod5qgNDaDWaYVJFhrm2EcM4dZKRfGQG2KgHg6c9CzQqccS4s1Zqa1kmCpwjhZBZB0RdYg0M4TwG2PdKuYRFZoMWGEVJsRhTrGmKkyohRRDUrMMy9oawpgZFSWS9oowFiBH6GYPlIx5jjHpdykV4q2W8vFVyhojLSW7H2N8lRaiNFrK8ZcTgAAONGs00byEuMAZAeVknyMuBAXAhANInAUUYAJpjSDmKwBkKxNjwh2NUboRxpBnFercVqvQPi/GkCdUE0JFoQC7AINlAg5BKANPiekRI/09B6oNUak1ZrLgWuKZgfARBCVa1SSID5UgZA/PLX8rQOgxlAtIOeLZPA6nqt0OooNYyvEVPuPGs6VAdX6sNca015rjiWogAESJ6Rrh2BSZsR1NyzGYAse69tfqA12PcTokNrgw0Ru2J671ehjiau7Xuw99jdB2HPR4rxS6TGbG2BdAo3h5BAA%3D%3D). Note the compiled assembly sees through `EventListener::dispatch()` and directly calls the subclass work function.

This way of doing it has two advantages to me: it avoids erasing any of the types through the call stack, and it reduces monstrous `else if` chains like you have in the debug listener (breaking them up into separate functions). Let me know what you think.

src/cache.cc
dpuckett98: Yes, my intention was for the debug listener to replace the existing debug print statements. Once I clean those out and update the pull request, I'll resolve this conversation

src/event_listener.cc
dpuckett98: They're a relic of the past. I'll remove them!

config/parse.py
dpuckett98: I instantiate the listeners in parse.py so I can compile them all in core_inst.cc.inc (see instantiation_file.py:452). Is there a simpler way to do that?

ngober: Yes, I think a better way is probably to move the `listeners/` directory to `src/listeners/`. This might even work straight out of the box, but you might need to add a line to the makefile.

The reason I raise this concern is because it's not a "module" in the same way that a prefetcher is, where a user might compile one in or not depending.

dpuckett98:
Thanks for your comments! I like your advice to use templates for each event -- I'll update the pull request to use them.

config/parse.py
dpuckett98: The listeners include both .cc and .h files, so it seems odd to me to put .h files in the src directory. Is that standard practice?

Additionally, aren't the listeners more similar to prefetchers, branch predictors, and replacement policies than the files found in src? The listener directory was modeled after the prefetcher directory in the first place.

ngober: Put the header files wherever they're findable, in `/include` or in `/src/listeners`. Personally, I prefer the latter, but this repository has traditionally used the former.

Concerning your second question, no, at least not in the sense of ChampSim's interface to the user. Those models are microarchitectural features that get integrated into ChampSim in a special way. Exactly what this way is has reasons for existing that are, again, mostly historical. Nevertheless, for the time being I think that these listeners should be considered internal functionality for now, and public interface once we work out what we mean by "modules". See my recent thoughts [here](https://github.com/ChampSim/ChampSim/issues/424#issuecomment-2353592540).

ngober:
Commenting here to make sure the relevant folks get a ping. I think there is an additional motivation for this patch in #594. It seems like it fits nicely into the same framework of ideas as this patch. I'd love to see both of them merged.

ngober:
Has there been any progress on this patch? The merge conflicts are starting to look hairy and I would hate to see this fall apart.

dpuckett98:
No more progress yet -- my April deadline morphed into a July deadline. I've been planning an (almost) total re-write anyway, so I don't expect the merge conflicts to be a big deal. I may make some progress on this before July, but I expect to finish the pull request sometime in August.

dpuckett98:
I've finished my above deadlines and am coming back to work on this PR. As you can see, I've started with a fresh repository due to the many changes in master and structural changes to the way I'm implementing this PR. The old code is at https://github.com/dpuckett98/ChampSim/tree/old_event_listeners.

I'm re-writing each event as its own function (rather than each event calling process_event with a void* to a unique struct filled with the relevant data) based on the comments above. I've also included the listeners in the src directory and register them automatically instead of implementing listeners as their own module, which results in simpler code and far fewer changes. The basic infrastructure should be set up -- I have verified it with one event and two listeners. I plan to implement the rest of the events and port over several useful listeners I've created over the next couple weeks.

Any comments on this new approach are welcome!

src/event_listeners.cc
```diff
@@ -0,0 +1,42 @@
...
+
+std::map<std::string, EventListener*(*)()>& get_registered_event_listeners() {
+  static std::map<std::string, EventListener*(*)()> registered_event_listeners;
+  return registered_event_listeners;
+}
+
+static std::vector<EventListener*> activated_event_listeners;
+
+void register_event_listener(std::string name, EventListener*(*create_el)()) {
+  get_registered_event_listeners()[name] = create_el;
```
ngober: I feel like this should be a `try_emplace`, but I note that there's no error handling for multiply-specified listeners. Maybe that doesn't matter?

src/event_listeners.cc
```diff
@@ -0,0 +1,42 @@
...
+#include <string>
+
+#include "event_listeners.h"
+
+std::map<std::string, EventListener*(*)()>& get_registered_event_listeners() {
+  static std::map<std::string, EventListener*(*)()> registered_event_listeners;
+  return registered_event_listeners;
+}
+
+static std::vector<EventListener*> activated_event_listeners;
```
ngober:
```suggestion
static std::vector<std::unique_ptr<EventListener>> activated_event_listeners;

```

If you do it this way, it basically makes `cleanup_event_listeners()` irrelevant. Yay for guaranteed destructor calls.

src/event_listeners.cc
```diff
@@ -0,0 +1,42 @@
...
+  activated_event_listeners.clear();
+  
+  // TODO: cleanup registered event listener constructors?
+}
+
+void handle_begin_phase(bool is_warmup) {
+  for (EventListener* e : activated_event_listeners) {
+    e->handle_begin_phase(is_warmup);
+  }
+}
```
ngober: In the snippet that I suggested previously, I used an enum constant-template-parameter for what you're doing here. Seeing this, I stand by that design suggestion. You're going to have to write this function for the cross product of every event that you want to observe and every listener that you create. That's going to get out of hand with just a few listeners. If you use a template here, you'll get generated code and may even be able to do some compile-time feature detection.

ngober (commented):
I'm excited for this! Thanks for the work that you've put into it.

I have a couple first-pass comments. I think it's worth particularly paying attention to how this system scales up to larger numbers of events.

Also, I think you should consider moving all of the listeners into a `src/listeners/` directory. There's probably going to be much more of these, so it'll be worth organizing them.

src/event_listeners.cc
dpuckett98: I'm excited for any way to reduce the amount of code I have to type, but I'm not sure the code snippet you sent me will work. I've tried implementing it, but I run into a problem because of the way I register and activate listeners. My current implementation puts each registered and activated listener into `vector<EventListener*>` then loops through each item in that list every time an event is generated. However, `vector<EventListener*>` is sub-class agnostic, so static casting the call to the subclass won't work. Do you have any suggestions for how to automatically register listeners while still using your suggested method?

src/event_listeners.cc
dpuckett98: It's conceivable that someone would accidentally register a listener with the same name as an existing listener, so I'll throw out a warning. The order that each listener is registered depends on the compilation order (right?), so I don't think it matters whether we insert or emplace.

src/event_listeners.cc
dpuckett98: Thanks for the suggestion, I'll use this!

src/event_listeners.cc
ngober: It depends on static initialization order, which is actually linker-dependent. You don't want to rely on it. I would say make a loud fuss if someone does weird things here, so that they know exactly where to go if weird things happen.

src/event_listeners.cc
ngober: I thought about it for a while, and you're right. It's not possible to do templates and virtual functions at the same time, for technical reasons (namely, template specialization, exactly what I was wanting to do). Now with that said, all of these things are known at compile time, so a tuple of all of the listeners zipped with a bool enable array would work, and the listeners wouldn't need to share a virtual base. The code is thornier (because tuples are somewhat harder to work with than vectors of base pointers), but should still be possible.

dpuckett98: It's probably worth it to do as much work as we can at compile time to avoid runtime overhead.

If we're building a tuple of listeners at compile time, then I can't dynamically register each listener at runtime anymore. The simplest solution to stick all the listeners into a tuple is to just manually add each listener to the tuple in event_listeners.cc -- what do you think?

ngober: You can still mark them as "enabled" at run time. Populate the entire set of listeners at compile time, then mark a boolean array for the ones the user wants.

dpuckett98: I'm trying to decide how to populate the entire set of listeners at compile time -- do I use ChampSim's modules, like the prefetchers and branch predictors use? Or do I make the user manually add the listener to the tuple?

ngober: Let's not overcomplicate. If this is going to be a user-facing feature, we can worry about it later. For whatever it's worth, there are plenty of kinds of work that people are doing where they're modifying source code.

dpuckett98: Sounds good to me!

dpuckett98:
I've re-worked most of the code based on our conversations above. Handling events now uses templates, so to add an event to file `X.cc` I only have to modify `X.cc` and any listeners that want to handle that event (previously, `event_listeners.cc` had to be modified as well).

Adding new listeners is now manual -- there are six lines of code a user has to add to `event_listeners.h` to hook up a new listener.

Listeners are no longer in `src/`, they're in `inc/listener/` (since they need to be included in `event_listeners.h`)

dpuckett98:
There are several things in the code I'm not sure about --
1. The `Event` enum is listed before the listeners are included in `event_listeners.h` (due to dependencies)
2. I had to mark functions and variables in `event_listeners.h` and `CPIPrintout.h` as inline to stop the linker from throwing errors.  This may mean there's a better way to implement this functionality.
3. It feels like there should be a way to automate `event_listeners.h` lines 40-45

inc/event_listeners.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include <vector>
+#include <string>
+#include <iostream>
+#include <tuple>
+
+enum Event {
+  BEGIN_PHASE,
+  DO_PREDICT_BRANCH
+};
```
ngober: The way to avoid this would be to put the events in their own header, which is included in each listener, too.

inc/event_listeners.h
```diff
@@ -0,0 +1,48 @@
...
+
+template<Event e, typename... Args>
+void handle_event(Args... args) {
+  if (listener_activation_map[0]) {
+    std::get<0>(listeners)->handle_event<e>(args...);
+  }
+  if (listener_activation_map[1]) {
+    std::get<1>(listeners)->handle_event<e>(args...);
+  }
+}
```
ngober: Are you ready for this? Tuples are fun sometimes.
```suggestion
template<Event e, std::size_t Idx, typename... Args>
void handle_listener_event(Args... args) {
    if (listener_activation_map[Idx]) {
        std::get<Idx>(listeners)->template handle_event<e>(args...);
    }
}

template<Event e, typename... Args>
void handle_event(Args... args) {
    [&] <std::size_t... Is> (std::index_sequence<Is...>){
        (handle_listener_event<e, Is>(args...), ...);
    }(std::make_index_sequence<std::tuple_size_v<decltype(listeners)>>{}); // immediately invoked
}
```

inc/listeners/test_listener.h
```diff
@@ -0,0 +1,31 @@
...
+}
+
+}
+
+class TestListener {
+public:
+  template<Event e, typename... Args>
+  void handle_event(Args... args) {
+    std::cout << "Test listener class handle event\n";
+    test_listener::handle_event<e>(args...);
```
ngober: You may eventually find that this is useful here:

```
if constexpr (e == BEGIN_PHASE) {
  // do stuff
} else {
  // do other stuff
}
```

inc/event_listeners.h
```diff
@@ -0,0 +1,48 @@
...
+enum Event {
+  BEGIN_PHASE,
+  DO_PREDICT_BRANCH
+};
+
+#include "listeners/cpi_printout.h"
+#include "listeners/test_listener.h"
+
+inline std::vector<bool> listener_activation_map;
+inline auto listeners = std::make_tuple(new CPIPrintout(), new TestListener());
```
ngober:
```suggestion
inline auto listeners = std::make_tuple(new CPIPrintout(), new TestListener());
inline std::bitset<std::tuple_size_v<decltype(listeners)>> listener_activation_map;
```

Might as well do something like this since the size is statically known.

ngober (commented):
I think this is a good foundation. If you don't want the linker to get all fussy with you about inline variables, you can declare them `extern` and put them in a source file somewhere. The reason that the linker is complaining is because the header gets copied into all files that include it, so you have multiple instances of a variable with the same name. Declaring them extern causes the compiler to skip instantiating it and waits for the linker to put things together.

dpuckett98:
> I think this is a good foundation. If you don't want the linker to get all fussy with you about inline variables, you can declare them `extern` and put them in a source file somewhere. The reason that the linker is complaining is because the header gets copied into all files that include it, so you have multiple instances of a variable with the same name. Declaring them extern causes the compiler to skip instantiating it and waits for the linker to put things together.

If I declare the variables extern then I can't use auto anymore to describe the type of the tuple... adding yet another place the user has to modify to add a new listener. I think I'll keep these functions as inline right now unless you have a strong opinion otherwise.

inc/event_listeners.h
dpuckett98: Makes sense, I'll do that!

inc/event_listeners.h
dpuckett98: I think it should be illegal to combine lambdas, templates, and variadic arguments all at the same time. The effect is very convenient, though. I didn't realize variadic arguments could extend to calling functions.

inc/listeners/test_listener.h
dpuckett98: Thanks for the pointer

ngober:
OK, now that we're at this point, we should talk about a merge roadmap. I think that we should merge this with a minimal listener (one might argue that has already happened) that works as a proof-of-concept. It doesn't have to be something that we keep long-term, but something that we can see working. Maybe reimplement the heartbeat as a listener? I'm open to ideas here. Then, we leave the task of writing other listeners and implementing other events to multiple later PRs. This would unblock #594 and it would make review easier.

dpuckett98:
I like the idea of implementing a heartbeat as a listener, although we should make it an "always-activated" listener so that users don't have to add "--listeners heartbeat" to all of their runs. Alternatively, I could add just the events to handle the debug printouts throughout ChampSim -- I could probably do that by early next week. Thoughts?

ngober:
My hesitation with the debug printouts is that it basically requires a comprehensive set of events right at the beginning. The heartbeat basically only needs to listen to the "instruction retired" event. I'm fine if this patch requires a bit of jank regarding being always-on. We can iron out the details as we go.

dpuckett98:
Okay, I'll work on that and push the minimum set of events/listeners to this PR.

dpuckett98:
I've changed the heartbeat to work as a listener, so this should the minimal code to implement the event listener framework. I'm marking this PR as ready for review!

inc/event_listeners.h
```diff
@@ -0,0 +1,49 @@
...
+#include <string>
+#include <iostream>
+#include <tuple>
+
+#include "events.h"
+
+#include "listeners/heartbeat.h"
+
+inline std::vector<std::string> listener_names = {"Heartbeat"};
+inline auto listeners = std::make_tuple(new Heartbeat());
```
ngober: Is there a reason to make this a tuple of pointers? Given that this is inline, I think there's nothing stopping you from making a tuple of objects.

inc/event_listeners.h
```diff
@@ -0,0 +1,49 @@
...
+#include <vector>
+#include <string>
+#include <iostream>
+#include <tuple>
+
+#include "events.h"
+
+#include "listeners/heartbeat.h"
+
+inline std::vector<std::string> listener_names = {"Heartbeat"};
```
ngober:
```suggestion
template <typename>
constexpr auto make_listener_names();

template <typename... Listeners>
constexpr auto make_listener_names<std::tuple<Listeners...>>() {
    return std::array<char const*, sizeof...(Listeners)>{{Listeners::cli_key...}};
}

constexpr inline auto listener_names = make_listener_names<decltype(listeners)>();

```
For your consideration: You may have an opportunity here to move the CLI key into the class (here, I've called it `Heartbeat::cli_key`). This is mostly just changing where the name is specified, but it means that one less line of this file needs to be changed. It also, but this doesn't matter, calculates the name at compile time rather than run time.

src/ooo_cpu.cc
```diff
@@ -225,11 +226,12 @@ void O3_CPU::do_check_dib(ooo_model_instr& instr)
   }
 
   instr.dib_checked = true;
-
+  
   if constexpr (champsim::debug_print) {
     fmt::print("[DIB] {} instr_id: {} ip: {} hit: {} cycle: {}\n", __func__, instr.instr_id, instr.ip, dib_result.has_value(),
                current_time.time_since_epoch() / clock_period);
   }
+
```
ngober:
```suggestion
```
Double check the whitespace additions here.

src/ooo_cpu.cc
```diff
@@ -52,8 +53,8 @@ long O3_CPU::operate()
   progress += check_dib();
   initialize_instruction();
 
-  // heartbeat
-  if (show_heartbeat && (num_retired >= (last_heartbeat_instr + STAT_PRINTING_PERIOD))) {
+  // heartbeat (in inc/listeners/heartbeat.h now)
```
ngober: You can go ahead and delete these things. The version control will keep them around.

inc/listeners/heartbeat.h
```diff
@@ -0,0 +1,91 @@
...
+#include <iostream>
+#include <deque>
+#include <vector>
+#include <chrono>
+#include <fmt/chrono.h>
+
+#include "events.h"
+#include "instruction.h"
+
+class Heartbeat {
```
ngober, line 13: If you add a constructor here that takes a `std::ostream&`, then we can control the print location.

inc/listeners/heartbeat.h
```diff
@@ -0,0 +1,91 @@
...
+
+template<>
+inline void handle_event<Event::BEGIN_PHASE>(Heartbeat* hb, bool& is_warmup) {
+  for (size_t i = 0; i < hb->switched_phase.size(); i++) {
+    hb->switched_phase[i] = true;
+  }
+}
+
+template<>
+inline void handle_event<Event::RETIRE>(Heartbeat* hb, uint32_t& cpu, std::deque<ooo_model_instr>::const_iterator& begin, std::deque<ooo_model_instr>::const_iterator& end, uint64_t& current_cycles) {
```
ngober, line 64:
```suggestion
template<typename ITER>
inline void handle_event<Event::RETIRE>(Heartbeat* hb, uint32_t& cpu, ITER begin, ITER end, uint64_t& current_cycles) {

```
It might be good not to depend on the exact type of the iterator.

inc/listeners/heartbeat.h
```diff
@@ -0,0 +1,91 @@
...
+  
+  if (hb->num_retired[cpu] > hb->num_retired_last_printout[cpu] + hb->cycles_between_printouts) {
+    
+    double heartbeat_instr = hb->num_retired[cpu] - hb->num_retired_last_printout[cpu];
+    double heartbeat_cycle = current_cycles - hb->cycles_last_printout[cpu];
+    
+    double phase_instr = hb->num_retired[cpu] - hb->num_retired_start_phase[cpu];
+    double phase_cycle = current_cycles - hb->cycles_start_phase[cpu];
+    
+    fmt::print("Heartbeat CPU {} instructions: {} cycles: {} heartbeat IPC: {:.4} cumulative IPC: {:.4} (Simulation time: {:%H hr %M min %S sec})\n", cpu,
```
ngober: `fmt::print` has an overload that takes a `std::ostream`. If you make it a `std::ostringstream`, then you can view it in a test.

ngober (commented):
> [Coverage Status](https://coveralls.io/builds/75296668)
> 
> coverage: 67.81% (-0.3%) from 68.139% when pulling **[dd4b1be](https://github.com/ChampSim/ChampSim/commit/dd4b1be7f5d53d45c5035d39c0dfa06b822ee00b) on dpuckett98:event_listeners** into **[f3db837](https://github.com/ChampSim/ChampSim/commit/f3db837911103c77c95512eb20502c6b442290e6) on ChampSim:develop**.

I was initially not too worried about this, but Coveralls's comment made me consider it more... You should write a few tests for this.

A test will probably look something like this (though this is a sketch):

```
TEST_CASE("The heartbeat listener prints one line after 1M instructions retired") {
    std::ostringstream stdout{};
    Heartbeat uut{stdout};
    for (int i = 0; i < 1000000; ++i) {
        std::vector<int> fake_instructions{{1,2,3,4}}; // This may need to be actual ooo_model_instrs
        uut.handle_event<Event::RETIRE>(0, std::cbegin(fake_instructions), std::cend(fake_instructions), i);
    }
    REQUIRE(stdout.str() == "ExPeCtEdOuTpUt");
}
```

Do a couple of things like this, testing the heartbeat frequency, making sure that the IPC is correct for the number of instructions retired, maybe also for multi-cpu.

inc/listeners/heartbeat.h
dpuckett98: It turns out you can't instantiate an std::ostream, because it's a wrapper for specific types of streams. So, I can't have a private member variable of type std::ostream because it involves a copy constructor. I decided to pass a pointer to an ostream instead, and that's working just fine.

inc/listeners/heartbeat.h
dpuckett98: I don't think we can do this, because partial specialization of function templates isn't allowed by C++.

ngober: Blast! Foiled again... I suppose it isn't necessary, I was mostly trying to avoid depending on the type when we only use it for the distance between begin and end.

dpuckett98:
I've added three tests -- one testing basic functionality, one testing correct behavior during phase change, and one testing multi-core behavior

test/cpp/src/020-heartbeat-listener.cc
```diff
@@ -0,0 +1,84 @@
...
+        uint32_t cpu = 0;
+        uint64_t curr_cycles = i;
+        auto cb = std::cbegin(fake_instructions);
+        auto ce = std::cend(fake_instructions);
+        uut.handle_event<Event::RETIRE>(cpu, cb, ce, curr_cycles);
+    }
+    
+    std::string res = stdout.str();
+    
+    REQUIRE(res.substr(0, res.find('(')) == "Heartbeat CPU 0 instructions: 10000000 cycles: 4999999 heartbeat IPC: 2 cumulative IPC: 2 ");
```
ngober: Catch2 provides [matchers](https://github.com/catchorg/Catch2/blob/devel/docs/matchers.md#stdstring-matchers) that may give more detailed error messages here:

```suggestion
    REQUIRE_THAT(res, Catch::Matchers::StartsWith("Heartbeat CPU 0 instructions: 10000000 cycles: 4999999 heartbeat IPC: 2 cumulative IPC: 2 "));

```

test/cpp/src/020-heartbeat-listener.cc
```diff
@@ -0,0 +1,84 @@
...
+    std::string rest = res.substr(res.find('\n')+1);
+    std::string l2 = rest.substr(0, rest.find('\n')+1);
+    rest = rest.substr(rest.find('\n')+1);
+    std::string l3 = rest.substr(0, rest.find('\n')+1);
+    
+    bool l1_correct = l1.substr(0, l1.find('(')) == "Heartbeat CPU 0 instructions: 10000000 cycles: 2499999 heartbeat IPC: 4 cumulative IPC: 4 ";
+    bool l2_correct = l2.substr(0, l2.find('(')) == "Heartbeat CPU 0 instructions: 20000000 cycles: 5999999 heartbeat IPC: 2.857 cumulative IPC: 2 ";
+    bool l3_correct = l3.substr(0, l3.find('(')) == "Heartbeat CPU 0 instructions: 30000000 cycles: 10999999 heartbeat IPC: 2 cumulative IPC: 2 ";
+    
+    REQUIRE((l1_correct && l2_correct && l3_correct));
```
ngober:
```suggestion
    REQUIRE_THAT(l1, Catch::Matchers::StartsWith("Heartbeat CPU 0 instructions: 10000000 cycles: 2499999 heartbeat IPC: 4 cumulative IPC: 4 "));
    REQUIRE_THAT(l2, Catch::Matchers::StartsWith("Heartbeat CPU 0 instructions: 20000000 cycles: 5999999 heartbeat IPC: 2.857 cumulative IPC: 2 "));
    REQUIRE_THAT(l3, Catch::Matchers::StartsWith("Heartbeat CPU 0 instructions: 30000000 cycles: 10999999 heartbeat IPC: 2 cumulative IPC: 2 "));

```

Best practice is to assert separately for each failure condition, that way you get more detailed feedback when something breaks.

inc/listeners/heartbeat.h
```diff
@@ -12,6 +12,12 @@
 
 class Heartbeat {
 public:
+  std::ostream* std_out;
+
+  Heartbeat(std::ostream* so) {
```
ngober:
```suggestion
  explicit Heartbeat(std::ostream* so) {

```

Danger: implicit conversions ahead.

inc/listeners/heartbeat.h
```diff
@@ -44,7 +50,7 @@ namespace heartbeat {
 
 template<Event e, typename... Args>
 inline void handle_event(Heartbeat* hb, Args&... args) {
-  //std::cout << "WARNING: generic handle event\n";
+  std::cout << "WARNING: generic handle event\n";
```
ngober: Was this change intentional?

ngober (commented):
I have some help to lend about tests. For the most part it looks good.

I want to flag something that may be important but currently does not cause big problems: Recent workflow runs [warn](https://github.com/ChampSim/ChampSim/actions/runs/17533664073/job/49850419410?pr=483#step:5:232) about the use of an explicitly-templated lambda, which is not available until C++20. That's my mistake, since it was my suggestion. However, it seems that [every compiler we test with](https://cppreference.com/w/cpp/compiler_support/20.html) supports this feature back to C++17. That will be worth paying attention to. I don't think upgrading to C++20 is bad (I've tried it locally, it would take some code changes), but may be on the roadmap if this causes problems.

inc/listeners/heartbeat.h
dpuckett98: Nope, good catch

test/cpp/src/020-heartbeat-listener.cc
dpuckett98: Didn't think to create separate assert statements. I'll implement this!

ngober:
🎉 Let's merge!

Next steps:
 - Add a few feature request issues that brainstorm good listeners to be clients of this system. Debug output is probably a good one. Stats collection is probably another. I'm sure we can come up with a few ideas. Issues make this trackable and hopefully leave breadcrumbs for interested contributors
 - Let's move forward on #594. Getting it merged will be a huge step of usability

PR #486, opened 2024-08-05
Berti Prefetcher

agusnt:
Berti prefetcher from MICRO 2022 adapted to the new master branch of ChampSim

### DIFF
--- a/prefetcher/berti/README.md
+++ b/prefetcher/berti/README.md
@@ -1,0 +1,12 @@
+# How to run Berti
+
+Ensure the L1D configuration in the `config.json` file looks like this before running Berti:
+
+```json
+"L1D": {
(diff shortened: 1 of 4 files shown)


ngober:
I'm really glad to see this. Would you create a separate pull request that adds your work to `/PUBLICATIONS_USING_CHAMPSIM.bib` so that your work can be referenced?

However, I'm conflicted about merging this into the trunk of ChampSim. If it gets merged in, we have to maintain it over time, and it becomes a burden. See the recent #495 as an example of things I don't want to be spending my time on.

I've had an idea to create a repository called, say, `champsim-contrib` that hosts all of the user-submitted modules, runs them occasionally to see if their performance changes, but provides no guarantees that I will be maintaining it in the future. What do you think about this?

PR #494, opened 2024-09-10
Dram size fix

maccoymerrell:
DRAM size is not being calculated properly in develop.

This code corrects this


ngober:
Could you add a test to validate this? I'd like to have something that says "We have X ranks, Y banks, etc... and we expect the size to be reported as ABC".

maccoymerrell:
Note here:
I didn't use GENERATE in these tests, since I would like to be able to use this test for Ramulator and integrated model checking. Ramulator has fixed sizes it can support, so having 3 tests for 3 specific sizes is doable, but generic size checking is not (at least when using Ramulator).

PR #498, opened 2024-09-13
Fix MacOS Compile Error

maccoymerrell:


ngober:
Looks like this didn't fix the issue. It still interprets `*.d` as a directory.

maccoymerrell:
I am going to try to make the other change a while back that I made to fix this on my branch.
Moved the directory rules below the .d rules.

maccoymerrell:
Bada bing

ngober:
I have no explanation for why this works.

PR #503, opened 2024-09-15
bug fixed: entries of DIB hit buffer and decode buffer exceed the limits

zmkzLe:
In the current implementation of promote_to_decode(), available_fetch_bandwidth does not account for the remaining space in the DIB_HIT_BUFFER and DECODE_BUFFER. As a result, the number of entries in these buffers can exceed their sizes. I have modified the bandwidth so it will consider the available space before promote them to decode.

### DIFF
--- a/src/ooo_cpu.cc
+++ b/src/ooo_cpu.cc
@@ -37,7 +37,6 @@ constexpr long long STAT_PRINTING_PERIOD = 10000000;
 long O3_CPU::operate()
 {
   long progress{0};
-
   progress += retire_rob();                    // retire
   progress += complete_inflight_instruction(); // finalize execution
   progress += execute_instruction();           // execute instructions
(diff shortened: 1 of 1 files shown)


src/ooo_cpu.cc
```diff
@@ -813,7 +814,7 @@ void LSQ_ENTRY::finish(std::deque<ooo_model_instr>::iterator begin, std::deque<o
 
 void LSQ_ENTRY::finish(ooo_model_instr& rob_entry) const
 {
-  assert(rob_entry.instr_id == this->instr_id);
+  // assert(rob_entry.instr_id == this->instr_id);
```
ngober, line 817:
```suggestion
  assert(rob_entry.instr_id == this->instr_id);
```

Is this left over from some earlier change? We probably should leave it uncommented if possible.

ngober (changes requested):

src/ooo_cpu.cc
zmkzLe: Oops my bad! I forgot to uncommented it out

PR #504, opened 2024-09-17
Added DRAM refresh, converted timings to cycles in DRAM controller, DRAM controller cycle correction

maccoymerrell:
Since the Ramulator integration pr might be out for a while, I decided to take all the changes made to the memory controller that have nothing to do with Ramulator from that pr and move them here. I also took the time to look at how the config file work for the memory controller and rework it so its more consistent with real parameters provided by DRAM manufacturers.

There are a lot of changes here:

- DRAM timings are now specified in cycles of the memory controller. This is how timings are specified in DRAM, so this is a consistency change so that writers can just copy/paste specs from DRAM modules. tRAS was added. I want to limit it to the four core DRAM timings so that the config file doesn't become massive.
- Frequency is changed to data_rate in the config. This differentiates the memory controller from the data bus. The data bus runs at the data rate, but the memory controller runs at half that. Since the data bus rate is the big frequency number advertised with DRAM, that's the one that is used in the config. Timings are based on the MC period, not the data bus.
- Constructors changed so MC operable runs at correct frequency.
- Turn-around time removed. Turn-around time is actually the sum of many different timings. tRAS is a decent approximation, and will now be used for that.
- These changes to timing are given some coverage by the dram_scheduler test.

- Address mapping struct created. The idea behind this was to pull all addressing out of the MEMORY_CONTROLLER and DRAM_CHANNEL classes and move them into a separate struct. This can help people modifying the dram addressing do it in a centralized and easy way.
- Columns are now defined in bits instead of cache blocks
- Block size is interpreted into a prefetch size based on the channel size of the DRAM. This explicitly states its the prefetch size of the memory controller in the code, which helps to clarify the internal operation of the memory controller.
- Originally, I wanted prefetch size to be a config parameter, but I think it adds way too much complexity to the memory controller to have a mismatch between block size and prefetch size.

- Refresh added to memory controller. This comes with two parameters. Refresh period (typically 32 or 64ms) and refreshes_per_period (this is how many refresh commands the memory controller issues per period, typically 8192). These are used to internally calculate tREF and rows_per_refresh, which control which rows are opened for refresh and when.
- Refreshes take tRP + (tRAS * prefetch_size) time
- Added refresh test

- Minor correction to access times on rows. If no row is open on row miss, access time is actually faster, which is now modeled.


maccoymerrell:
All these changes to the config may have some people worried. I would like to be able to catch config files that are still using the old format. Would the best way to do that be to change the name of the entry from "physical memory" to "DRAM" or something and then catch when a user has their block named "physical memory"?

src/dram_controller.cc
```diff
@@ -425,14 +500,15 @@ bool MEMORY_CONTROLLER::add_rq(const request_type& packet, champsim::channel* ul
 
 bool MEMORY_CONTROLLER::add_wq(const request_type& packet)
 {
-  auto& channel = channels[dram_get_channel(packet.address)];
+  auto& channel = channels[address_mapping.get_channel(packet.address)];
 
   // search for the empty index
-  if (auto wq_it = std::find_if_not(std::begin(channel.WQ), std::end(channel.WQ), [](const auto& pkt) { return pkt.has_value(); });
+  if (auto wq_it = std::find_if_not(std::begin(channel.WQ), std::end(channel.WQ), [this](const auto& pkt) { return pkt.has_value(); });
```
ngober, line 506: Why are you capturing `this` here?

test/cpp/src/702-dram-refresh.cc
```diff
@@ -0,0 +1,101 @@
...
+    while (uut->current_time < champsim::chrono::clock::time_point{} + (tREF*refresh_cycles))
+    {
+        //generate a random packet
+        generate_packet(channel_uut,(uint64_t)cycles, DRAM_CHANNELS, DRAM_RANKS, DRAM_BANKS, DRAM_COLUMNS, DRAM_ROWS);
+        //operate mem controller
+        uut->_operate();
+        cycles++;
+        //make sure that for every refresh cycle, each bank undergoes refresh at least once
+        std::vector<bool> bank_under_refresh;
+        std::transform(std::begin(uut->channels[0].bank_request),std::end(uut->channels[0].bank_request),std::back_inserter(bank_under_refresh),[](const auto& entry){return(entry.under_refresh);});
```
ngober, line 62: I'll cite [my earlier review](https://github.com/ChampSim/ChampSim/pull/390#discussion_r1759801874) and say that this test should measure externally-visible interfaces. If the refresh behavior is externally visible, you should make a test that causes it to happen. If it is not externally visible, we don't need to be modeling it.

config/parse.py
```diff
@@ -313,9 +313,9 @@ def transform_for_keys(element, keys, transform_func):
         )
 
         pmem = util.chain(self.pmem, {
```
ngober, line 315: There are ways to keep support for `"frequency"`, which defaults to half the data rate, if you want.

champsim_config.json
```diff
@@ -159,19 +159,21 @@
...
 
     "physical_memory": {
-        "frequency": 3200,
+        "data_rate": 3200,
         "channels": 1,
         "ranks": 1,
         "banks": 8,
         "rows": 65536,
-        "columns": 128,
+        "columns": 1024,
```
ngober, line 167: This change is pretty substantial, and I think it's worth at least considering changing the key so that the different behavior doesn't rug pull someone. Could this be `"column_width"`?

ngober (changes requested):
I like almost all of this. I'd like to see the test improved (I know this is kinda hard, but I insist), and I want to think together through the configuration change to "columns", but I like the rest. I like that you've lined up the constructor parameters with the way that memory is documented. It's going to make it easier to use.

src/dram_controller.cc
maccoymerrell: Oopsies

champsim_config.json
maccoymerrell: I think column_width implies that its the addressing bits width we are defining, which would be misleading.
I think we could change rows and columns to bank_rows and bank_columns which would be accurate.

I think if the config file specified a unit for all its numerical values, like a number followed by some string indicating the unit, that would resolve issues like this in the future as well. So users could define 1024b for bits or 128B for bytes. We could then also take care of the units not being clear from the config file. I actually really like this idea.

config/parse.py
maccoymerrell: I think these changes I am about to push will solve this. In my code, frequency is half the data rate. To the user, frequency or data_rate can be used in the config, but internally frequency sets the data_rate to frequency and then halves itself before being passed on to the rest of the config script.

I am not sure if there's a cleaner way to do this, but it allows for support of both.

test/cpp/src/702-dram-refresh.cc
maccoymerrell: It is externally visible and very important for performance modeling, just not in a way I would consider easy to test. We would be talking about hard-coding cycle counts like in 701-dram-scheduler (which I really hate, you basically have to rewrite a significant part of the test if you modify timing values). 
Measuring for increases in packet latencies as the controller operates would be the way to do it.

maccoymerrell: Remember the FR-FCFS scheduler has a significant impact on packet scheduling order when latencies change, ergo everything changes when latencies change.

champsim_config.json
ngober: Something like [this](https://github.com/ChampSim/ChampSim/blob/33dca7a859a273b04da10784b25faafc2acdfca8/config/parse.py#L93)?

ngober: I realize it isn't exactly what you're looking for, but the bones are there. I think `bank_rows` and `bank_columns` communicate better that these are rows per bank and columns per bank, if you don't already know that.

config/parse.py
ngober: Oh, I see, you've got an if-clause up the file a bit. Yea, it's kinda clunky but it works...

test/cpp/src/702-dram-refresh.cc
ngober: Let me paint it a different way: Say that someone comes along in the distant future and decides that they want to revise the DRAM model and improve it. Say that this person realizes a library exists that models different kinds of DRAM very well, and it's called "spamulator". The intrepid researcher builds a bridge between ChampSim and spamulator and wants to verify that they did it correctly. How should they test that ChampSim sees the refreshes it expects? The spamulator implementation does not include a `channels[0].bank_request[0].under_refresh` member. What is the adventurous poor researcher to do?

maccoymerrell: I think it would be wrong to say ChampSim "sees" any refreshes. Latencies spike, but from outside the memory controller, it cannot determine why. I think this could also apply to packet scheduling, row buffer hits, etc... 
By its construction alone, I don't think you can infer these things easily. Its not like a cache. DRAM has a lot of transient aspects. In ChampSim its doable because we know so much about the internal construction and the DRAM model is primitive. We aren't going into any of these tests blind.

In Ramulator, a verified external simulator, we don't know that information, but also I think its strange to be testing for correctness of an external simulator. If Ramulator says a packet took this long, then it took this long. Thats why Ramulator is being used. If we are going to do that, and act like Ramulator can't be trusted, then why not build a test suite inside ChampSim that fully verifies Ramulator?

I think test efforts for Ramulator should focus on the aspects of the interface between Ramulator and ChampSim we control. Refresh clearly falls outside that, and thus I don't think it can be tested. Packet return seems like a much better way to test the correctness of integration, along with checking internal stats of Ramulator (which I will update the tests to do).

If we are going to measure refresh from the external interface, we can do it with loose constraints but not tight ones.
We can expect that if we burst packets for a full tREF, there will at least 1 packet that have a spike in latency and at least 1 that doesn't. We can expect the difference in latencies to be around (tRP + tRAS) * rows refreshed. That is the best we can do.

For now, I would like to focus on testing refresh for our integrated model. The Ramulator pr can incorporate test modifications for Ramulator.

champsim_config.json
maccoymerrell: Yeah. I would like it to work for all of ChampSim's types if possible (and have a way to catch incompatible type conversions if possible). So instead of just bytes, also times, bits, and cycles.
I don't think that would be too difficult to implement, but it will take some tedious changes to the config scripts.

maccoymerrell:
I would say this is analogous to the cache tag checking conversation. We are actually modeling the ideal.

I am going to go on a tangent here to describe what the bankgroups are supposed to do, and how things are currently modeled.

Bankgroups are supposed to be used to hide the some of the prefetch latency for larger consecutive block accesses. Large prefetches take time for banks to issue, but are also very useful for maximizing available bandwidth. Bank groups essentially say "this set of banks is responsible for prefetching blocks with addresses that end with XX". The banks need some downtime after a prefetch burst, so we evenly spread out where the cache blocks go and hopefully no banks get a lot of accesses and we hide that additional latency.

If two cache blocks are requested from the same bankgroup back-to-back, there is some additional latency as compared to fetching from two separate bankgroups. What we are modeling is the ideal case, where the bankgroup is always different between accesses.

To add this, we need to do three things.
1. Track the last-used bankgroup
2. Incur a latency penalty for new accesses inside that bankgroup within a certain time period. I don't know exactly what this latency would be. I would need to look it up. Hopefully we can approximate it through other timing parameters, I don't want to add new ones if possible.
3. Add a bankgroup field to the addressing scheme. This should probably sit between the prefetch offset and the banks.

ngober:
That feels like not a huge departure from what we do now: we track the last open row and we get a latency benefit for hitting the buffer. I would be on the side of approximating the timing from other parameters. Not that they scale proportionally, but if someone actually for real wants to dig into this, they probably ought to be using a different tool.

maccoymerrell:
Okay. So a prefetch size of 8 incurs 8 clock cycles on the dbus to get the data out (4 cycles for the mc).
In reality, it actually takes closer to 10 or 12 clock cycles to pull it off. This number increases as our dbus frequency increases, but these are the numbers that I see. So I think we could model it as a static 33% increase in time to transmit on the dbus if we hit the same bankgroup twice. By that same token, once that time passes in the mc, we shouldn't incur that penalty at all, even if we hit the same bankgroup. I think I can manage this, give me a day or so.

maccoymerrell:
Bankgroups liveth  ⛈️ ⚰️

config/parse.py
```diff
@@ -267,6 +267,18 @@ def __init__(self, config_file, verbose=False):
...
+        if('frequency' in self.pmem.keys()):
+            self.pmem['data_rate'] = self.pmem['frequency']
+            self.pmem['frequency'] = self.pmem['frequency']/2
+        elif('data_rate' in self.pmem.keys()):
+            self.pmem['frequency'] = self.pmem['data_rate']/2
+        
+        if('columns' in self.pmem.keys()):
+            print('W: physical_memory, "columns" is deprecated, use "bank_columns" instead (set to columns*8)')
+        if('rows' in self.pmem.keys()):
+            print('W: physical_memory, "rows" is deprecated, use "bank_rows" instead')
```
ngober, line 281: I think you should use [this](https://github.com/ChampSim/ChampSim/blob/c90724c44405c1a553e2d66036c666fd107aea9a/config/parse.py#L159)

maccoymerrell: Functions are overrated. Think about how much compute is wasted on jumps 🧠. In a performative scripting language like Python, you need to consider these things 🧐

ngober: I wish there were some microarchitectural way we could predict the destination of jumps. It would probably enable testable, reusable code.

maccoymerrell: I don't think this function plays nice with pmem. I have tried for the last 30 minutes to get it to work, and it won't

maccoymerrell: Figured it out. I guess util.combine_named doesn't like scalar elements.

config/parse.py
```diff
@@ -168,7 +177,7 @@ def do_deprecation(element, deprecation_map):
     retval = { 'name': element['name'] }
     for old, new in deprecation_map.items():
         if old in element:
-            print(f'WARNING: key "{old}" in element {element["name"]} is deprecated. Use "{new}" instead.')
+            print(f'WARNING: key "{old}" in element {element["name"]} is deprecated. Use "{new}" instead. ' + (warning_msg_map[old] if old in warning_msg_map else ''))
```
ngober, line 180:
```suggestion
            print(f'WARNING: key "{old}" in element {element["name"]} is deprecated. Use "{new}" instead. {warning_msg_map.get(old, "")}'))
```

src/dram_controller.cc
```diff
@@ -317,12 +412,19 @@ void MEMORY_CONTROLLER::end_phase(unsigned cpu)
...
+}
+
 void DRAM_CHANNEL::check_write_collision()
 {
   for (auto wq_it = std::begin(WQ); wq_it != std::end(WQ); ++wq_it) {
     if (wq_it->has_value() && !wq_it->value().forward_checked) {
-      auto checker = [check_val = champsim::block_number{wq_it->value().address}](const auto& pkt) {
-        return pkt.has_value() && champsim::block_number{pkt->address} == check_val;
+      auto checker = [chan = this, check_val = wq_it->value().address](const auto& pkt) {
+        return pkt.has_value() && chan->address_mapping.is_collision(pkt.value().address,check_val);
```
ngober, line 427:
```suggestion
      auto checker = [addr_map=address_mapping, check_val = wq_it->value().address](const auto& pkt) {
        return pkt.has_value() && addr_map.is_collision(pkt.value().address,check_val);
```

This would be a more minimal capture

src/dram_controller.cc
```diff
@@ -343,18 +445,22 @@ void DRAM_CHANNEL::check_read_collision()
 {
   for (auto rq_it = std::begin(RQ); rq_it != std::end(RQ); ++rq_it) {
     if (rq_it->has_value() && !rq_it->value().forward_checked) {
-      auto checker = [check_val = champsim::block_number{rq_it->value().address}](const auto& x) {
-        return x.has_value() && champsim::block_number{x->address} == check_val;
+      auto checker = [chan = this, check_val = rq_it->value().address](const auto& x) {
+        return x.has_value() && chan->address_mapping.is_collision(x.value().address, check_val);
```
ngober, line 449:
```suggestion
      auto checker = [addr_map=address_mapping, check_val = rq_it->value().address](const auto& x) {
        return x.has_value() && addr_map.is_collision(x.value().address, check_val);
```

src/dram_controller.cc
```diff
@@ -27,37 +27,60 @@
...
   params.at(SLICER_ROW_IDX) = rows;
-  params.at(SLICER_COLUMN_IDX) = columns;
+  params.at(SLICER_COLUMN_IDX) = columns / pref_size;
   params.at(SLICER_RANK_IDX) = ranks;
   params.at(SLICER_BANK_IDX) = banks;
-  return std::apply([start = start_pos](auto... p) { return champsim::make_contiguous_extent_set(start, champsim::lg2(p)...); }, params);
+  params.at(SLICER_BANKGROUP_IDX) = bankgroups;
+  params.at(SLICER_CHANNEL_IDX) = channels;
+  params.at(SLICER_OFFSET_IDX) = channel_width.count() * pref_size;
+  return std::apply([start = 0](auto... p) { return champsim::make_contiguous_extent_set(start, champsim::lg2(p)...); }, params);
```
ngober, line 83:
```suggestion
  return std::apply([](auto... p) { return champsim::make_contiguous_extent_set(0, champsim::lg2(p)...); }, params);
```

ngober (approved):
I have only clerical suggestions remaining.

Eventually, I'd like you to write up a documentation page about the model, particularly the parameters. There are certain things that we should explain to the interested, such as the fact that the bankgroup stall delay is related to the prefetch size, and how the address map is laid out.

Issue #505, opened 2024-09-19
DRRIP Incorrectly Updates RRPVs


Quangmire:
**Describe the bug**
DRRIP currently updates RRPVs as follows:
```cpp
auto victim = std::max_element(begin, end);
for (auto it = begin; it != end; ++it)
  *it += ::maxRRPV - *victim; // Bug
```
After `it == victim`, all subsequent RRPV updates are 0.

**Expected behavior**
```cpp
auto victim = std::max_element(begin, end);
auto rrpv_update = ::maxRRPV - *victim;
for (auto it = begin; it != end; ++it)
  *it += rrpv_update;

ngober:
Oh, that's a sneaky bug. Good catch.

ngober:
Resolved by #506.

PR #506, opened 2024-09-19
Fix for Broken DRRIP RRPV Update

Quangmire:
See #505 for more details.

In addition to the fix, I also:
- Predicated the RRPV update with a check that it's non-zero
- Updated the SRRIP + SHiP RRPV update to have the same logic (which should be marginally faster)


ngober:
Could you please rebase this onto develop?

ngober:
Also, I'm curious if/why [test  444](https://github.com/ChampSim/ChampSim/blob/develop/test/cpp/src/444-srrip-replacement.cc) wouldn't catch this bug.

Quangmire:
In regards to the test not catching it, only SRRIP seems to use `srrip_set_helper` and not DRRIP / SHiP. 

Since all 3 of them use the same `victim` logic, it would probably be good to have a more generic class, `rrip_set_helper`, to eliminate the multiple implementations.

As for the `update` function, the options that come to mind are:
1. DRRIP / SRRIP / SHiP all subclass `rrip_set_helper` to implement `update`
2. `rrip_set_helper` can either implement
    - 1 `update` function that toggles behavior based on an `enum` argument (`rrip_type::{BRRIP,SRRIP,SHiP}`)
    - 3 separate functions (`srrip_update`, `brrip_update`, `ship_update`).
    - Both of the above would have the added benefit over (1) of eliminating having two independent SRRIP `update` implementations between DRRIP and SRRIP

ngober:
You're definitely onto something, and I've thought about this, but didn't know a good way to lay out the code. At present, the code for each of the modules is self-contained, which is good for people who are just getting started, but it means that code is duplicated, rather than being reused. I've thought about moving the code from `/replacement/srrip/` to `/replacement/rrip/srrip/`, etc. and putting the common code in `/replacement/rrip`. I haven't tested how this would work with the configuration file, I don't know if things would break if we specified

    { "L1D": { "replacement": "rrip/srrip" } }

maccoymerrell:
This has broken develop compilation

ngober:
Working on it!

PR #514, opened 2024-10-03
Update champsim_config.json to reflect low-grade DDR5 memory, fix timing issue with refresh

maccoymerrell:
With a release coming up, I would like to make a final tweak to the dram.
Default config numbers are updated and taken from real DDR5 at the 3200MHz frequency and at the config's density.
Refresh period has been reduced in half to 32ms, as this is standard for DDR5.

These newer updates revealed an issue with the timings as they stood. tRAS * rows_to_refresh should approximate the refresh delay somewhat better than previously. Previously I was doing (tRP + tRAS) * rows_to_refresh, but this is very conservative in terms of timing (20-30% too high). This doesn't become an issue until we start refreshing more often like in DDR5, and then it is very noticeable.

From what I can tell, tRFC (time after a refresh is issued before a bank can be used again) does not follow a strict tRAS + tRP per-row trend, as one might expect. It seems for refresh bursts can be more or less efficient than that with timing depending on density of the dram. Overall, this change gives us a decent approximation, but we can still be off by 10-20% at times.

There is another option here, which is to add tRFC as a config parameter and stop trying to approximate it from the other timings, but this is something that is not typically given by DRAM manufacturers up front (you can find it if you dig around your bios settings for long enough). Otherwise, we are stuck with this loose approximation. If someone is trying out things that are that dependent on DRAM configuration though, then they should probably be using ramulator as the memory model.

### DIFF
--- a/champsim_config.json
+++ b/champsim_config.json
@@ -162,19 +162,19 @@
         "data_rate": 3200,
         "channels": 1,
         "ranks": 1,
-        "bankgroups": 4,
-        "banks": 8,
+        "bankgroups": 8,
+        "banks": 4,
         "bank_rows": 65536,
         "bank_columns": 1024,
         "channel_width": 8,
         "wq_size": 64,
         "rq_size": 64,
-        "tRP":  18,
-        "tRCD": 18,
-        "tCAS": 18,
-        "tRAS": 38,
-        "refresh_period": 64,
-        "refreshes_per_period": 4096
+        "tCAS":  24,
+        "tRCD": 24,
+        "tRP": 24,
+        "tRAS": 52,
+        "refresh_period": 32,
+        "refreshes_per_period": 8192
     },
 
     "virtual_memory": {
--- a/config/parse.py
+++ b/config/parse.py
@@ -329,9 +329,9 @@ class NormalizedConfiguration:
         )
 
         pmem = util.chain(self.pmem, {
-            'name': 'DRAM', 'data_rate': 3200, 'frequency': 1600, 'channels': 1, 'ranks': 1, 'bankgroups': 4, 'banks': 8, 'bank_rows': 65536, 'bank_columns': 1024,
-            'channel_width': 8, 'wq_size': 64, 'rq_size': 64, 'tRP': 18, 'tRCD': 18, 'tCAS': 18, 'tRAS' : 38,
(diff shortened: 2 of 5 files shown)


maccoymerrell:
I think point 1 is something that will work fine for now. I think once runtime config is done (haha) it will give us a lot more options on how to pursue this.

I think points 2 and 3 are correct.
Maybe this ties back into getting ramulator integration done.
However, since that probably won't happen before the next release its important to make sure the existing model is in a decent state.

I have actually been plotting out real timing values for DDR and what the approximation will give us, and have made some tweaks to the algorithm that will hopefully manage decent accuracy.

Since scaling really happens with the density more than the # of rows, the following gives us a better approximation:
tRFC = SQRT(density in Gib) * tRAS

For DDR5, the difference is as follows:
Real: 8Gib = 115 ns; 16Gib = 130ns; 32Gib = 190ns
Approx: 8Gib = 91ns; 16Gib = 130ns; 32Gib = 184ns
This also uses DDR5's bank-level refresh (instead of rank-level) timings, which is also more accurate to what we are modeling.

ngober:
I think we're in agreement here. I would say that if you pick either the integrated model or the ramulator model to spend time on, you should pick the ramulator integration. It's going to pay off better in the long-term. Do you want me to leave this open for you to commit more changes?

maccoymerrell:
There should be 1 more coming through with this new timing algorithm for tRFC, and then it should be finished.

maccoymerrell:
Also, does ChampSim have a way to do integer square roots?

ngober:
> Also, does ChampSim have a way to do integer square roots?

No, there's an `ipow` in `util/bits.h`, but not an `isqrt`. [This page](https://stackoverflow.com/questions/4930307/fastest-way-to-get-the-integer-part-of-sqrtn) suggests that casting `std::sqrt` is good enough.

maccoymerrell:
@ngober should be ready after these checks pass

maccoymerrell:
@ngober also regarding config, I kind of like some aspects of how Ramulator does it. There are various DDR configurations inside ramulator that come from DDR standards, including timing.

Perhaps we should provide some set of configs that match real processors (Raptor, Alder, Golden, etc..), at least to a reasonable extent.
That way, even if they just use a default config, they are using a machine that's mostly real.
Caveats being core count and LLC sizing of course.

Issue #351, opened 2023-05-31
Trace Generator


GinoAC:
The goal of this feature is to allow for ChampSim to generate its own traces based on a set of parameters defined at runtime. This would mainly be beneficial for testing modules and rapid prototyping.

ngober:
I've had a good think about this one. I think writing a trace generator is probably extremely valuable in the long run. There is an idea in software testing called fuzzing where you write a test that varies its input with certain parameters in order to tease out bugs that are hard to see normally. 

One example is the FuzzTest framework: https://github.com/google/fuzztest

This can serve as a starting point for a trace generator. It starts as an integration test: we feed the simulator randomized traces (that may or may not be parameterized), and we wait for ChampSim to crash. If it does not, that increases our confidence in its correctness. As our capabilities grow, we can extend this to well-structured parameterizations, and eventually I think the fuzz testing and the trace generation could begin to look pretty similar.

GinoAC:
I think that's a good game plan for this moving forward. Do you think there are any parameters required for the basic trace gen/fuzzer? 

I believe I have a prototype trace generator that just takes in basic block size and data contiguity. I'll dig through it and see if there's anything useful I can push as a starting point.

ngober:
We can parameterize the fuzz tester if we want, but a basic fuzz test is entirely "The program does not catastrophically fail when given completely random inputs". From there, we can say "The program doesn't fail when the input is bounded in X way" or "The program raises Y error when the input is bounded in Z way". Later on, we can get to "The data accesses fit within the L1 cache", or other particular things.

GinoAC:
I agree, this is a good starting point. I was thinking more that the fuzz won't have a high coverage if we have purely random accesses since it will reduce collisions in specific structures (e.g. We'll likely reach the MSHR capacity but it might be unlikely to trigger an MSHR hit).

ngober:
That's true, but we also already have a test that performs MSHR hits. The point of fuzz testing is to illuminate weird cases that your tests don't already cover.

GinoAC:
That's partially my point, we need these events to occur simultaneously during fuzzing. I've played with fuzzing before and if we don't have some direction to it to incur the weird corner cases, it can potentially diverge into just random accesses that don't place strain on multiple structures at the same time. I guess I personally lean more towards mutation based fuzzing though.

ngober:
Ah, I see your point. There's probably something we can do, but if you have some kind of generator in progress, that might make a better starting point.

PR #521, opened 2024-10-12
Attempt to Fix AMPM Performance

maccoymerrell:
AMPM simulation performance is currently awful. It is nearly impossible to run simulations with it, and many simulations simply stop after a short period of time without a deadlock notification. IPC speedup seems incredibly good for certain traces, but those traces quickly livelock.

This is an attempted fix for what is causing this.

I have replaced the region-lru system with an lru-table with 128 ways. I don't believe I have modified anything else about the prefetcher besides this.

The resulting simulation time is much shorter, 50x+ faster for gcc-2226B.
However, it is currently failing the AMPM test.

I would appreciate a second set of eyes on this code to help identify why this behavior is different than the previous one, outside of different eviction policy.

Another note:
With this change, I ran a sample trace for 300k instructions and got the following results for prefetch activity:
cpu0_L2C TOTAL        ACCESS:      21353 HIT:        115 MISS:      21238
cpu0_L2C LOAD         ACCESS:      21138 HIT:         74 MISS:      21064
cpu0_L2C RFO          ACCESS:          1 HIT:          0 MISS:          1
cpu0_L2C PREFETCH     ACCESS:         92 HIT:          0 MISS:         92
cpu0_L2C WRITE        ACCESS:         29 HIT:         29 MISS:          0
cpu0_L2C TRANSLATION  ACCESS:         93 HIT:         12 MISS:         81
cpu0_L2C PREFETCH REQUESTED:         92 ISSUED:         92 USEFUL:         80 USELESS:          7

Without this change, for the same trace for 300k instructions:
cpu0_L2C TOTAL        ACCESS:      45381 HIT:       4910 MISS:      40471
cpu0_L2C LOAD         ACCESS:      21137 HIT:       4871 MISS:      16266
cpu0_L2C RFO          ACCESS:          1 HIT:          0 MISS:          1
cpu0_L2C PREFETCH     ACCESS:      24133 HIT:          3 MISS:      24130
cpu0_L2C WRITE        ACCESS:         29 HIT:         29 MISS:          0
cpu0_L2C TRANSLATION  ACCESS:         81 HIT:          7 MISS:         74
cpu0_L2C PREFETCH REQUESTED:     579638 ISSUED:      24172 USEFUL:       5590 USELESS:        322

This is a very different amount of prefetcher activity, but once again I don't see why this change would happen.

### DIFF
--- a/prefetcher/va_ampm_lite/va_ampm_lite.cc
+++ b/prefetcher/va_ampm_lite/va_ampm_lite.cc
@@ -4,7 +4,6 @@
 
 #include "cache.h"
 
-uint64_t va_ampm_lite::region_type::region_lru = 0;
 
 template <typename T>
 auto va_ampm_lite::page_and_offset(T addr) -> std::pair<champsim::page_number, block_in_page>
@@ -15,17 +14,17 @@ auto va_ampm_lite::page_and_offset(T addr) -> std::pair<champsim::page_number, b
 bool va_ampm_lite::check_cl_access(champsim::block_number v_addr)
 {
   auto [vpn, page_offset] = page_and_offset(v_addr);
-  auto region = std::find_if(std::begin(regions), std::end(regions), [vpn = vpn](auto x) { return x.vpn == vpn; });
+  auto region = regions.check_hit(region_type{vpn});
 
-  return (region != std::end(regions)) && region->access_map.at(page_offset.to<std::size_t>());
+  return (region.has_value() && region->access_map.at(page_offset.to<std::size_t>()));
 }
 
 bool va_ampm_lite::check_cl_prefetch(champsim::block_number v_addr)
 {
   auto [vpn, page_offset] = page_and_offset(v_addr);
-  auto region = std::find_if(std::begin(regions), std::end(regions), [vpn = vpn](auto x) { return x.vpn == vpn; });
+  auto region = regions.check_hit(region_type{vpn});
 
-  return (region != std::end(regions)) && region->prefetch_map.at(page_offset.to<std::size_t>());
+  return (region.has_value() && region->prefetch_map.at(page_offset.to<std::size_t>()));
 }
 
 uint32_t va_ampm_lite::prefetcher_cache_operate(champsim::address addr, champsim::address ip, uint8_t cache_hit, bool useful_prefetch, access_type type,
@@ -33,15 +32,13 @@ uint32_t va_ampm_lite::prefetcher_cache_operate(champsim::address addr, champsim
 {
   auto [current_vpn, page_offset] = page_and_offset(addr);
   champsim::block_number block_addr{addr};
-  auto demand_region = std::find_if(std::begin(regions), std::end(regions), [vpn = current_vpn](auto x) { return x.vpn == vpn; });
+  auto demand_region = regions.check_hit(region_type{current_vpn});
 
-  if (demand_region == std::end(regions)) {
+  if (!demand_region.has_value()) {
     // not tracking this region yet, so replace the LRU region
-    demand_region = std::min_element(std::begin(regions), std::end(regions), [](auto x, auto y) { return x.lru < y.lru; });
-    *demand_region = region_type{current_vpn};
+    regions.fill(region_type{current_vpn});
     return metadata_in;
   }
-
   // mark this demand access
   demand_region->access_map.at(page_offset.to<std::size_t>()) = true;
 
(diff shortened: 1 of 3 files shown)


prefetcher/va_ampm_lite/va_ampm_lite.cc
```diff
@@ -15,33 +14,31 @@ auto va_ampm_lite::page_and_offset(T addr) -> std::pair<champsim::page_number, b
...
+  return (region.has_value() && region->prefetch_map.at(page_offset.to<std::size_t>()));
 }
 
 uint32_t va_ampm_lite::prefetcher_cache_operate(champsim::address addr, champsim::address ip, uint8_t cache_hit, bool useful_prefetch, access_type type,
                                                 uint32_t metadata_in)
 {
   auto [current_vpn, page_offset] = page_and_offset(addr);
   champsim::block_number block_addr{addr};
-  auto demand_region = std::find_if(std::begin(regions), std::end(regions), [vpn = current_vpn](auto x) { return x.vpn == vpn; });
+  auto demand_region = regions.check_hit(region_type{current_vpn});
```
ngober, line 35: I think the issue is probably here. This region is returned by value, so when it gets updated, only the copy is updated, so the AM/PM doesn't get written back.

maccoymerrell: Okay. It feels bad, but I just call regions.fill() to write back the change.
I guess it makes sense that it doesn't pass by reference, since if you are going to modify a cache value it should be done as a fill.

I am running some test now. I will let you know if this fixes the problems.

ngober: That's the standard usage. This is the same thing that would happen for a hardware array: it needs to be written back after being updated.

maccoymerrell:
After these changes, for the trace that I was having trouble with, the runtime to get to 30M instructions has gone from indefinite (3 hrs++) to around 16 minutes. IPC seems to match pretty closely to before this change.

I am not sure why this is faster. In my mind, the lru table is doing far more work than what ampm was doing. Oh well

ngober:
It's hard to argue with benchmark results. Do you think it's worth writing tests for this? Or benchmarks? Catch2 can do benchmarks.

maccoymerrell:
I think that's probably worthwhile. Being able to see performance for commits could help us increase ChampSims performance as we continue with future updates (or at least stop it from getting worse).

It would also be nice to log simulation results for a couple of the traces as well after each commit. That would have helped catch the cache dirty-bit reset bug a while back. Even if we can't prevent a pull because of a change in the stats, when we are searching for a bug it would reduce the time to track it down a lot.

ngober:
I agree entirely with your point. That was part of my intention with [this CI flow](https://github.com/ChampSim/ChampSim/blob/master/.github/workflows/buildcheck.yml). First of all, it catches some bugs the unit tests don't; second of all it tests the build system; third, it archives the JSON results file. All we would need is some reporting on the back end (maybe put it in the documentation?) that plots the drift over time.

This was also my sole purpose behind #351. If we can manufacture workloads, we can do a bunch of drift checks with very particular behaviors (cc: #513). Most of the time users will want traces, but there are some times when we the maintainers want very specific behaviors.

But, my earlier comment was more fine-grained. We can write a function-level benchmark to see if the runtime of `va_ampm_lite::prefetcher_cache_operate()` increases suddenly. See [the Catch2 docs](https://github.com/catchorg/Catch2/blob/v2.x/docs/benchmarks.md#top).

maccoymerrell:
@ngober I have added benchmark tests to the test suite for AMPM.
These invoke each of the prefetcher functions and will record their runtimes in the test output.
Right now, these tests feed in default values {} for each field of each function.

maccoymerrell:
Curse you curl

test/cpp/src/453-va-ampm-lite-behavior.cc
```diff
@@ -96,3 +96,91 @@ SCENARIO("The va_ampm_lite prefetcher issues prefetches when addresses stride in
...
+    CACHE uut{champsim::cache_builder{champsim::defaults::default_l1d}
+      .name("453-uut-benchmark[va_ampm_lite::prefetcher_final_stats()]")
+      .upper_levels({&mock_ul.queues})
+      .lower_level(&mock_ll.queues)
+      .lower_translate(&mock_lt.queues)
+      .prefetcher<va_ampm_lite>()
+    };
+    meter.measure([&] { return uut.impl_prefetcher_final_stats(); });
+  };
+  REQUIRE(true);
```
ngober, line 185:
```suggestion
  SUCCEED();
```

ngober:
I'm glad this works so easily, and outputs its information nicely. I agree, we need some sort of infrastructure to collate the results. [This action](https://github.com/benchmark-action/github-action-benchmark) looks promising, but I haven't looked too deeply at it yet. I'll put it on my todo-list. For now, we can call this done and say we have made progress.

maccoymerrell:
image

PR #526, opened 2024-10-22
Fix for latency issue in cache builders.

maccoymerrell:
This makes total latency take priority over hit and fill latencies when all 3 are specified.
This seems like the simplest solution to the problems experienced from cache defaults.


maccoymerrell:
I think another alternative to this would be to have the defaults define only the total latency, and having hit and fill latency take priority over total latency, like it is now.

ngober:
I think that the thing I would most rather do is to situate the size-based latencies so that they default to acceptable values. I'm thinking about the user experience as falling into one of three situations:

1. I need an X kiB cache, and I don't care much about the latency details
2. I need a cache with Y cycles of latency, but I don't know the details about the tag/fill pipelines
3. I know the hit latency and fill latency of my cache. I want to specify them precisely.

Those users are operating with increasing levels of detail. Your patch swaps the order of 2 and 3. I think that the instances specified in `defaults.hpp` should focus on getting 1 right, so that when the user comes in with 2 or 3, they get the behavior they expect.

Now, the hard question is: what are the good latency values given X? I don't know that answer right now.

maccoymerrell:
Here are some experimental results, these are averages hence the partial clock cycles:

xeon platinum 8360
L1D (48 KiB): 3.5 cycles, 1.6ns
L2C (1.3 MiB): 12.8 cycles, 4.2ns
LLC (54 MiB): 38.7 cycles, 20.1ns

intel i7 12700k
L1D (48 KiB): 5 cycles (1.4ns)
L2C (1.3 MiB): 10 cycles (2.8ns)
LLC (25 MiB): 45 cycles (12.6ns)

AMD 7700X
L1D (32 KiB): 3.3 cycles (0.7ns)
L2C (1 MiB): 17 cycles (3.8ns)
LLC (32 MiB): 50 cycles (11.1ns)

ngober:
I suppose these are round-trip times observed by the core?

maccoymerrell:
I have adjusted them from round-trip times. So the LLC latency I quote there is the observed l3 latency - observed l2 latency

maccoymerrell:
I am using Intel's MLC tool to spam the memory with random accesses from a certain sized buffer.  To test each level, I set the buffer size to halfway between the previous cache level and cache level I am trying to measure.

maccoymerrell:
cache_latencies

maccoymerrell:
I guess a breakdown by sets and ways instead of just size would probably give a better trend. I don't feel like doing that right now though.

ngober:
The cycle equation is nice, and logarithmic complexity matches intuition. It would be better if it were base 2, though. The coefficient on that would then be 7.42.

maccoymerrell:
The equation that I have come up with is:
latency (cycles) = LOGBASE(2,x(KiB)) * 3.585 - 15

This is pretty spot on for L1, LLC latencies. L2 latencies are overestimated pretty significantly. It seems like the curve doesn't seem to be quite logarithmic, more like a hockey stick in the log space and the L2 seems to be at the elbow.
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">

<html>
<head>
	
	
	<title></title>
	<meta name="generator" content="LibreOffice 6.0.7.3 (Linux)"/>
	<style type="text/css">
		body,div,table,thead,tbody,tfoot,tr,th,td,p { font-family:"Liberation Sans"; font-size:x-small }
		a.comment-indicator:hover + comment { background:#ffd; position:absolute; display:block; border:1px solid black; padding:0.5em;  } 
		a.comment-indicator { background:red; display:inline-block; border:1px solid black; width:0.5em; height:0.5em;  } 
		comment { display:none;  } 
	</style>
	
</head>

<body>

Size | Latency (ns) | Latency (cycles) | ChampSim
-- | -- | -- | --
48 | 1.6 | 3.5 | 5.02325325989571
1280 | 4.2 | 12.8 | 22.0062610711666
49152 | 20.1 | 38.7 | 40.8753350909021
48 | 1.4 | 5 | 5.02325325989571
1280 | 2.8 | 10 | 22.0062610711666
20480 | 12.6 | 45 | 36.3470938035692
32 | 0.7 | 3.3 | 2.9260409155032
1024 | 3.1 | 13.7 | 20.8520818310064
32768 | 7.3 | 33 | 38.7781227465096


</body>

</html>

maccoymerrell:
Alternatively, here is a power fit that works pretty well:
latency (cycles) = x(KiB) ^ 0.35

Size | Latency (ns) | Latency (cycles) | ChampSim (log) | ChampSim (pow)
-- | -- | -- | -- | --
48 | 1.6 | 3.5 | 5.02325325989571 | 3.87645219971935
1280 | 4.2 | 12.8 | 22.0062610711666 | 12.2327327957256
49152 | 20.1 | 38.7 | 40.8753350909021 | 43.857050197873
48 | 1.4 | 5 | 5.02325325989571 | 3.87645219971935
1280 | 2.8 | 10 | 22.0062610711666 | 12.2327327957256
20480 | 12.6 | 45 | 36.3470938035692 | 32.2823753886618
32 | 0.7 | 3.3 | 2.9260409155032 | 3.36358566101486
1024 | 3.1 | 13.7 | 20.8520818310064 | 11.3137084989848
32768 | 7.3 | 33 | 38.7781227465096 | 38.0546276800871

ngober:
I think if we need to choose between an L2 latency too long or an L3 latency too short, we should prefer modeling the L2 latency too long. If someone cares about this, it's better for them to adjust in the faster direction, versus the person who doesn't care getting speed for free.

The power fit is interesting. I really like the fit.

maccoymerrell:
Yeah, R^2 on the power fit is 0.96, so I think that is probably the way to go.
I will revert my changes and do that instead.
I will also add to the cache builder test to ensure that projected latency values are within certain limits.
 Maybe the max and min latencies we have observed here for each cache size.

ngober:
I think we should lower bound the latency at 1 cycle, but not upper bound it. If someone wants a 2 TiB cache, they should feel the weight of that.

maccoymerrell:
From what I have seen from the code, total latency has to be bounded at a minimum of 2 cycles. 1 for fill and 1 for hit. Latencies of 0 in hit or fill seem to break ChampSim.

maccoymerrell:
> I think we should lower bound the latency at 1 cycle, but not upper bound it. If someone wants a 2 TiB cache, they should feel the weight of that.

That would be 1709 cycles with the new model

maccoymerrell:
> > I think we should lower bound the latency at 1 cycle, but not upper bound it. If someone wants a 2 TiB cache, they should feel the weight of that.
> 
> That would be 1709 cy

ngober (approved):

Issue #527, opened 2024-10-23
Register producers removed prematurely


joshualmashburn:
**Describe the bug**
In the reg_producers array of vectors, when an instruction that produces a register value reaches do_complete_execution(), that instruction is found in the vector that corresponds to the destination register and removed.
However, if a slower, older instruction hasn't completed yet, it will still reside in the vector and become the "newest" producer for that register. So in that cycle, when an instruction schedules and reads from that same register, instead of being dependent on the completed producer, it is marked dependent on the old one, delaying execution by at least one cycle.

**To Reproduce**
1. Use trace 400.perlbench-41B.champsimtrace.xz
2. at cycle 13066, output the ROB and all register sources and destinations for instruction IDs 2107-2112. Output the producers vector for register 26.
3. 2107 and 2109 write to register 26, but 2109 finishes first. Upon complete, 2109 is erased from the producers vector.
4. 2112 waits to execute because the scheduler made it dependent on 2107 erroneously, delaying it (and 2116 for similar reasons) a cycle, so this cycle, only 2111 executes despite having free execute bandwidth.

**Expected behavior**
Producing instructions for a given register should not be removed from the vector upon complete. Commits to the register file state should happen when the instructions retire, in-order.

**Output**
If applicable, add the output from ChampSim to help explain your problem.
```
bin/champsim --warmup-instructions 0 --simulation-instructions 2120 ~/ChampSim/400.perlbench-41B.champsimtrace.xz           3b9e7c7 
WARNING: physical memory size is smaller than virtual memory size.

*** ChampSim Multicore Out-of-Order Simulator ***
Warmup Instructions: 0
Simulation Instructions: 2120
Number of CPUs: 1
Page size: 4096

Off-chip DRAM Size: 16 GiB Channels: 1 Width: 64-bit Data Rate: 3205 MT/s
Warmup finished CPU 0 instructions: 0 cycles: 1 cumulative IPC: 0 (Simulation time: 00 hr 00 min 00 sec)
Warmup complete CPU 0 instructions: 0 cycles: 1 cumulative IPC: 0 (Simulation time: 00 hr 00 min 00 sec)
Producer of arch. reg: 26 is instr: 2107
Producer of arch. reg: 56 is instr: 2108
Producer of arch. reg: 26 is instr: 2109
Producer of arch. reg: 56 is instr: 2110
Producer of arch. reg: 8 is instr: 2111

Checking sources
2112 depends on 2107 for arch reg 26
[...]

PR #528, opened 2024-10-23
Fix for issue #527. Reg producers erase moved to retire

joshualmashburn:
I moved the producers_at(reg) search and erase from do_complete_execution to retire_rob.

This necessitated another check in the scheduler.
ie. if the instruction is still in the list, but
    has completed, don't depend on it since it's ready.

### DIFF
--- a/test/cpp/src/200-rob-scheduling.cc
+++ b/test/cpp/src/200-rob-scheduling.cc
@@ -63,10 +63,10 @@ SCENARIO("The scheduler can detect RAW hazards") {
         op->_operate();
 
       THEN("The second instruction is dependent on the first") {
-        REQUIRE(uut.ROB[0].num_reg_dependent == 0);
-        REQUIRE(uut.ROB[1].num_reg_dependent == 1);
-        REQUIRE(uut.ROB[0].scheduled);
-        REQUIRE(uut.ROB[1].scheduled);
+        REQUIRE(uut.ROB.at(0).num_reg_dependent == 0);
+        REQUIRE(uut.ROB.at(1).num_reg_dependent == 1);
+        REQUIRE(uut.ROB.at(0).scheduled);
+        REQUIRE(uut.ROB.at(1).scheduled);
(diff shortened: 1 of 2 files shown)


ngober:
Oh, I see what's going on here. This is a good find.

How hard is this to write a test for? We should test that when a core receives instructions A, B, and C, where A and B write a register and C reads it, and when B completes before A (say, A has a long-latency memory), C executes immediately after B.

joshualmashburn:
Done. Let me tell you, it was difficult to keep an instruction from completing within a cycle (in the second scenario I just added). Something in the core overrides my manual changes to ready_time. So I made it a load and it worked long enough for the test.

test/cpp/src/200-rob-scheduling.cc
```diff
@@ -130,3 +130,144 @@ SCENARIO("The scheduler can detect RAW hazards") {
...
+      REQUIRE(uut.ROB[2].num_reg_dependent == 1);
+      // Execute
+      for (auto op : std::array<champsim::operable*,3>{{&uut, &mock_L1I, &mock_L1D}})
+        op->_operate();
+
+      THEN("The third instruction does not execute") {
+        REQUIRE(uut.ROB[0].executed == true);
+        REQUIRE(uut.ROB[1].executed == true);
+        REQUIRE(uut.ROB[2].executed == false);
+        REQUIRE(uut.ROB[2].num_reg_dependent == 1);
```
ngober, line 179: I am somewhat concerned with examining the internals of `O3_CPU` and `ooo_model_instr` in this way. This test works and accomplishes what we want to do, but this test will break on the day when we modify the internals of either one of those classes.

joshualmashburn: It's difficult, since to ensure for instructions A,B,C that B depends on C and that this dependency is respected in the scheduler, without the internals, we can, at best, confirm that 3 instructions retire in X cycles rather than X+1 cycles, but not draw any conclusions about how the instructions were scheduled.

PR #532, opened 2024-10-30
Dual RAT register rename scheme

joshualmashburn:
Created a new RegisterAllocator class that encapsulates the register rename mechanism.
A front-end register allocation table tracks the speculative register state, while a back-end RAT tracks the architectural state at instruction commit. The physical register file is unified and for performance, the list of free entries is tracked in a queue rather than a sequential search. (The scheme is described in chapter 5 of https://link.springer.com/book/10.1007/978-3-031-01729-2 )

Note: The physical register file is unified and size is currently a constant modified in ooo_cpu.h. I'm opening a PR now so review of the implementation can begin, but before merge, I will add a parameter to the json config file and parser to modify the size of the register file at configuration time.


inc/instruction.h
```diff
@@ -117,8 +117,8 @@ struct ooo_model_instr : champsim::program_ordered<ooo_model_instr> {
   unsigned completed_mem_ops = 0;
   int num_reg_dependent = 0;
 
-  std::vector<uint8_t> destination_registers = {}; // output registers
-  std::vector<uint8_t> source_registers = {};      // input registers
+  std::vector<int16_t> destination_registers = {}; // output registers
+  std::vector<int16_t> source_registers = {};      // input registers
```
ngober, line 121: These feel like they should have type `std::vector<PHYSICAL_REGISTER_ID>`. Consider the logical hierarchy of the way that these headers are included.

ngober:
Noting that this should be merged _after_ #528.

src/register_allocator.cc
```diff
@@ -0,0 +1,107 @@
...
+    physical_register_file[phys] = {(uint16_t)reg, 0, true, true}; //arch_reg_index, producing_inst_id, valid, busy
+  }
+
+  return phys;
+}
+
+void RegisterAllocator::complete_dest_register(PHYSICAL_REGISTER_ID physreg)
+{
+  // mark the physical register as valid
+  physical_register_file[physreg].valid = true;
```
ngober, line 46:
```suggestion
  physical_register_file.at(physreg).valid = true;
```

Given that the index type is signed, using the checked accessor here and throughout is probably wise.

src/register_allocator.cc
```diff
@@ -0,0 +1,107 @@
...
+unsigned long RegisterAllocator::count_free_registers() { return std::size(free_registers); }
+
+int RegisterAllocator::count_reg_dependencies(ooo_model_instr& instr) {
+  // mainly for unit tests. This is not used in the core model
+  int depcount = 0;
+  for (auto reg : instr.source_registers){
+    if (!isValid(reg)) depcount++;
+  }
+  return depcount;
+}
```
ngober, line 81: Should this just be a free function in the test file?

ngober (changes requested):
This is a pretty good patch. I think the tests are good, and pretty comprehensive. I have some style notes, but nothing much.

src/register_allocator.cc
```diff
@@ -0,0 +1,107 @@
...
+{
+  for (uint16_t i = 0; i < num_physical_registers; ++i) {
+    free_registers.push(static_cast<PHYSICAL_REGISTER_ID>(i));
+  }
+  physical_register_file = std::vector<physical_register>(num_physical_registers, {0, 0, false, false});
+  frontend_RAT.fill(-1); //default value for no mapping
+  backend_RAT.fill(-1);
+}
+
+PHYSICAL_REGISTER_ID RegisterAllocator::rename_dest_register(int16_t reg, ooo_model_instr &instr)
```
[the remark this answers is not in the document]
joshualmashburn: I agree. It also makes it more obvious that there aren't any other operations on the instruction itself.

src/register_allocator.cc
joshualmashburn: I had just finished moving it and updating all the calls when I realized I write this loop verbatim in the deadlock print, so it will be called there to avoid code repetition.

ngober:
```suggestion
int RegisterAllocator::count_reg_dependencies(ooo_model_instr& instr) {
  return std::count_if(std::begin(instr.source_registers), std::end(instr.source_registers), [](auto reg){ return !isValid(reg); });
}
```

Alright then. How about this instead?

joshualmashburn: Looks good. Modern even.

joshualmashburn:
I have a note now. I added the register file size to the configuration parser, defaults, example json file, etc.
To avoid many warnings about a narrowing cast to a 16 bit number, I made the register file use a size_t for instantiation. I still want the upper limit to be 8192 registers (signed 16-bit max, and way more than anyone would ever hope to ever use).
I also want to set the lower bound to 96, to prevent anyone from setting it so low that the core deadlocks quickly.

Where should these checks go? In the config script, the parser, or the constructor of the register file (at runtime)?

ngober:
I think these checks can go in the constructor. You can assert `num_registers <= std::numeric_limits<PHYSICAL_REGISTER_ID>::max()` and that should do it. The lower bound should be a warning, I think. Because we don't have any hard limits (like the size of the type, for the max), we shouldn't firmly reject low numbers. If the user goes there, we should tell them a reason why their simulation _may have crashed_, but let it crash all the same.

joshualmashburn:
I think this is good to go now.

ngober:
There are still some merge conflicts, could you resolve them?

test/cpp/src/201-register-rename.cc
```diff
@@ -0,0 +1,373 @@
...
+                  REQUIRE(ra.count_free_registers() == PHYSICALREGS-1);
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+
+    WHEN("Ten writes to the same logical register occur"){
```
ngober, line 229: I can't help but feel like there's a better way to write this. Even a loop would help comprehensibility here.

ngober (approved):
I think it's pretty much good to go. I have some comments about style and clarity that could be helpful. If you think they're not worth it let me know and we'll merge.

src/ooo_cpu.cc
```diff
@@ -470,9 +467,19 @@ long O3_CPU::execute_instruction()
...
-      exec_bw.consume();
+    if (rob_it->scheduled && !rob_it->executed && rob_it->ready_time <= current_time) {
+      bool ready = true;
+      //ensure all source register operands are ready
+      for (auto srcreg : rob_it->source_registers) {
+        if (!reg_allocator.isValid(srcreg)) {
+          ready = false;
+          break;
+        }
+      }
```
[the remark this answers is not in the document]
joshualmashburn: Good to know! I really need to learn these.

PR #536, opened 2024-11-02
Added downstream stats for cache, fixed cache miss latency calculations to use downstream.

### LINKED ISSUE #535: Cache miss latency misreporting
**Describe the bug**
When reporting the miss latency of the cache, it is done in the following way:
1. Record when a MSHR type is created, this becomes the "time_enqueued" value
2. When merging, "time_enqueued" is set to the predecessor when the successor is a prefetch and the successor otherwise
3. Total miss latency cycles is the sum of the differences between cycle when the MSHR is closed and the time enqueued.
4. The average miss latency is calculated by dividing this value by ALL cache misses

**Expected behavior**
Average miss latency should report the average lifespan of an MSHR entry.
By allowing time_enqueued to be altered when multiple demands merge into a MSHR, it reduces the apparent cache latency.
By dividing by the total misses to the cache, we reduce the apparent miss latency by counting MSHR merges towards our total count.

**What are we trying to track?**
It seems to me that its better to track miss latency by recording the total cycles MSHR entries stay open and dividing it by the number of miss (non-writeback) packets sent downstream.
It also seems more correct to only update the time_enqueued value from the initial MSHR entry if a demand merges into a prefetch. Otherwise, to keep it unchanged. This _should_ make it so the reduction in cache latency from prefetching is apparent in the reporting, if there is a lot of demand->prefetch merging in the MSHR.

I have noticed that prefetches often repeatedly issue into open MSHRs, resulting in much higher "misses" in a cache than actual packets sent downstream. This results in our cache latency reporting at something ridiculously low (like 30 or 40 cycles) due to the over-report of misses. Without prefetching, the latency returns to reasonable values like 140-170 cycles. My suggested changes to how its calculated  resolve this.

maccoymerrell:
Three things this pr covers:
1. Added downstream stats to the cache. This shows how many of each type of packet went downstream from said cache. This should always be <= misses, but does not count misses that hit in the mshr. It should match up with accesses to the downstream cache/dram.
2. Cache miss latency is now calculated as the total time each MSHR was open divided by the number of downstream packets (excluding writes) from that cache. If a demand merges into a prefetch, then right now it updates the open time to the demand. I am not sure if this should be done or not. I can imagine that many prefetches never get a demand merge, so this change doesn't seem too meaningful, unless we only count demands when calculating the miss latency.
3. instantiation_file.py had a bug where clock_period would not be set for operables. This was due to clock period not being a defined field within the cache, cpu, and ptw by the user. I made some modifications that should fix that issue.
resolves #535

### DIFF
--- a/config/instantiation_file.py
+++ b/config/instantiation_file.py
@@ -55,7 +55,7 @@ core_builder_parts = {
     '_branch_predictor_data': '.branch_predictor<{^branch_predictor_string}>()',
     '_btb_data': '.btb<{^btb_string}>()',
     '_index': '.index({_index})',
-    '^clock_period': '.clock_period(champsim::chrono::picoseconds{{{^clock_period}}})'
+    'frequency': '.clock_period(champsim::chrono::picoseconds{{{^clock_period}}})'
 }
 
 dib_builder_parts = {
@@ -84,7 +84,7 @@ cache_builder_parts = {
     '_prefetcher_data': '.prefetcher<{^prefetcher_string}>()',
     'lower_translate': '.lower_translate(&{^lower_translate_queues})',
     'lower_level': '.lower_level(&{^lower_level_queues})',
-    '^clock_period': '.clock_period(champsim::chrono::picoseconds{{{^clock_period}}})'
+    'frequency': '.clock_period(champsim::chrono::picoseconds{{{^clock_period}}})'
 }
 
 ptw_builder_parts = {
@@ -94,7 +94,7 @@ ptw_builder_parts = {
     'mshr_size': '.mshr_size({mshr_size})',
     'max_read': '.tag_bandwidth(champsim::bandwidth::maximum_type{{{max_read}}})',
     'max_write': '.fill_bandwidth(champsim::bandwidth::maximum_type{{{max_write}}})',
-    '^clock_period': '.clock_period(champsim::chrono::picoseconds{{{^clock_period}}})'
+    'frequency': '.clock_period(champsim::chrono::picoseconds{{{^clock_period}}})'
 }
 
 def vector_string(iterable):
@@ -128,7 +128,7 @@ def get_cpu_builder(cpu, caches, ul_pairs):
     builder_parts = itertools.chain(util.multiline(itertools.chain(
         ('champsim::core_builder{{ champsim::defaults::default_core }}',),
         required_parts,
-        *(util.wrap_list(v) for k,v in core_builder_parts.items() if k in cpu or k in local_params),
+        *(util.wrap_list(v) for k,v in core_builder_parts.items() if k in cpu),
         (v for k,v in dib_builder_parts.items() if k in cpu.get('DIB',{}))
     ), indent=1, line_end=''))
     yield from (part.format(**cpu, **local_params) for part in builder_parts)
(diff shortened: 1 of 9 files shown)


ngober:
Could you split this into multiple pull requests? I don't want comments on one to block the others. You should be able to 

    git switch develop
    git switch -c downstream_stats
    git cherry-pick -n ..cache_miss_latency_fix
    git add -p

maccoymerrell:
I don't know exactly what I would be splitting. The downstream stats are just printing out data that is already being collected for the latency fix. I can't exactly treat them as separate, since they both depend on each other.

I might be able to move the frequency fix for the config out.

test/cpp/src/498-cache-plain-printer.cc
```diff
@@ -8,12 +8,12 @@ TEST_CASE("An empty cache stat block prints all zeros") {
...
-    "test_cache RFO          ACCESS:          0 HIT:          0 MISS:          0",
-    "test_cache PREFETCH     ACCESS:          0 HIT:          0 MISS:          0",
-    "test_cache WRITE        ACCESS:          0 HIT:          0 MISS:          0",
-    "test_cache TRANSLATION  ACCESS:          0 HIT:          0 MISS:          0",
+    "test_cache TOTAL        ACCESS:          0 HIT:          0 MISS:          0 DOWNSTREAM:          0",
+    "test_cache LOAD         ACCESS:          0 HIT:          0 MISS:          0 DOWNSTREAM:          0",
+    "test_cache RFO          ACCESS:          0 HIT:          0 MISS:          0 DOWNSTREAM:          0",
+    "test_cache PREFETCH     ACCESS:          0 HIT:          0 MISS:          0 DOWNSTREAM:          0",
+    "test_cache WRITE        ACCESS:          0 HIT:          0 MISS:          0 DOWNSTREAM:          0",
+    "test_cache TRANSLATION  ACCESS:          0 HIT:          0 MISS:          0 DOWNSTREAM:          0",
```
ngober, line 16: I have mixed feelings about this reporting. On the one hand, I have wanted for a long time to add in some kind of "MSHR merged" indicator. On the other hand, I think that "downstream" is either not the right name or not the right statistic to be showing to the user.

This information, by your admission, is already shown in the lower-level accesses, and we shouldn't show redundant information. We get so many questions already about how to interpret these statistics that I want to make sure any changes are in the direction of increased clarity.

maccoymerrell: I think I get more questions about why misses don't equal accesses in the downstream cache more than anything else. We could report the # of merges instead, which would be misses - downstream.

ngober: I would rather do that, because that says positively what happened to each packet. However, if we simply call it `MERGED`, then the question will be "merged with what?", so we need a name like `MSHR_MERGE` or something.

maccoymerrell:
Also, are you of the opinion that we should only be counting the latency of demand packets, and ignore the prefetch latency?
I can make a small change that would do that.

ngober:
We definitely shouldn't count the latency for prefetches from the current level, unless it gets promoted. I think we also should not count prefetches coming from the upper level. To include them could inflate the latency if the prefetcher is bad. On the other hand, maybe because it's being averaged it won't matter.

config/instantiation_file.py
```diff
@@ -122,6 +122,7 @@ def cache_index(name):
         '^l1i_ptr': f'(*std::next(std::begin(caches), {cache_index(cpu.get("L1I"))}))',
         '^l1d_ptr': f'(*std::next(std::begin(caches), {cache_index(cpu.get("L1D"))}))'
     }
+
```
ngober, line 125:
```suggestion
```

ngober (approved):
Removing some blank lines that were the only changes in the file.

PR #537, opened 2024-11-02
Fix config scripts bug that causes frequency to always be 4GHz regardless of config

maccoymerrell:


ngober:
I think the bug here is actually on line 58. I think the key needs to be `'frequency'`. Is that a simpler solution?

maccoymerrell:
Pytest seems to sometimes call these builders when frequency is not defined, causing pytest to fail. I could remove the added checks for local_params, but I need to keep the lines that check to see if frequency is given.

ngober:
Oh, I see. `'^clock_period`' requires `'frequency'` in `local_params`, and in the current version `'frequency'` requires `'^clock_period'` in `core_builder_parts`. Yea, I don't like the if-statements, but this is a smaller change than you had initially...

config/instantiation_file.py
```diff
@@ -151,14 +153,15 @@ def get_cache_builder(elem, ul_pairs):
...
     local_params = {
-        '^clock_period': int(1000000/elem['frequency']),
         '^defaults': elem.get('_defaults', ''),
         '^upper_levels_string': vector_string(f'&channels.at({ul_pairs.index(v)})' for v in uppers),
         '^prefetch_activate_string': ', '.join('access_type::'+t for t in elem.get('prefetch_activate',[])),
         '^replacement_string': ', '.join(f'class {k["class"]}' for k in elem.get('_replacement_data',[])),
         '^prefetcher_string': ', '.join(f'class {k["class"]}' for k in elem.get('_prefetcher_data',[])),
         '^lower_level_queues': f'channels.at({ul_pairs.index((elem.get("lower_level"), elem.get("name")))})'
     }
+    if('frequency' in elem):
```
ngober, line 163:
```suggestion
    if 'frequency' in elem:
```

config/instantiation_file.py
```diff
@@ -191,10 +194,11 @@ def get_ptw_builder(ptw, ul_pairs):
 
     uppers = (v for v in ul_pairs if v[0] == ptw.get('name'))
     local_params = {
-        '^clock_period': int(1000000/ptw['frequency']),
         '^upper_levels_string': vector_string(f'&channels.at({ul_pairs.index(v)})' for v in uppers),
         '^lower_level_queues': f'channels.at({ul_pairs.index((ptw.get("lower_level"), ptw.get("name")))})'
     }
+    if('frequency' in ptw):
```
ngober, line 200:
```suggestion
    if 'frequency' in ptw:
```

config/instantiation_file.py
```diff
@@ -122,6 +122,8 @@ def cache_index(name):
         '^l1i_ptr': f'(*std::next(std::begin(caches), {cache_index(cpu.get("L1I"))}))',
         '^l1d_ptr': f'(*std::next(std::begin(caches), {cache_index(cpu.get("L1D"))}))'
     }
+    if('frequency' in cpu):
```
ngober, line 125:
```suggestion
    if 'frequency' in cpu:
```

ngober (approved):

PR #538, opened 2024-11-08
Added vmem randomization in config

maccoymerrell:
I have had this change on my research forks for quite a while. I was doing it via command-line argument, but the way ChampSim is set up, this always guaranteed that the main page structure for the virtual memory would be linear-mapped since they are assigned immediately, before command-line arguments can be parsed.

This instead uses a config option to control it, as a list of two parameters [randomization_enabled,randomization_seed].
Since this is passed to the VMEM constructor, we can avoid the page tables getting linear-mapped when we don't want it.
[false,n] == disabled randomization
[true,n] == enabled randomization at seed n.


ngober (changes requested):

maccoymerrell:
Here's what's changed:
1. There is a single vmem constructor with an std::optional randomization_seed argument at the end. It has a default value of {} if not specified.
2. There is a single value passed in the config for randomization. If it is false, '{}' is used within the vmem constructor for randomization seed, else (true or an integer) it is passed as an integer value.
3. Tests have been altered to use the constructor without defining the randomization seed. DRAM still had to be changed for these tests, as vmem cannot support memories smaller than 1 MiB. I don't know what was allowing these tests to pass before, but it shouldn't have been possible since no physical memory should have been available.
4. An assertion has been added to enforce that DRAM > 1 MiB.

test/cpp/src/092-ptw-builder.cc
```diff
@@ -9,7 +9,7 @@
 #include "channel.h"
 
 TEST_CASE("The MSHR factor uses the number of upper levels to determine the PTW's default number of MSHRs") {
-  MEMORY_CONTROLLER dram{champsim::chrono::picoseconds{3200}, champsim::chrono::picoseconds{6400}, std::size_t{18}, std::size_t{18}, std::size_t{18}, std::size_t{38}, champsim::chrono::microseconds{64000}, {}, 64, 64, 1, champsim::data::bytes{8}, 1, 128, 1, 1, 1, 8192};
+  MEMORY_CONTROLLER dram{champsim::chrono::picoseconds{3200}, champsim::chrono::picoseconds{6400}, std::size_t{18}, std::size_t{18}, std::size_t{18}, std::size_t{38}, champsim::chrono::microseconds{64000}, {}, 64, 64, 1, champsim::data::bytes{8}, 1024, 1024, 4, 4, 4, 8192};
```
ngober, line 12: What is the purpose of these edits? They don't seem to have anything to do with the patch.

ngober (approved):
I generally disagree with the choice to use a default constructor argument, see [here](https://quuxplusone.github.io/blog/2020/04/18/default-function-arguments-are-the-devil/), particularly the points about overload sets. In my opinion, C++ has function overloads for exactly this purpose, and it's better for code health. But, I don't feel it's necessary to labor on that point. If you're satisfied, I'm satisfied.

maccoymerrell:
My reasoning behind this was that with the overloaded constructor approach we would have two constructors with identical code to maintain except for the initialization of the randomization_seed member in one over the other.
Unless there is a way to reuse the constructor declaration for definitions?

I have altered the code to use two different constructors instead, and tried to minimize the code inside each constructor.

ngober:
You mean something like a [delegating constructor](https://en.cppreference.com/w/cpp/language/constructor#Delegating_constructor)?

maccoymerrell:
Can't use that, as we need to initialize the member variable (randomization_seed)  _before_ calling the constructor for it to work, or shuffle_pages() isn't going to shuffle because randomization_seed is not set yet. GCC was spamming errors about that.

maccoymerrell:
Hold up I can do it backwards though.

ngober:
I'm rerunning that check. Looks like something went wrong with `curl | xz`.

src/vmem.cc
```diff
@@ -26,25 +26,44 @@
...
+void VirtualMemory::populate_pages() {
+  assert(dram.size() > 1_MiB);
+  ppage_free_list.resize(((dram.size() - 1_MiB)/PAGE_SIZE).count());\
+  assert(ppage_free_list.size() != 0);
+  champsim::page_number base_address = champsim::page_number{champsim::lowest_address_for_size(std::max<champsim::data::mebibytes>(champsim::data::bytes{PAGE_SIZE}, 1_MiB))};
+  for(auto it = ppage_free_list.begin(); it != ppage_free_list.end(); it++) {
+    *it = base_address;
+    base_address++;
+  }
+  fmt::print("[VMEM] Freed {} ppages\n",ppage_free_list.size());
```
ngober, line 60: This prints every time. Is this intentional? It feels like a debug statement. Also L66.

PR #544, opened 2024-11-13
Changed plain-printer to indicate origin cpu of cache stats.

maccoymerrell:
There have been questions about what the cache stats from the plain printer are indicating for multi-core runs.
This change indicates the origin core that the cache stats are being split by. The json files already do this, but the plain printer does not, so this brings it into sync.

### DIFF
--- a/src/plain_printer.cc
+++ b/src/plain_printer.cc
@@ -76,26 +76,19 @@ std::vector<std::string> champsim::plain_printer::format(CACHE::stats_type stats
 
   //build a vector of all existing cpus
   auto stat_keys = {stats.hits.get_keys(), stats.misses.get_keys(), stats.mshr_merge.get_keys(), stats.mshr_return.get_keys()};
-  for (auto keys : stat_keys)
-    for (auto key : keys)
-      if(std::find_if(cpus.begin(), cpus.end(), [key=key](const auto val) {return key.second == val;}) == cpus.end())
-        cpus.push_back(key.second);
-        
-  //ensure stats were collected for active cpu, print warning otherwise
-  //always display stats for active cpu
-  if(std::find_if(cpus.begin(), cpus.end(), [key=stats.cpu](const auto val) {return key == val;}) == cpus.end())
-  {
-    cpus.push_back(stats.cpu);
-    fmt::print("[PLAIN PRINTER] WARNING: cpu{} stats were not collected for {}\n",stats.cpu,stats.name);
+  for (auto keys : stat_keys) {
+    std::transform(std::begin(keys), std::end(keys), std::back_inserter(cpus), [](auto val){ return val.second; });
   }
-  
-  std::sort(cpus.begin(),cpus.end());
+  std::sort(std::begin(cpus), std::end(cpus));
+  auto uniq_end = std::unique(std::begin(cpus), std::end(cpus));
+  cpus.erase(uniq_end, std::end(cpus));
 
   for (const auto type : {access_type::LOAD, access_type::RFO, access_type::PREFETCH, access_type::WRITE, access_type::TRANSLATION}) {
     for (auto cpu : cpus) {
-      stats.hits.set_default(std::pair{type, cpu});
-      stats.misses.set_default(std::pair{type, cpu});
-      stats.mshr_merge.set_default(std::pair{type, cpu});
+      stats.hits.allocate(std::pair{type, cpu});
+      stats.misses.allocate(std::pair{type, cpu});
+      stats.mshr_merge.allocate(std::pair{type, cpu});
+      stats.mshr_return.allocate(std::pair{type, cpu});
     }
   }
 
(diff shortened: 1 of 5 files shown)


ngober:
Interesting... you added the cpu index to the plain printer, but didn't add a test for the multicore case?

maccoymerrell:
How can I test the multicore case when NUM_CPUS is a const?

test/cpp/src/498-cache-plain-printer.cc
```diff
@@ -24,26 +24,26 @@ TEST_CASE("An empty cache stat block prints all zeros") {
...
+      std::tuple{1, access_type::LOAD, "cpu0->test_cache LOAD         ACCESS:        255 HIT:        255 MISS:          0 MSHR_MERGE:          0"},
+      std::tuple{2, access_type::RFO, "cpu0->test_cache RFO          ACCESS:        255 HIT:        255 MISS:          0 MSHR_MERGE:          0"},
+      std::tuple{3, access_type::PREFETCH, "cpu0->test_cache PREFETCH     ACCESS:        255 HIT:        255 MISS:          0 MSHR_MERGE:          0"},
+      std::tuple{4, access_type::WRITE, "cpu0->test_cache WRITE        ACCESS:        255 HIT:        255 MISS:          0 MSHR_MERGE:          0"},
+      std::tuple{5, access_type::TRANSLATION, "cpu0->test_cache TRANSLATION  ACCESS:        255 HIT:        255 MISS:          0 MSHR_MERGE:          0"}
   );
 
   cache_stats given{};
   given.name = "test_cache";
   given.hits.set({hit_type, 0}, num_hits);
```
ngober, line 36:
```suggestion
  given.hits.set({hit_type, 0}, num_hits);
  given.hits.set({hit_type, 1}, num_hits*3);
```

maccoymerrell: I have tried implementing them, but because plain printer only checks for cpus less than NUM_CPUS, so plain printer doesn't print those stats out for comparison.

maccoymerrell:
This is the only way I could think of solving this problem.
1. Added a function to the event_counter that returns a vector of its keys
2. Plain printer searches the vectors of keys from the event counters for the highest cpu number.
3. If this value exceeds NUM_CPUS, it becomes the new cpu count.
4. Plain printer backfills the stats with zeros for all missing stats (already did this)

ngober:
I don't actually dislike this solution as much as you do. I think it's probably good to always print cores with `i < NUM_CPUS`, but printing anything else that shows up could be an early indicator of bugs.

ngober:
I'm trying to think of a good way to print all of `i < NUM_CPUS` without printing all of `i < max(cpu_seen)`. If, for example, some bug happens and we see `i = std::numeric_limits<uint32_t>::max()`, that's going to blow out a terminal and be a bad day.

What if we gave `event_counter` a member `set_default()`, something like [a python dict has](https://docs.python.org/3/library/stdtypes.html#dict.setdefault). Then, right before we print, we can stripe through the list with

    for (auto type : types) {
        for (uint32_t i = 0; i < NUM_CPUS; ++i) {
             stats.set_default({type, i}, 0);
        }
    }

Then we just print everything.

Now, I would really love to avoid printing stats for cores that are inaccessible to a certain cache. Does it make much sense to print cpu6 stats for cpu1's instruction cache? Maybe it makes sense in the presence of coherence invalidations?

maccoymerrell:
I also don't like differentiating a cache by its name, which is what we would end up doing to some extent if we start determining accessibility.

We could print out stats for every core that ever accessed the cache, and skip cores we never saw. That does mean that for edge cases we could see an LLC that only prints out stats for a few cores because warmup hid the accesses from the other cores. It protects us against blowing out the terminal though.

src/plain_printer.cc
```diff
@@ -72,41 +72,62 @@ std::vector<std::string> champsim::plain_printer::format(CACHE::stats_type stats
...
   using mshr_return_value_type = typename decltype(stats.mshr_return)::value_type;
 
+  std::vector<std::size_t> cpus;
+
+  //build a vector of all existing cpus
+  auto stat_keys = {stats.hits.get_keys(), stats.misses.get_keys(), stats.mshr_merge.get_keys(), stats.mshr_return.get_keys()};
+  for (auto keys : stat_keys)
+    for (auto key : keys)
+      if(std::find_if(cpus.begin(), cpus.end(), [key=key](const auto val) {return key.second == val;}) == cpus.end())
+        cpus.push_back(key.second);
```
ngober, line 82:
```suggestion
  for (auto keys : stat_keys) {
    std::transform(std::begin(keys), std::end(keys), std::back_inserter(cpus), [](auto val){ return val.second; });
  }
  std::sort(std::begin(cpus), std::end(cpus));
  auto uniq_end = std::unique(std::begin(cpus), std::end(cpus));
  cpus.erase(uniq_end, std::end(cpus));
```
This avoids searching the vector dozens of times.

src/plain_printer.cc
```diff
@@ -72,41 +72,62 @@ std::vector<std::string> champsim::plain_printer::format(CACHE::stats_type stats
...
+        
+  //ensure stats were collected for active cpu, print warning otherwise
```
ngober, line 84: You note here that it ensures stats were printed for the "active" cpu, but this is not quite the case. It is the most recently used cpu whenever the phase began. It seems like it's a bit of a hack that works for caches that are connected only to one core, but for caches downstream of more than one core It doesn't actually accomplish what you want.

ngober (approved):

PR #545, opened 2024-11-13
Added modern dram mappings to memory controller.

maccoymerrell:
Modernized dram mappings have been added to the memory controller.
Because our model allows for arbitrary values for # of rows,banks, etc..., we cannot directly copy mappings from real machines.

This mapping is heavily based on the re'd zen4 mappings.

Each channel bit is xor'd with all row bits.

Bankgroup and bank bits are xor'd all at once, as follows.
[bank bits][bank group bits]
  ^ ^ ^ ^ ^ ^ ^ ^ ^ ^ ^ ^ ^ ^ ^
[row bits [bg bits + bk_bits:0]]
  ^ ^ ^ ^ ^ ^ ^ ^ ^ ^ ^ ^ ^ ^ ^
[row bits [(bg_bits + bk_bits)*(i+1)][(bg_bits + bk_bits)*i]]
 ....
 ....
 Until we are out of row bits to xor with.

New tests were added to test this, and ensure that no aliasing occurs in these mappings.

### DIFF
--- a/inc/dram_controller.h
+++ b/inc/dram_controller.h
@@ -60,6 +60,7 @@ struct DRAM_ADDRESS_MAPPING {
   unsigned long get_bank(champsim::address address) const;
   unsigned long get_row(champsim::address address) const;
   unsigned long get_column(champsim::address address) const;
+  unsigned long swizzel_bits(champsim::address address, unsigned long segment_size, unsigned long segment_offset, unsigned long field, unsigned long field_bits) const;
 
   bool is_collision(champsim::address a, champsim::address b) const;
 
--- a/inc/dram_controller.h
+++ b/inc/dram_controller.h
@@ -60,7 +60,23 @@ struct DRAM_ADDRESS_MAPPING {
   unsigned long get_bank(champsim::address address) const;
   unsigned long get_row(champsim::address address) const;
   unsigned long get_column(champsim::address address) const;
-  unsigned long swizzel_bits(champsim::address address, unsigned long segment_size, unsigned long segment_offset, unsigned long field, unsigned long field_bits) const;
+
+   /**
+   * Perform the hashing operations for indexing our channels, banks, and bankgroups.
+   * This is done to increase parallelism when serving requests at the DRAM level.
+   *
+   * :param address: The physical address at which the hashing operation is occurring.
+   * :param segment_size: The number of row bits extracted during each iteration. (# of row bits / segment_size) == # of XOR operations
+   * :param segment_offset: The bit offset within the segment that the XOR operation will occur at. The bits taken from the segment will be [segment_offset + field_bits : segment_offset]
+   * 
+   * :param field: The input index that is being permuted by the operation.
+   * :param field_bits: The length of the index in bits.
+   * 
+   * Each iteration of the operation takes the selected bits of the segment and XORs them with the entirety of the field
+   * which should be equal or greater than length (in the case of the last iteration). This continues until no bits remain
+   * within the row that have not been XOR'd with the field.
+   */
+  unsigned long swizzle_bits(champsim::address address, unsigned long segment_size, unsigned long segment_offset, unsigned long field, unsigned long field_bits) const;
 
   bool is_collision(champsim::address a, champsim::address b) const;
 
--- a/inc/dram_controller.h
+++ b/inc/dram_controller.h
@@ -76,7 +76,7 @@ struct DRAM_ADDRESS_MAPPING {
    * which should be equal or greater than length (in the case of the last iteration). This continues until no bits remain
    * within the row that have not been XOR'd with the field.
    */
-  unsigned long swizzle_bits(champsim::address address, unsigned long segment_size, unsigned long segment_offset, unsigned long field, unsigned long field_bits) const;
+  unsigned long swizzle_bits(champsim::address address, unsigned long segment_size, champsim::data::bits segment_offset, unsigned long field, unsigned long field_bits) const;
 
   bool is_collision(champsim::address a, champsim::address b) const;
 
--- a/src/dram_controller.cc
+++ b/src/dram_controller.cc
@@ -552,54 +552,43 @@ bool MEMORY_CONTROLLER::add_wq(const request_type& packet)
   return false;
 }
 
+unsigned long DRAM_ADDRESS_MAPPING::swizzel_bits(champsim::address address, unsigned long segment_size, unsigned long segment_offset, unsigned long field, unsigned long field_bits) const {
+  champsim::address row = champsim::address{get_row(address)};
+  unsigned long permute_field = field;
+
+  unsigned long r_bits = champsim::size(get<SLICER_ROW_IDX>(address_slicer));
+  for(unsigned long r = 0; r < r_bits; r += segment_size) {
+    permute_field ^= row.slice(champsim::dynamic_extent{champsim::data::bits{segment_offset + field_bits},champsim::data::bits{segment_offset}}).to<unsigned long>();
+    row = champsim::address{row.to<unsigned long>() >> segment_size};
+  }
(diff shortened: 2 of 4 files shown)


src/dram_controller.cc
```diff
@@ -554,19 +554,55 @@ bool MEMORY_CONTROLLER::add_wq(const request_type& packet)
...
+  champsim::address row = champsim::address{get_row(address)};
+
+  //channel bits should be xor'd with each row bit
+  unsigned long c_bits = champsim::size(get<SLICER_CHANNEL_IDX>(address_slicer));
+  unsigned long r_bits = champsim::size(get<SLICER_ROW_IDX>(address_slicer));
+  for(unsigned long r = 0; r < r_bits; r++) {
+    for(unsigned long c = 0; c < c_bits; c++)
+      channel ^= row.slice_lower<champsim::data::bits{0}>().to<unsigned long>() << c;
+    row = champsim::address{row.to<unsigned long>() >> 1};
+  }
```
ngober, line 567: Similar code shows up 3 times here. Is there a way to extract this into a `DRAM_ADDRESS_MAPPING::swizzle_bits(...)` function?

inc/dram_controller.h
```diff
@@ -60,6 +60,7 @@ struct DRAM_ADDRESS_MAPPING {
   unsigned long get_bank(champsim::address address) const;
   unsigned long get_row(champsim::address address) const;
   unsigned long get_column(champsim::address address) const;
+  unsigned long swizzel_bits(champsim::address address, unsigned long segment_size, unsigned long segment_offset, unsigned long field, unsigned long field_bits) const;
```
ngober, line 63:
```suggestion
  unsigned long swizzle_bits(champsim::address address, unsigned long segment_size, unsigned long segment_offset, unsigned long field, unsigned long field_bits) const;
```

src/dram_controller.cc
```diff
@@ -552,21 +552,46 @@ bool MEMORY_CONTROLLER::add_wq(const request_type& packet)
...
 
+unsigned long DRAM_ADDRESS_MAPPING::swizzle_bits(champsim::address address, unsigned long segment_size, unsigned long segment_offset, unsigned long field, unsigned long field_bits) const {
+  champsim::address row = champsim::address{get_row(address)};
+  unsigned long permute_field = field;
+
+  unsigned long r_bits = champsim::size(get<SLICER_ROW_IDX>(address_slicer));
+  for(unsigned long r = 0; r < r_bits; r += segment_size) {
+    permute_field ^= row.slice(champsim::dynamic_extent{champsim::data::bits{segment_offset + field_bits},champsim::data::bits{segment_offset}}).to<unsigned long>();
+    row = champsim::address{row.to<unsigned long>() >> segment_size};
+  }
```
ngober, line 563:
```suggestion
  for(champsim::dynamic_extent subextent{champsim::data::bits{segment_offset}, field_bits}; subextent.upper <= row.upper_extent(); subextent = champsim::dynamic_extent{subextent.upper, field_bits}) {
    permute_field ^= row.slice(subextent).to<unsigned long>();
  }
```
Again, you've taken a detour through the integral types when what you want to do can be achieved in the high-level interface. Also, I've here used an alternate constructor for `champsim::dynamic_extent` that takes a lower bit and a size.

src/dram_controller.cc
```diff
@@ -552,21 +552,46 @@ bool MEMORY_CONTROLLER::add_wq(const request_type& packet)
   return false;
 }
 
+unsigned long DRAM_ADDRESS_MAPPING::swizzle_bits(champsim::address address, unsigned long segment_size, unsigned long segment_offset, unsigned long field, unsigned long field_bits) const {
```
ngober, line 555:
```suggestion
unsigned long DRAM_ADDRESS_MAPPING::swizzle_bits(champsim::address address, unsigned long segment_size, champsim::data::bits segment_offset, unsigned long field, unsigned long field_bits) const {
```
This value is only ever used as this type, so you should make it part of the interface.

ngober (changes requested):
Your bit swizzling takes too many descents into the domain of raw integer types. I've added some suggestions to avoid that.

src/dram_controller.cc
```diff
@@ -552,21 +552,44 @@ bool MEMORY_CONTROLLER::add_wq(const request_type& packet)
   return false;
 }
 
+unsigned long DRAM_ADDRESS_MAPPING::swizzle_bits(champsim::address address, unsigned long segment_size, champsim::data::bits segment_offset, unsigned long field, unsigned long field_bits) const {
+  champsim::address_slice row{get<SLICER_ROW_IDX>(address_slicer), address};
+  unsigned long permute_field = field;
+
+  for(champsim::dynamic_extent subextent{champsim::data::bits{0}, segment_size}; subextent.upper <= row.upper_extent(); subextent = champsim::dynamic_extent{subextent.upper, segment_size}) {
+    permute_field ^= row.slice(subextent).slice(champsim::dynamic_extent{segment_offset,field_bits}).to<unsigned long>();
```
ngober, line 560: I think the double slice is somewhat different than my suggestion. Could you explain how this works? The `address_slice::slice()` family of functions are relative to the domain of the slice. They are not absolute. So, a repeated slice is a subset of a subset of bits. Is that what you intended?

I see you also added some test cases to 751. Are these related?

maccoymerrell: The segment being used from the row address and the bits being xor'd are not identical.

The function travels over the bits of the row address in segments of segment_size.
Bits are taken from [segment_offset + field_bits : segment_offset] of the segment to xor with the field.

The bits chosen for bankgroup and bank are a good example of why this is needed.
Say we have 2 bg bits and 2 bk bits. Zen4 takes the first two bits of the row and xor's it with the first 2 bg bits.
It then takes the next 2 bits from the row and xor's it with the bk bits. It then takes the next two bits and xor's it with the bg bits and so on, back and forth.
So swizzle_bits(address, bk_bits + bg_bits, 0, bankgroup, bg_bits) gives us the bankgroup index
while swizzle_bits(address, bk_bits + bg_bits, bg_bits, bank, bk_bits) gives us the bank index.

The testcases in 751 ensure that these mappings aren't broken. It tests to make sure that for a given row the channel, banks, and bankgroups never alias back into themselves and that each channel, bank, and bankgroup is visited exactly once as the row is traversed.

ngober: OK, I understand. The segment size is the stride, and the segment offset is the location within that stride. That matches with what's given here, then.

src/dram_controller.cc
```diff
@@ -552,21 +552,44 @@ bool MEMORY_CONTROLLER::add_wq(const request_type& packet)
   return false;
 }
 
+unsigned long DRAM_ADDRESS_MAPPING::swizzle_bits(champsim::address address, unsigned long segment_size, champsim::data::bits segment_offset, unsigned long field, unsigned long field_bits) const {
+  champsim::address_slice row{get<SLICER_ROW_IDX>(address_slicer), address};
+  unsigned long permute_field = field;
+
+  for(champsim::dynamic_extent subextent{champsim::data::bits{0}, segment_size}; subextent.upper <= row.upper_extent(); subextent = champsim::dynamic_extent{subextent.upper, segment_size}) {
```
ngober, line 559:
```suggestion
  using namespace champsim::data::data_literals;
  for(champsim::dynamic_extent subextent{0_b, segment_size}; subextent.upper <= row.upper_extent(); subextent = champsim::dynamic_extent{subextent.upper, segment_size}) {
```

By the way, I don't know if you know that this syntax exists. If you want it, here it is.

ngober (approved):

maccoymerrell:
I think we are good to go on this. I am happy with it.

maccoymerrell:
Looks like something happened with curl.

PR #546, opened 2024-11-15
Added component class, added parameter dump

### LINKED ISSUE #539: Parameter dumps for configuration testing
We have recently had two pull requests (#526, #537) where configuration options would be specified but would not make it into the C++ side of the build path. Currently, there is no way to integration-test the configuration all the way through. There are unit tests that test config file parsing and defaulting, there are tests for instantiation file generation, and there are tests for builders. However, these two bugs are both the result of one of these components not talking to another correctly. We need a way to test that a given configuration compiles to a given architecture.

I would like to see a command-line option for ChampSim that, rather than running any simulation, prints all of the properties of all components in a machine-readable format, then quits. For example:

    $ ./bin/champsim --dump
    CACHE cpu0_L1I.sets 512
    CACHE cpu0_L1I.ways 8
    (many more lines of output...)
    CPU cpu0.retire_width 6
    CPU cpu0.lq_size 72
    (and so on...)

Here, I've used `--dump` as the option, but the name could be anything else. The benefit here is that the values printed _are the values seen by the simulation runtime_, so we are testing that the intent of the user is respected.

Once we have this, we could run any configuration files, perhaps starting with the ones in [this](https://github.com/ChampSim/ChampSim/tree/develop/test/config/compile-only) directory, through the configure-make-run process and show that they do not break over time. I think this is especially important in the context of #424, because the inputs (configuration files) and outputs (machine-readable dump) do not change, only the process by which they are brought together.

I would love for someone to volunteer for this. This is likely an easy way to get started with contributions to ChampSim, since it requires no prior knowledge of how the insides of the system work.

maccoymerrell:
Passing the argument _--dump_ when running ChampSim will now dump the internal configuration of ChampSim's components and then quit.

The component class was added to help differentiate two classes of interfaces within ChampSim.
The component class exposes interfaces to the runtime, while the operable class exposes interfaces to the clock bus.
Most ChampSim operables inherit from both, while certain components like VMEM only inherit from one.

### DIFF
--- a/config/instantiation_file.py
+++ b/config/instantiation_file.py
@@ -414,6 +414,18 @@ def get_instantiation_lines(cores, caches, ptws, pmem, vmem, build_id):
     ), rtype='std::vector<std::reference_wrapper<champsim::operable>>')
     yield ''
 
+    yield from cxx.function(f'{classname}::component_view', (
+        'std::vector<std::reference_wrapper<champsim::component>> retval{};',
+        'auto make_ref = [](auto& x){ return std::ref<champsim::component>(x); };',
+        'std::transform(std::begin(cores), std::end(cores), std::back_inserter(retval), make_ref);',
+        'std::transform(std::begin(caches), std::end(caches), std::back_inserter(retval), make_ref);',
+        'std::transform(std::begin(ptws), std::end(ptws), std::back_inserter(retval), make_ref);',
+        'retval.push_back(std::ref<champsim::component>(DRAM));',
+        'retval.push_back(std::ref<champsim::component>(vmem));',
+        'return retval;'
+    ), rtype='std::vector<std::reference_wrapper<champsim::component>>')
+    yield ''
+
     yield from cxx.function(f'{classname}::dram_view', [f'return {pmem["name"]};'], rtype='MEMORY_CONTROLLER&')
     yield ''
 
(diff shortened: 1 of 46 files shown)


maccoymerrell:
The exact formatting and what information to display is up for debate. I would like to show which modules are connected to the cores and caches, but I am not sure as to the best way to do that.

ngober:
Initial thoughts:
- This is more or less what I thought this patch might look like. Good job on the first pass.
- All those changes to the tests are maximally oof. I wonder if there were a way to make them less tedious. I've been sitting on a draft of a test fixture that abstracts out the `std::array<champsim::operable*, N> elements` stuff for a while now. Should I knock the dust off of it and maybe we can put it in line ahead of this patch?
- I think you should add a `constexpr static std::string_view fmtstr{"{} {} {:35} {:>0}\n"}` member to `champsim::component`. This would help standardize the format down the inheritance hierarchy.
- I don't have a lot of great ideas right now regarding the module connections. It's an important part of this, though, and we need to find a way to do it. This is one of those stringification things that have had people begging C++ for a reflection API. We could generate a constexpr mapping from module types to strings, then read those at run time. But, we also erase the types of the modules in the caches/cores. Here's what I'm thinking:

    template <typename M>
    constexpr std::string_view module_name(const M& mod) {
      using namespace std::literals;
      if constexpr (std::same_as_v<bimodal, M>) { return "bimodal"sv; }
      else if constexpr (std::same_as_v<gshare, M>) { return "gshare"sv; }
      // ... and so on
    }

This would have to be generated, and I don't really like that. I would like to get away from relying on generated code so that we can do runtime configuration.

maccoymerrell:
I could probably modify print_dump() for our components with modules to invoke print_dump() for their modules as well. I think that would mean modules would need to inherit from component, which adds some additional problems (like we would need to actually invoke each module's calls for the warmup, phase, initialize, etc...).

ngober:
Is that a bad thing? It might be nice if modules were notified of the beginnings and ends of simulation phases.

maccoymerrell:
I standardized the formatting for the dump.

I also added made the modules inherit from component. I am not super familiar with how the modules are actually connected to the cache, core, etc..., so I tried a simple implementation of how the cache can invoke begin_phase(), print_deadlock(), print_dump(), etc... for each of its modules, but that didn't seem to work. I am sure I am missing something. Defining print_dump() in spp_dev didn't do anything, it still printed used the base function (print nothing) defined in component.

maccoymerrell:
You are welcome to bring in that stuff you were working on for test abstraction. I was changing all those tests last night and wondering what I was doing with my life.

ngober:
It will be another pull request, and maybe won't happen immediately. I'll take a look at it and see it's serviceable.

Issue #471, opened 2024-05-17
A moved `champsim::environment` causes segmentation fault


ngober:
**Describe the bug**
The top-level object for any ChampSim simulation is a `champsim::environment`. If that environment is moved for any reason, all of the member objects will be moved. Some of these maintain pointers internal to the environment that will then segfault if the simulation continues to run.

148f9f5cc1848411a0372f48ef5087b212816581 fixed a similar bug that occurred in the prefetcher modules, but this bug still exists between the core model and its instruction cache.

The cache is directly accessed by the core on each branch instruction:

https://github.com/ChampSim/ChampSim/blob/87f8b59b771efdd77edb90f5715fa63b5c3698c2/src/ooo_cpu.cc#L169

There is currently no way for the core to be notified that the cache has moved, or for the cache to notify the core that it has moved and the pointer should be updated.

**To Reproduce**
I have created a failing test to demonstrate this happening in e6e258d881d3de020e26f22a729c0b5a7d9de63b. If it seems obviously wrong to move from an object just after a pointer to it has been taken, that's because it is.

**Expected behavior**
Test 433 passes.

**Additional context**
This is probably a low-priority bug, since the current iteration of ChampSim constructs its environments in place and does not move them for the entire program. However, it would only require someone to want to create an environment in a function (say, with a runtime reading of the configuration file #424) for this bug to show up, and it would be a sneaky one. It should be fixed if possible.

ngober:
I'm going to close this as not planned after discussion in #564. I maintain that it is a good principle for the object to manage its pointers, but it's maybe out of scope.

Issue #422, opened 2024-01-02
Parameterized Modules


ngober:
### Discussed in https://github.com/ChampSim/ChampSim/discussions/415

<div type='discussions-op-text'>

<sup>Originally posted by **Setu-Gupta** December 15, 2023</sup>
I want to implement a replacement policy which has a tuning knob. I want to pass this constant to my policy as a parameter via the `CACHE::initialize_replacement()` call. Is it possible to convert this signature to something like `CACHE::initialize_replacement(uint32_t param)` so that I can pass a parameter to my policy when it is created? Moreover, if it is indeed possible to do so, how can I specify the value of this parameter in my config file?

I know that I can hardcode this value in my policy bu I don't want to do that. I want to be able to change this value via the config.</div>

The module system could benefit from parameterization in the configuration file. I'm opening this issue to centralize discussion and to call for assistance on implementing this.

I think the parameters should be specified as:

    { "L2C": { "prefetcher": {
        "class": my_pref,
        "path": // ...
        "params": {
            "cat": 1,
            "dog": 2016
        }
    }}}

That is, as an object with keyed parameters. This is preferable to specifying it as a list, since the list, while ordered, is not as descriptive and prone to bugs. I think the parameters should then be collected into a `std::map<std::string_view, long>` and passed to the initialization hook, if one is specified that supports it:

    my_pref::initialize_prefetcher(std::map<std::string_view, long> params) { }

There are a couple challenges: 
 - First, from the JSON perspective, the parameters may have arbitrary types, including objects themselves. Should state in our specifications that the parameters must be integers, and provide validation? Is there a better way? Some users may prefer doubles, and once we allow flexible types, we have a more complicated specification.
 - Second, it is not clear to me that the `initialize` hook is the best place for this. Probably the best place is in the module object constructor, but then we must require the user to specify a constructor that takes the model pointer, which is a clunky requirement.

    my_pref::my_pref(CACHE* c, std::map<std::string_view, long> params) : champsim::modules::prefetcher(c) { }

This would be a great way for someone looking to contribute to get familiar with the configuration and module systems.

PR #564, opened 2024-12-05
Runtime Config

### LINKED ISSUE #424: Runtime configuration

### LINKED ISSUE #548: Create a build-time mapping of module name strings to module types
To support runtime configuration, and possibly #546, we need a way to convert between the strings given in a configuration file ("hashed_perceptron", "lru", "ip_stride", etc.) to the type of the module object. As a stretch goal, we should see if we can map both directions.

There are a number of possible options, not all of which will work for our purposes.

- Some solutions use `__FUNCTION__` and parse the resulting string. ([link](https://blog.molecular-matters.com/2015/12/11/getting-the-type-of-a-template-argument-as-string-without-rtti/))
- Magic_enum also uses `__FUNCSIG__`. ([link](https://github.com/Neargye/magic_enum/blob/e046b69a3736d314fad813e159b1c192eaef92cd/include/magic_enum/magic_enum.hpp#L578))
- Since a string will be provided, it could be possible to create a constexpr function that lexicographically iterates through all possible identifiers, and returns a truthy value only for those which match a valid identifier. That might be totally off into the weeds, though.

Any solution must be generated at build-time, _not_ at configure time, and callable at runtime.

### LINKED ISSUE #549: Create a C++ replica of the Python configuration parser
The Python-based configuration step has a lot of advantages. It parses very easily, applies defaults in an intuitive way, and it is importable as a module. Its primary disadvantage is that it is a separate step that generates C++ files and compiles multiple executables for trivial changes. 

I would like a C++ version that we could compile and link in, to support runtime configuration.

As test cases, the new configuration parser should be able to parse all of the files in `test/config/compile-only/` to the same configuration that the Python parser does. More test cases may need to be added, but these are the ones in the documentation, so they are the expected behavior from the users.

ChampSim already leverages the nlohmann-json library for stats printing, so no new dependencies need to be introduced.

maccoymerrell:
This pr reworks modules for two purposes.
1. Support runtime configuration
2. Create a strong foundation for future expansions, including new interfaces and module types.

Modules are reworked to allow for identifying modules by a given string, creating an instance using this name, and binding them to a ChampSim component. This allows for runtime configuration to create and bind modules using nothing but the name of the module (provided by the user).
There are other ways that this system could be utilized in the future.

The module interface has been modified to be simplified, reducing the complexity of how modules work. This involved replacing the template-based system with a class-inheritance based system. Code inside cache.cc, ooo_cpu.cc, and modules.h/modules.cc has been greatly reduced.


maccoymerrell:
Moving forward with this, I think we could extend this to get us most of the way to runtime config.
In the future, all components of ChampSim could be module types, and if so would use a single constructor interface (a single string-keyed map of the arguments). This gets around the current limits of this system by allowing for any number of arguments to be passed to the modules.

This ties into runtime config, since these arguments could be specified and passed dynamically to the modules, and would require no change to ChampSim's base code to add or remove different parameters.
Inspired by other approaches, we could add wrappers around these maps so modules can specify things that are required, optional, and their default values (so we can error-out if a required value is not specified).

If we convert all of ChampSim's components to this, runtime config would essentially be done, with only the task of populating these maps using the json and relevant pointers (which we could also avoid by instead passing module names and having a module name -> reference map).

This really extends this pr a lot, but I think it more-comprehensively addresses why this hasn't been merged yet, and addresses #548 #471 #598 #424 #422 in what I consider a nice way.

ngober:
I'm going to talk a little bit about the problem that I was trying to solve when I wrote what we have now, so that my priorities (such as they were) are laid out. But, I want to start by saying I'm generally supportive of this work. If we can make it work cheaper and better, I'm for it.

I started out by wanting to reduce the number of pointer indirections in ChampSim. I wanted to know the concrete type of the module, and call the hooks as appropriate for the type, and I was trying to avoid virtual member functions because of the indirection. I started out trying compile time feature detection. All of the decltypes and constexpr bools in the module header is that. I was able to do pretty well with this. The system detects which set of parameters the module wants, and converts appropriately.

The current system uses type erasure to hide the module's type. I liked this because it didn't require the user to inherit from a certain supertype. It just detects that a type implements the hooks and builds the vtable appropriately. But, what ended up happening is that we needed a `CACHE*` in the prefetcher to call `prefetch_line()`. So, there needed to be some inheritance. I wasn't happy with how that turned out.

So, go ahead with this work if you think it will be fruitful. I tried to get as much work done at compile-time as I could, but maybe it isn't worth it. I would much rather have runtime configuration.

maccoymerrell:
Just merged in develop changes. Looks like this didn't cause any major issues (I will fix the lcov thing).

At the moment, all modules require a 3 item constructor (name, parent pointer "intern", and the new ModuleBuilder type).
The latter of these can be generated directly from the input config file when we have that infrastructure in place.
At the moment, these are unpopulated, so they don't provide any information. For example, replacement policies rely as normal on the cache pointer to retrieve sets and ways and prefetchers for prefetch_line(), which I don't think is bad but it does enforce that the cache must be fully instantiated before trying to create the modules. If we make all parts of ChampSim modules, then they will be in these new static maps, and we won't need to worry about pointer invalidation (hopefully).
Maybe override the copy constructors with asserts so they aren't accidentally moved? In some respects, I think pointers are unavoidable. Even if we avoid using them, users will likely find they need a pointer to the cache or other component for their work, so we should provide some way to do it without having them rewrite half of ChampSim's interfaces.

Back on track:
The idea here would be that the config file provides ModuleBuilder types for all modules in the instantiation files similarly to how we are doing the cache and core builders. Hopefully by the end of the weekend I can have everything converted over to this, including the caches and cores and memory controller (channels are a different matter, not sure how to approach yet).
With the addition of this system, alternative cache, core, and memory controller models could be substituted at runtime just like modules, which gives us a lot of possibilities!

Then, we are two steps away from runtime config
1. Runtime channel creation and connection (presumably these would be sized using the ModuleBuilders of the connected caches/cores like is being done currently)
2. Runtime .json config -> ModuleBuilder conversion. This can be achieved two ways:
- Direct read and conversion of the current .json format
- Read and conversion of a new .json format, with the old config.sh script acting as a conversion tool to the new .json format from the old.

The first seems the most straightforward, but with the addition of module parameterization I am unsure how easily the current format could be extended without things getting messy. A simplified format that fully defines the hierarchy (environment defines and builds [channels, cores, caches, memory controllers] which define and build [btbs, replacement policies, prefetchers, etc...]) would make runtime interpretation easiest (particularly if we drop all the inference the current config script is doing), but would probably be a lot more verbose and puts system creation more directly in user control (this could be both good and bad).

maccoymerrell:
I have just pushed a trial version of how I want the module system to work (it compiles but that's about all).
Actual functionality is untested, I imagine that _most_ configurations at this time will be non-functional for various reasons.

All champsim components (including core, cache, ptw, vmem, memory controller, replacement, btb, branch predictor, prefetcher, and channel) can now be swapped at runtime and utilize a standardized constructor.
This constructor takes an implementation, parent pointer, and name as an input (the name needing to be unique and this is verified during runtime, the implementation needing to be a registered module somewhere in champsim's binary).
Entries in the constructor are indexed by string, and can be templated to any type (this will require some work in the future).
By default, all entries are required and an incomplete ModuleBuilder will throw a runtime error. When accessing these, a flag and default value can be provided inside each module to fall back to a default value if an input parameter was not specified (avoiding the error).

Once fully completed, this system will allow us to remove the builder classes, lay the groundwork for runtime config and module parameterization, and avoid reference invalidations by avoiding all movement of champsim module classes. It will allow all champsim components to be swapped out by users without modifying the champsim codebase.

What still needs to be done:
1. Tests need to be updated to utilize the new system
2. Script parameter -> constructor parameter harmonization (this was done by the builder classes, but can be offloaded to the module constructors in _most_ cases. Cache latencies and default sizing are a notable exception.
3. Conversion of environment into a module, which will read and construct the system described in the config file (full runtime config).
4. Module support for soon-to-be environment class members instead of static-value constants (LOG2_BLOCK_SIZE, NUM_CPUS, etc...) (some of this has already been done in previous updates).

inc/modules.h
```diff
@@ -17,185 +17,410 @@
...
+      fmt::print("[MODULE] ERROR: parameter {} not found for module construction\n",name);
+      exit(-1);
+    }
+    try {
+      return std::any_cast<T>(parameters[name]);
+    }
+    catch(const std::bad_any_cast& caught) {
+      fmt::print("[MODULE] ERROR: Casting failed while retrieving parameter {}, is your parameter type correct?\n",name);
+      exit(-1);
+    }
```
ngober:
```suggestion
    if(auto it = parameters.find(name); it != parameters.end()) {
      try {
        return std::any_cast<T>(it->second);
      }
      catch(const std::bad_any_cast& caught) {
        fmt::print("[MODULE] ERROR: Casting failed while retrieving parameter {}, is your parameter type correct?\n",name);
        exit(-1);
      }
    } else {
      if(optional) {
        return default_value;
      }
      fmt::print("[MODULE] ERROR: parameter {} not found for module construction\n",name);
      exit(-1);
    }
    
```

You'll want to avoid throwing away the return value of `map::find()`. This algorithm traversed the map twice.

ngober (commented):
Just a single drive-by comment for now, to prove that I'm paying attention. I'll probably add more comments later.

maccoymerrell:
1 and 2 are mostly knocked-out now.
3 and 4 is the only thing left, which I plan to do by converting the environment to a module.
I will end up ripping out most of the build system as a result (for example, multi-config won't make sense anymore) so we don't need to template or do anything fancy with the environments.

The environment is going to be the json interpreter (unless you have a better idea) which will take in the json file and then be in charge of connecting everything and passing on relevant ModuleBuilders to the other components.

maccoymerrell:
Spent 11 hours today with co-pilot getting this put together. I think it's nearly there - but I haven't done any validation comparing the performance of identical configs across pre and post-runtime config.

Btw gcc 9-11 seems to choke on hashed perceptron. There might be some things in that implementation that are cruel to older compilers.

maccoymerrell:
Besides more testing, I think an alternative environment module should be included. This would interpret config files as explicit system configuration files (including channels) instead of relying on any formatting or implicit relationships.

To do this, module interfaces themselves should be registered, the environment can identify the model name and interface name and then instantiate a model of that interface. Modules should be constructed in the order they appear in the config file, so dependencies need to be resolved through proper ordering of the config file (for example, channels will likely need to come first).

The environment module then becomes very simple: determine the interface and model of each component defined in the config, instantiate it with all provided parameters (including refs/pointers to other previously instantiated components when necessary), and continue until all components are created.

maccoymerrell:
Explicit environment was added. It it far simpler than the legacy environment but requires explicit declarations within the config (sample provided).

*EDIT*
Finally got tests to pass - coverage up to 75%

inc/modules.h
ngober, line 1: I can tell that it's going to take me a while to work through this patch, at least because of this file. You've got some good ideas in here, I can tell, but we may need to work through them together.

inc/modules.h
```diff
@@ -17,185 +17,672 @@
...
-  constexpr static bool has_cycle_operate = decltype(cycle_operate_member_impl<T, Args...>(0))::value;
-
-  template <typename T, typename... Args>
-  constexpr static bool has_final_stats = decltype(final_stats_member_impl<T, Args...>(0))::value;
+}
 
-  template <typename T, typename... Args>
-  constexpr static bool has_branch_operate = decltype(branch_operate_member_impl<T, Args...>(0))::value;
+template <>
+struct fmt::formatter<champsim::modules::ModuleBuilder::nested_params_type> : fmt::formatter<std::string> {
```
ngober: We already link nlohmann::json. Could you use that here?

test/cpp/src/035-data-size.cc
```diff
@@ -171,7 +171,7 @@ TEST_CASE("A byte size prints something to libfmt")
 {
   using namespace champsim::data::data_literals;
   REQUIRE_THAT(fmt::format("{}", 5_B), Catch::Matchers::Matches("5 B"));
-  REQUIRE_THAT(fmt::format("{}", 5_kiB), Catch::Matchers::Matches("5 kiB"));
+  REQUIRE_THAT(fmt::format("{}", 5_kiB), Catch::Matchers::Matches("5 KiB"));
```
ngober, line 174: Hmm... This seems like an unrelated change. What happened here?

.github/workflows/valgrind.yml
ngober, line 1: I don't disagree that we should add a valgrind workflow, but this is probably not the patch for it. You should factor this out into a separate PR.

Makefile
```diff
@@ -1,343 +1,116 @@
 override ROOT_DIR = $(patsubst %/,%,$(dir $(abspath $(firstword $(MAKEFILE_LIST)))))
 
-# Customization points:
-#  - BIN_ROOT: at make-time, override the binary directory
-#  - OBJ_ROOT: at make-time, override the object file directory
-#  - DEP_ROOT: at make-time, override the dependency file directory
-BIN_ROOT:=bin
-OBJ_ROOT:=.csconfig
-DEP_ROOT:=$(OBJ_ROOT)
```
ngober, line 9: It's probably not wise to remove this comment.

Makefile
```diff
@@ -1,343 +1,116 @@
...
+executable_name := $(BIN_ROOT)/champsim
+test_main_name  := test/bin/000-test-main
 
-# Remove all configuration files
-configclean: clean compile_commands_clean
-	@-find $(module_dirs) -name 'legacy*' -delete &> /dev/null
-	@-$(RM) $(generated_files) _configuration.mk
+### Source discovery
+src_sources    := $(wildcard src/*.cc)
+module_sources := $(shell find branch btb prefetcher replacement -name "*.cc" 2>/dev/null)
```
ngober, line 28: You have removed the feature where a user can specify an external module directory. This is going to make the forthcoming user-contributed modules repo much harder to use.

Makefile
```diff
@@ -1,343 +1,116 @@
...
-$(inc_compile_commands_file): $(call rwildcard,$(base_include_dir),*.h)
-	python3 $(ROOT_DIR)/config/compile_commands/inc.py --champsim-dir $(ROOT_DIR) --config-dir $(OBJ_ROOT)
-
-$(test_compile_commands_file): $(call rwildcard,$(test_source_dir),*.cc)
-	python3 $(ROOT_DIR)/config/compile_commands/test.py --champsim-dir $(ROOT_DIR) --config-dir $(OBJ_ROOT)
-
-$(module_compile_commands_files): $(call rwildcard,$(call parent_dir,$@),*.cc)
-	python3 $(ROOT_DIR)/config/compile_commands/module.py --module-dir $(call parent_dir,$@) --champsim-dir $(ROOT_DIR) --config-dir $(OBJ_ROOT)
-
-compile_commands: $(src_compile_commands_file) $(inc_compile_commands_file) $(test_compile_commands_file) $(module_compile_commands_files)
```
ngober, line 322: You've removed the feature to make the `compile_commands.json` database. I know there will be users that miss this feature.

ngober (commented):
I'm keeping an eye on this PR. Some comments as we go along...

test/cpp/src/035-data-size.cc
maccoymerrell: This was a consistency change - the explicit environment config interpreter utilizes Ki/KiB to indicate kibi and kilobytes. I can revert this and change the explicit environment config to utilize ki/kib, but that seems inconsistent with the given M/G/T formatting for megabytes, gigabytes, and terabytes.

.github/workflows/valgrind.yml
maccoymerrell: I tend to add tests as problems arise and are resolved as I work - in this case, the lack of default constructors was causing MacOS Clang to fail and was only caught by valgrind.

I can factor the test out, but it will mean I need to
1. Remove from this pr
2. Open a new pr
3. Commit to work with current develop
4. Resolve conflict (revert to what is currently being used here) when this pr merges or vice-versa

test/cpp/src/035-data-size.cc
ngober: Oh, it seems I was wrong here. I assumed that the kibi- prefix was capitalized the same way as the kilo- prefix. Turns out, it isn't.

inc/bandwidth.h
```diff
@@ -85,4 +87,11 @@ class bandwidth
 };
 } // namespace champsim
 
+template <>
+struct fmt::formatter<champsim::bandwidth::maximum_type> : fmt::formatter<long int> {
```
ngober: Should this be a `format_as()`?

inc/defaults.hpp
```diff
@@ -17,120 +17,288 @@
...
-                             .reset_wq_checks_full_addr()
-                             .prefetch_activate(access_type::LOAD, access_type::PREFETCH);
-
-const auto default_ptw = champsim::ptw_builder{}.bandwidth_factor(2).mshr_factor(5).add_pscl(5, 1, 2).add_pscl(4, 1, 4).add_pscl(3, 2, 4).add_pscl(2, 4, 8);
+// O3_CPU parameters - types must exactly match get_parameter<T>() calls in ooo_cpu.h
+inline champsim::modules::ModuleBuilder default_core()
+{
+    return champsim::modules::ModuleBuilder{}
+        .add_parameter("clock_period", champsim::chrono::picoseconds{250})
+        .add_parameter("cpu", static_cast<uint8_t>(0))
```
ngober: You're saying `static_cast<uint8_t>(0)` in this file where I think it would be better to say `uint8_t{0}`. You could also say `.add_parameter<uint8_t>("cpu", 0)`. Maybe these cause the compiler to emit narrowing warnings, though...

inc/defaults.hpp
```diff
@@ -17,120 +17,288 @@
...
+        .add_parameter("l1i_bandwidth", champsim::bandwidth::maximum_type{1})
+        .add_parameter("l1d_bandwidth", champsim::bandwidth::maximum_type{1})
+        .add_parameter("fetch_queues", static_cast<champsim::modules::channel_module*>(nullptr))
+        .add_parameter("data_queues", static_cast<champsim::modules::channel_module*>(nullptr))
+        .add_parameter("l1i", static_cast<champsim::modules::cache_module*>(nullptr))
+        .add_parameter("branch_predictor_modules", std::vector<std::string>{})
+        .add_parameter("btb_modules", std::vector<std::string>{});
+}
+
+// CACHE parameters - types must exactly match get_parameter<T>() calls in cache.h
```
ngober, line 71: It might be nice if this wasn't the case. It would be better if `add_parameter<T>()` followed by `get_parameter<U>()` worked if `T` was convertible to `U` and non-narrowing. But, also, this isn't user-facing, so maybe the tight constraints are good defenses against bugs.

inc/legacy_environment.h
```diff
@@ -0,0 +1,62 @@
+/*
+ *    Copyright 2023 The ChampSim Contributors
```
ngober: Update the copyright year.

maccoymerrell:
I added a change here to formalize how nested modules are treated, should simplify how caches, cores, etc... obtain and construct submodules (prefetcher, replacement, btb, branch predictor).

maccoymerrell:
Hopefully the last set of major changes to this pr coming down the pipe now (modularization of champsim frontend). Infrastructure to test equivalence between new and old config system done, just need to launch the jobs.

maccoymerrell:
To expedite this pr, I am going to take all my frontend modularization and make that a separate pr once this is merged (it will also allow us to analyze the changes more thoroughly).

test/cpp/src/401-hit-latency.cc
```diff
@@ -16,12 +16,12 @@ SCENARIO("A cache returns a hit after the specified latency")
     constexpr auto hit_latency = 7;
     do_nothing_MRC mock_ll;
     to_rq_MRP mock_ul;
-    CACHE uut{champsim::cache_builder{champsim::defaults::default_l1d}
-                  .name("401-uut-" + std::string(str))
-                  .upper_levels({&mock_ul.queues})
-                  .lower_level(&mock_ll.queues)
-                  .hit_latency(hit_latency)
-                  .prefetch_activate(access_type::LOAD, access_type::RFO, access_type::PREFETCH, access_type::WRITE, access_type::TRANSLATION)};
+    CACHE uut{champsim::modules::ModuleBuilder{"uut_cache", "DEFAULT_CACHE", champsim::defaults::default_l1d()}
```
ngober: A lot of the uuts in the tests were named after the test number (here, `401-uut-N`). This makes debugging easier if you can just print the name. Your changes have glossed lots of them into `uut_cache`, which is less useful.

test/cpp/src/502-explicit-environment-config.cc
```diff
@@ -0,0 +1,685 @@
...
+#include "cache.h"
+
+using json = nlohmann::json;
+using champsim::modules::ModuleBuilder;
+
+namespace {
+
+// Build a minimal 1-core explicit config JSON.
+// This is the minimum viable config: channels, DRAM, vmem, PTW, caches, core.
+json minimal_explicit_config() {
```
ngober: I think you should put this into a file, and test other files too. This is really verbose to put in the test file, and it won't scale well.

Probably, this should just live in `/test/config/`. We already have a couple of configs in there. Go ahead and write tests for those.

src/environment.cc
```diff
@@ -0,0 +1,359 @@
...
+    }
+    out = result;
+    return true;
+  }
+  return false;
+}
+
+// Check if a string is an @-reference
+bool is_ref(const std::string& s) { return !s.empty() && s[0] == '@'; }
+std::string ref_name(const std::string& s) { return s.substr(1); }
```
ngober:
```suggestion
// Check if a string is an @-reference
std::optional<std::string> ref_name(const std::string& s) {
  if (!s.empty() && s[0] == '@')
    return s.substr(1);
  return std::nullopt;
}
```

This is a good place for an optional. Later, you can use `.has_value()` to check if the parse succeeded.

src/environment.cc
```diff
@@ -0,0 +1,359 @@
...
+  // Count cores for num_cpus
+  auto it = modules_by_type_.find("core");
+  if (it != modules_by_type_.end()) {
+    num_cpus_ = it->second.size();
+  }
+
+  // Compute deadlock threshold purely from parameter types.
+  // Every champsim::chrono::picoseconds parameter is a time value; the minimum
+  // is the time quantum and the maximum is the worst-case single latency.
+  // 2x max/min gives worst-case cycles, floored at 500.
```
ngober: Should this be related to the sum of the latencies, rather than the maximum?

src/cache.cc
```diff
@@ -32,63 +32,14 @@
...
-
-      cpu(other.cpu), NAME(std::move(other.NAME)), NUM_SET(other.NUM_SET), NUM_WAY(other.NUM_WAY), MSHR_SIZE(other.MSHR_SIZE), PQ_SIZE(other.PQ_SIZE),
-      HIT_LATENCY(other.HIT_LATENCY), FILL_LATENCY(other.FILL_LATENCY), OFFSET_BITS(other.OFFSET_BITS), block(std::move(other.block)), MAX_TAG(other.MAX_TAG),
-      MAX_FILL(other.MAX_FILL), prefetch_as_load(other.prefetch_as_load), match_offset_bits(other.match_offset_bits), virtual_prefetch(other.virtual_prefetch),
-      pref_activate_mask(std::move(other.pref_activate_mask)),
-
-      sim_stats(std::move(other.sim_stats)), roi_stats(std::move(other.roi_stats)),
-
-      pref_module_pimpl(std::move(other.pref_module_pimpl)), repl_module_pimpl(std::move(other.repl_module_pimpl))
+CACHE::CACHE(CACHE&& /*other*/) : champsim::modules::cache_module(champsim::chrono::picoseconds{})
```
ngober, line 35: This feels like a feature regression. Is there any reason why we aren't able to continue supporting move-construction here?

maccoymerrell:
@ngober 
Okay, validation is running.
Here was the validation process:
1. Build two champsim variants, current pr and develop
2. Add parameter dumps to develop
3. Build scripts to generate random (valid) configs, and compare the parameter dumps and final stats.
4. Run 20k different configs with random traces from SPEC2017 and check for differences.

Once this passes, I will push a final commit (pending any requested changes) and I will also attach some text files which will detail the coverage of different parameters, traces, and overall simulation speed (compared to develop).

In the process of bringing these two branches into sync, I have noticed a few issues that have or should be corrected:
1. PSCLs were defined for at least 5 levels regardless of actual levels. This causes valid virtual address slicing to trigger assertions since it would be checked for a 5 level page table instead of the actual number of levels>
2. DRAM-LLC Channels are uncapped. This should be fixed with a channel rework. Right now, channels act as queues, but they should instead act as channels (bandwidth in packets/cycle or even bytes if we decide to associate sizes with packets, no internal queue except for the 'inflight' packets)
3. VMEM minor fault penalty was in cycles not time, which caused some weird behaviors when PTWs had different frequencies. VMEM minor fault penalty is now given in ns for legacy configs.
4. Translation loop puts a ton of pressure on the L1D in terms of bandwidth. For some workloads, setting the MSHR size too low in the L1D effectively deadlocks the core. This might be by design, but it seems like we shouldn't starve translation LOADs if possible.

Some improvements being made:
1. Configs now have a _description field which prints when it's fed into ChampSim. Disclaimers were added to the documentation config files.
2. When the legacy environment provides a calculated value, a [DEFAULT] print will happen. For example, derivation of cache latency, latency splitting, or queue sizing.

ngober (commented):
Here's my next round of feedback. I'm starting to understand how the pieces fit together, and in general I like what you've got going on here.

inc/bandwidth.h
maccoymerrell: I can fix this

inc/defaults.hpp
maccoymerrell: I am not sure. It's bad for the champsim data and chrono types, but it was useful to ensure type consistency as parameters were passed from the config to the modules. I could go either way (unless there is a non-ugly way to differentiate the non-narrowing conversions from the narrowing and only allow the latter).

One thing I have added in the branch where I was working on modularizing the main simulation loop for champsim (primarily the trace readers / "workload sources") was a type registry, so that you could indicate in the config file a string type followed by a string of the value ("bandwidth" : "8"), and a registered reverse-formatter would convert the std::string to the registered type. These could be registered just like the modules so that if a module needs some special type (bad example: float), they could register float and then specify in the config {"float" : "0.8"} and ChampSim would automatically use their registered formatter to convert the type and pass it to the module.

inc/defaults.hpp
maccoymerrell: I think the only other way to do it without the warnings is the uint8_t{} approach. That might look cleaner. I will probably retain the static_cast for the pointer stuff.

src/cache.cc
maccoymerrell: I dislike the idea of allowing any module to move since we risk invalidating references. If the module system is working correctly, moving an object *should* be impossible (at least in my thinking).

I didn't want to fully remove this, in case there was a strong reason for keeping it, but I would prefer not to allow movement.

I would make it so moving any module causes an assert to trigger, but I think it would be up to all the final module types to enforce that, so not even worth doing since a user who defines their own module won't bother to.

src/environment.cc
maccoymerrell: Sum over min, yes that seems better. We expect min to be the shortest latency/fastest frequency.
Sum of all the times divided by the shortest time should guarantee we don't have a false deadlock?

inc/defaults.hpp
ngober: It's not wrong to keep it for pointers, but for integers, you're saying "take this `int` and turn it into a `uint8_t`", when you could just say "make me a `uint8_t`".

inc/defaults.hpp
ngober: I can think of a non-ugly way to do it, but it requires C++20 concepts. Tight constraints might just be the way to go here.

src/environment.cc
ngober: I've found that trying to algorithmize this tends to underestimate what kinds of things can go wrong. Go for the absolute most pessimistic case, round to the next thousand cycles, then triple it.

src/cache.cc
ngober: This is a place where thinking like a computer designer and thinking like a programmer are different things.

The computer designer thinks, "We have these things, and they just exist. They shouldn't ever go anywhere", and is correct.

But, the programmer says that taking something from the left side of your desk to the right side shouldn't be illegal. "If I need to, I should be able to, or at least have a good reason why I can't."

Even putting things in a `std::vector` might implicitly move things. If that ever ends up being something we want to do, I don't want this to get in our way if it doesn't need to.

test/cpp/src/502-explicit-environment-config.cc
maccoymerrell: Will do

src/environment.cc
maccoymerrell: Will do

test/cpp/src/401-hit-latency.cc
maccoymerrell: Will do

src/cache.cc
maccoymerrell: Here's the reason you can't: 
system_diagram

Right now, ChampSim is heavily interconnected, and under the current module system we don't have to manage all these references, but under a system where modules can be moved, we do. The only way I see of making that possible is an indirection layer on top of the current one that can simplify updating references, but all that does is shift the 'problem' of having an unmovable object onto the 'problem' of having an unmovable reference table, and since the objects are heap-allocated anyway, we gain nothing except an extra layer of indirection.

ngober: In your diagram, things generally only have pointers to channels or to user modules (that themselves are leaf modules). The only exception is the core model's pointer to L1I. Sure, there are a lot of arrows, but they're all pretty tame.

maccoymerrell: Seems arbitrary to pin some modules and not others, and one that is actually not enforceable given the changes already in this pr:
1. References to modules can be passed as parameters in the explicit config file. Pinning channels doesn't fix the idea of stale references.
2. Cache, core, dram, ptw, vmem exist as interfaces now, the current CACHE is just one implementation. A user could write an alternative and substitute it in, thus the cache is now a user module. As is everything else.
3. I don't see any possible utility in the ability to move modules. Surely there is nothing you couldn't do on a container of references to the modules that you could do to a container of the modules themselves?
4. Virtual interfaces force us into pointer land, all modules are passed as pointers, so even the environment is only holding vectors of pointers, instead of the modules themselves.

ngober: I don't want us to get bogged down by this, so I'll be brief with my response, then concede.

I'm not being arbitrary. I do not think anything should be non-copiable, non-movable without specific reason. If the reason is general rather than specific, then I argue that architectural changes are needed. This applies to channels, to caches, to cores, etc. The complaint I raise came from the work I had done to move ChampSim towards that ideal, and this comes across as a regression.

You're right that throwing an assert here alerts us to places where the assumptions are violated. For that, I'll back off of this point.

maccoymerrell:
All validation tests passed, attached is the coverage report. Simulation speed was nearly identical, 0.9932 RT/CT speed ratio, so RT is ever so slightly faster (could also be that CT runs first then RT).
[coverage_report_final.txt](https://github.com/user-attachments/files/26835593/coverage_report_final.txt)

inc/modules.h
```diff
@@ -560,31 +560,11 @@ struct module_base {
...
-                                                [[maybe_unused]] access_type type, [[maybe_unused]] uint32_t metadata_in) {
-        return prefetcher_cache_operate(addr,ip,cache_hit,useful_prefetch,champsim::to_underlying<access_type>(type),metadata_in);
-      }
-      virtual uint32_t prefetcher_cache_operate([[maybe_unused]] champsim::address addr, [[maybe_unused]] champsim::address ip, [[maybe_unused]] bool cache_hit, [[maybe_unused]] bool useful_prefetch,
-                                                [[maybe_unused]] std::underlying_type_t<access_type> type, [[maybe_unused]] uint32_t metadata_in) {
-        return prefetcher_cache_operate(addr.to<uint64_t>(),ip.to<uint64_t>(),cache_hit,type,metadata_in);
-      }
-      virtual uint32_t prefetcher_cache_operate([[maybe_unused]] uint64_t addr, [[maybe_unused]] uint64_t ip, [[maybe_unused]] bool cache_hit,[[maybe_unused]] std::underlying_type_t<access_type> type, 
-                                                [[maybe_unused]] uint32_t metadata_in) {return metadata_in;}
+                                                [[maybe_unused]] access_type type, [[maybe_unused]] uint32_t metadata_in) {return metadata_in;}
```
ngober: Are we deliberately removing these legacy overloads?

maccoymerrell: Yes. I originally added it to support legacy modules, but you had pointed out before that this prevents us from making the hooks purely virtual (requiring an override in the module). I think I still need to make the hooks pure virtual in another commit, but it will prevent situations where someone thinks they are overriding a hook but they aren't actually.
The override keyword could also be used, but I can easily see someone leaving that out, and since neither ChampSim or the compiler yells at them, they don't realize the problem.

Long story short, let the compiler enforce this. Users will already need to make some modifications to existing modules to work in the new system, lets just make this another step.

Open to objections on this though, in case you see good reason to leave this in. I was also concerned that the nested calls were giving us a lot of indirect branches and causing performance loss, but it seemed like it had negligible impact.

ngober: If users are making big changes with the registration system, I see no harm in also saying that they need to update their function signatures also.

ngober:
I'm really pleased that the validation went well. These checks all look good. We're adding lots of test coverage, and this is a big improvement to the user experience (not having to compile so much).

Now, would you update the documentation to reflect the new behavior?

maccoymerrell:
Yay! Documentation!

maccoymerrell:
Documentation has been updated.
Besides updating the existing documentation, I added:

- Documentation on user calls for module system, modulebuilders, etc...
- Tutorial on creating modules and interfaces, converting old modules to the new system
- Information on how to use the explicit config format
- MSL documentation on the lru table and existing stat method types
- Fixed some existing issues with the documentation (publications didn't seem to be working properly)
- Documentation on the main interfaces (core_module, cache_module, memory_controller_module)

maccoymerrell:
Just pushed some changes to the doc, name changes to LEGACY_ENVIRONMENT and ENVIRONMENT.
get_submodules() was slightly changed to allow setting a bit to indicate the submodule is required. Useful to distinguish whether a child module is required or not (prefetcher vs. replacement policy). A missing required submodule leads to an assertion at build time.

maccoymerrell:
Saw that explicit config file didn't define heartbeat frequency. Tried defining it and changing it, nothing happened. Seems that the addition of the event listeners severed the arg passing (or perhaps as the result of a merge error on my part). This is now fixed, and is being passed properly now. Tests added.

maccoymerrell:
PSCL is now passed as a vector. I also note that for the explicit config format it is possible someone may not define some upper-level pscls, and as a result skip levels of the ptw walk (since it default starts the walk at the top level pscl).
If you, for example, don't defined the top-level pscl, it would have previously skipped walking the page table entirely for that level, a bug. 

We now pad it with 0-sized pscls.

ngober:
I have merged the release candidate into master, and I want to turn my focus back to this patch. Is there anything that would prevent this from being merged to develop?

maccoymerrell:
No, testing came back clean. Doc is up-to-date. Further improvements (for the modularized trace providers + wrong-path) are separately being worked on, off this branch.

Issue #570, opened 2024-12-30
Issue with Update Replacement State not tracking hits


ChristopherThomasHill:
**Describe the bug**
update_replacement_state() doesn't get called on fills whenever replacement_cache_fill() is not defined. From the documentation it states that update_replacement_state() should be called on both hits and misses (at fill time) whenever replacement_cache_fill() isn't defined.

**To Reproduce**
1. Go the following assertion to srrip.cc into the update_replacement_state(): `assert(hit);`
2. Compile and run your code with any functioning trace that gets LLC misses.
3. The assertion never fails and the program completes normally

**Expected behavior**
The program should terminate on the assertion during the first LLC miss since update_replacement_state() should be called with hit equal to false.

**Desktop:**
 - OS: Ubuntu 20.04
 - Compiler GCC
 - Version 11

**Additional context**
This error stems from the following code segment in cache.h

`if (hit || replacement::has_cache_fill<decltype(r), uint32_t, long, long, champsim::address, champsim::address, champsim::address, access_type>)`

Since, this check surrounds the call to update_replacement_state(), it won't be called for misses when the replacement_cache_fill() isn't defined.

PR #571, opened 2024-12-30
Fix Issue #570: Unique if condition for impl_replacement_cache_fill

ChristopherThomasHill:
This is a first attempt at resolving Issue #570. It does have some code repetition issues, but does resolve the main bug with it not calling `update_replacement_state()` for misses when `replacement_cache_fill()` isn't defined.

### DIFF
--- a/inc/cache.h
+++ b/inc/cache.h
@@ -512,7 +512,27 @@ void CACHE::replacement_module_model<Rs...>::impl_replacement_cache_fill(uint32_
       r.replacement_cache_fill(triggering_cpu, set, way, full_addr, ip, victim_addr, type);
 
     else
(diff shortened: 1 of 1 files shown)


maccoymerrell:
This looks like it should resolve the issue.
Do you think you could add a change to our existing tests that would catch this bug you are describing?
Test 442-replacement-interface.cc or 443-replacement-interaface-selection.cc would probably be good candidates, and only need minor additions to do this.

adeelurehman:
Finally getting through the backlog, looks good. Go ahead and merge it in.

ngober:
I don't want to merge this until the tests are written. Adding tests will ensure this issue doesn't come up again.

PR #574, opened 2025-01-07
Fix for issue #573. Added test to confirm.

joshualmashburn:


joshualmashburn:
Still getting the same deadlock as in #573 . While this does correct the bug identified there, I'll have to do some deeper debug to diagnose this one.

ngober:
Do you think this bug is severe enough to hotfix to master?

joshualmashburn:
I think so, for correctness of the rename implementation. This does leave a variable number of registers (depending on the trace) occupied for the duration of execution. The more registers whose first reference in the trace is a read, the fewer usable physical registers. I don't think it's a four-alarm issue, but it may become more common and the fix was simple and something obvious I missed.

PR #576, opened 2025-01-08
Generate compile_commands.json for LSP support

cmolder:
I've been playing with the 2024 release of ChampSim and I've figured out how to get [clangd](https://clangd.llvm.org/) (and, in theory, other LSPs which use compile_commands.json) to work with ChampSim. This should make ChampSim development a more pleasurable experience for VS Code with the [clangd extension](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd).

This pull request adds a few new functions to the build scripts that generate a [compile_commands.json](https://clang.llvm.org/docs/JSONCompilationDatabase.html) file. This JSON file describes how each source file is built, which helps LSPs understand how ChampSim is built.

I did a little reverse engineering to figure out how to set up this file automatically. Initially, I used [bear](https://github.com/rizsotto/Bear) to automatically generate a `compile_commands.json`, and realized that the individual commands it generated are quite simple. So, I went ahead and replicated a similar `compile_commands.json` by reading through each module and source file. This *should* work just as well as the "true" output from bear, but it should probably be tested a bit first.

Why not just use bear? First, it's not installed on many systems (including ours at UT), and building and using it can be a pain since it requires a few external libraries. This should avoid the need to use bear or another tool while giving us all the benefits of having a `compile_commands.json` file.

The only issue I have seen so far is that clangd will complain about a missing `module_decl.inc` and `legacy_bridge.h` file if the `.csconfig` directory is fresh and has not had `make` called yet. These two files are generated when calling `make`, but are not cleaned out when running `make clean`. It might be a good idea to pre-generate these files in the config scripts, if that's possible. But, this is a very minor issue and goes away as soon as `make` is run for the first time.

I'm hoping that the use of clangd can help improve ChampSim's development, _e.g._ by helping enforce [include what you use (IWYU)](https://include-what-you-use.org/) and spotting warnings/errors before compiling.

Let me know what y'all think, and if you have any suggestions! Thank you :)


ngober:
This is an interesting addition. I do have a soft spot for good tooling improvements.

I think I want to steer you in a more static direction. You are generating `compile_commands.json` at configure-time. However, one of the goals for the 2025 release is #424, which moves everything from configure-time to run-time. In fact, the absence of `module_decl.inc` and `legacy_bridge.h` are exactly due to the beginnings of this migration.

I think you should aim to have a make target for `compile_commands.json` present in the top-level makefile. That probably makes it tricky for the main object file, but I'm hopeful it won't be too hard.

cmolder:
Thanks for the feedback, Nathan. I've updated the PR with two main changes:

1. `compile_commands.json` files are now generated using Makefile rules, specficially `make compile_commands` and`make compile_commands_clean`.

2. I've separated the `compile_commands.json` files for each module, and for the `src/` and `test/` directories (see [here](https://clangd.llvm.org/design/compile-commands#where-do-compile-commands-come-from)). This should greatly speed up clangd indexing. For instance, we don't have to index all the tests if we're just looking at source files. As long as nobody tries explicitly importing a test or module file, this shouldn't cause any issues.

I'd like to rely a little more on the Makefile variables so I don't have to infer things like the ChampSim root directory or location of modules. So, I might simplify `config/compile_commands.py` further such that one run of the script only generates one module's file (similarly for src and test), and we can just pass Makefile variables as arguments. This would also be nice as we could explicitly define compile_commands.json files as rules/dependencies.

ngober:
Nice, that's in line with what I was envisioning. This is the general path that I want to put us on, to do more things statically and rely less on the configuration step.

As you're tweaking your script, one thing I note in particular is that you've hardcoded the '.csconfig' path. This is actually derivative from a make-time customization point, see makefile line 8. So, you should at least pass that into your scripts. Speaking for myself, I actually rely on that one often in HPC situations where I want to keep the object files locally on a node.

cmolder:
Thanks! I made a few more commits and I think this is pretty close to the finished product (after some testing, of course). Here's the changes:

- The compile_commands.py script is now broken into separate scripts which are called for each compile_commands.json file we create, e.g. for each module, src/, test/, etc.
- We now also generate a compile_commands.json file for the headers in inc/. This fixes a lot of issues with missing headers in clangd when editing the header files.
- Unneeded dependencies for the `compile_commands` and `compile_commands_clean` make rules are removed. Now, it should just create/clean the relevant files without building unnecessary object files.

The Python scripts could possibly benefit from a little refactoring (e.g. define a class for compile commands entries), but overall I think it's in good shape.

config/compile_commands/inc.py
```diff
@@ -0,0 +1,69 @@
...
+                *get_options(champsim_dir / "absolute.options"),
+                f"-I{config_dir}",
+            ],
+            "directory": f"{champsim_dir.absolute()}",
+            "file": f"{inc_file.absolute()}",
+        })
+
+    # Save the compile_commands.json file
+    with open(inc_dir / "compile_commands.json", "wt", encoding="utf-8") as f:
+        f.write(json.dumps(entries, indent=2))
```
ngober, line 41:
```suggestion
    (inc_dir / "compile_commands.json).write_text(json.dumps(entries, indent=2))
```

I recently discovered that this method exists. Could be nice here.

ngober (approved):
This looks like a really clean patch. I'm impressed with your Python style. This seems mostly fit to merge, unless you want to refactor a bit. You haven't included any tests, but I'm struggling to identify a good place to factor out testable code. Most of this is just interacting with the filesystem. Overall, I'm pretty happy with it.

cmolder:
Thanks, Nathan! I went ahead and refactored a bit, hopefully this is a little easier to work with and extend :) I also incorporated your suggestion into CompileCommandManfiest.save.

It would be nice to add tests somehow. One idea could be to add a GitHub action which generates these files and runs [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) over the code to check for warnings/etc.

ngober:
I definitely use clang-tidy in development, but there's currently not something in the CI enforcing it. I've had my eye on something like [this](https://github.com/marketplace/actions/clang-tidy-review).

I think it could be good to do, but we should be careful. I'm a big fan of `readability-function-cognitive-complexity`, but it could be really frustrating to contributors if we enable it for PRs.

Maybe that is a second step. Maybe we need to merge this, and look at ways to improve the CI next. What do you think?

PR #581, opened 2025-01-17
Format Json

akihikodaki:
git clang-format is a convenient tool that integrates Git and clang-format. However, it will complain if champsim_config.json is modified:

Configuration file(s) do(es) not support Json: /home/odaki/ChampSim/.clang-format
error: `clang-format -lines=7:7 -lines=31:31 -lines=71:71 champsim_config.json` failed

Let's add an empty configuration for Json first to satisfy git clang-format. Running clang-format with this configuration actually fixes some style inconsistency in champsim_config.json.


.clang-format
```diff
@@ -22,5 +22,7 @@ IncludeCategories:
     Priority: 3
 PointerAlignment: Left
 SortIncludes: true
+---
+Language: Json
```
ngober, line 26:
```suggestion
Language: Json
IndentWidth: 4
```

Many of the changes in this patch come down to indentation. Could we add an entry to this file to reflect the current state?

ngober (changes requested):
Thanks for the patch. I hadn't realized that clang-format could format JSON. That's rather nice. I think you should also modify `.github/workflows/main.yml` so that the automatic linter is aware of the JSON files. Otherwise, this will be a one-time fix.

.clang-format
akihikodaki: That is possible, but perhaps indenting with 2 columns may be better for consistency with most source files, which are in C++.
If you still prefer to maintain the current indent width for Json, please let me know; I'll apply the suggested change.

ngober: The maximum indent depth of any of the JSON files we have is 5 levels. With 4 spaces, that would be 20 columns, probably more than we want. I suppose 2 columns is okay. But, let's make it explicit. Do you know how the `IndentWidth` was determined in your suggestion? Was it inherited from the prior language, or some invisible default?

akihikodaki: My current understanding is that it falls back to the LLVM style, which sets `IdentWidth` to 2.

ngober: OK, this seems fine. We can always update it later if we want. Please update `.github/workflows/main.yml` so that this is enforced across the codebase.

ngober (changes requested):

.clang-format
akihikodaki: I updated that file. Adding `test` to format JSON files in that directory also caused many C++ files to be formatted.

PR #582, opened 2025-01-17
fix: deprecated options in README example

santimc:
the `--warmup_instructions` and `--simulation_instructions` options were deprecated, changed for the "-" version in the README


ngober:
Great catch! Let's get this merged...

PR #584, opened 2025-01-27
ooo_cpu: Do not copy RegisterAllocator

akihikodaki:
O3_CPU::schedule_instruction() accidentally copies RegisterAllocator. Use a reference to avoid this copy. This change gives 7x speedup.

### DIFF
--- a/src/ooo_cpu.cc
+++ b/src/ooo_cpu.cc
@@ -432,7 +432,7 @@ long O3_CPU::schedule_instruction()
   for (auto rob_it = std::begin(ROB); rob_it != std::end(ROB) && search_bw.has_remaining(); ++rob_it) {
     // if there aren't enough physical registers available for the next instruction, stop scheduling
     unsigned long sources_to_allocate = std::count_if(rob_it->source_registers.begin(), rob_it->source_registers.end(),
-                                                      [alloc = std::as_const(reg_allocator)](auto srcreg) { return !alloc.isAllocated(srcreg); });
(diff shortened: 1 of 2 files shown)


maccoymerrell:
Can confirm that this change leads to a significant performance increase.
From my own benchmarks:
ooo_cpu::operate_cycle()
  W/o fix: 2.87245 us (avg)
  W fix: 378.58 ns (avg)

ngober:
Did this bug make it into the release? Is this important enough to pull to master?

ngober:
@akihikodaki Could you please rebase your changes onto master? If I simply change the base branch on Github, it will merge the entire develop branch.

akihikodaki:
@ngober Rebased with commit 6fcf78df9c62d302b6e3da5653112e40d2090918.

ngober:
@maccoymerrell Is it possible for you to merge your benchmark into here? I'd like to add in some measure of protection against something like this slipping in.

maccoymerrell:
I opened a pr with it on the fork.

cmolder:
There's also another case in line 474, adding the & in front of alloc there too gives me a ~3x speedup.

akihikodaki:
> There's also another case in line 474, adding the & in front of alloc there too gives me a ~3x speedup.

Fixed this with commit eb24fb11ac35496cbbf37718dcbf5dbb09e73d60.

ngober:
I would like to merge akihikodaki/ChampSim#1 into this, so that we have some measurements in the future.

PR #585, opened 2025-01-31
Replace uses of <random> with Boost.random

cmolder:
While building ChampSim on two different machines, I noticed that I was getting different results for the exact same configuration. After some investigation, I found out it was due to the machines (Ubuntu 20.04 and Arch Linux) having different libstdc++ versions with different implementations of [std::uniform_int_distribution](https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution). This led to the shuffling of vmem pages inside std::shuffle to differ between the two machines. I'm happy to provide more details on the exact issue I observed and the two systems in question.

As a quick fix for this, I replaced all uses of <random> (which thankfully weren't many) with [Boost.random](https://www.boost.org/doc/libs/1_87_0/doc/html/boost_random.html), since Boost's implementation is platform-agnostic. This appears to have resolved the issue.

To avoid this in the future, it might be a good idea to add a CI test which ensures that the outputs of the simulations match for every compiler.

### DIFF
--- a/src/vmem.cc
+++ b/src/vmem.cc
@@ -17,10 +17,12 @@
 #include "vmem.h"
 
 #include <cassert>
+#include <boost/random.hpp>
 #include <fmt/core.h>
 
 #include "champsim.h"
 #include "dram_controller.h"
+#include "util/algorithm.h"
 #include "util/bits.h"
 
 using namespace champsim::data::data_literals;
@@ -70,7 +72,7 @@ void VirtualMemory::populate_pages()
 void VirtualMemory::shuffle_pages()
 {
   if (randomization_seed.has_value())
-    std::shuffle(ppage_free_list.begin(), ppage_free_list.end(), std::mt19937_64{randomization_seed.value()});
+    champsim::shuffle(ppage_free_list.begin(), ppage_free_list.end(), boost::mt19937_64{randomization_seed.value()});
 }
 
 champsim::dynamic_extent VirtualMemory::extent(std::size_t level) const
--- a/inc/util/algorithm.h
+++ b/inc/util/algorithm.h
@@ -18,6 +18,7 @@
 #define UTIL_ALGORITHM_H
 
 #include <algorithm>
+#include <boost/random.hpp>
 
 #include "bandwidth.h"
 #include "util/span.h"
@@ -46,6 +47,33 @@ long int transform_while_n(R& queue, Output out, bandwidth sz, F&& test_func, G&
   std::transform(begin, end, out, std::forward<G>(transform_func));
   queue.erase(begin, end);
   return retval;
+}
+
+/**
+ * This is a re-implementation of std::shuffle that uses Boost's implementation
+ * of uniform_int_distribution instead of the standard library's.
+ *
+ * This is necessary because std::uniform_int_distrubtion is implemented
+ * differently across platforms/compilers/etc. This can cause ChampSim to
+ * produce different results for the same seed on different platforms.
+ *
+ * Since boost's implementation is platform-agnostic, this version should yield
+ * identical results regardless of platform.
+ *
+ * \ref https://en.cppreference.com/w/cpp/algorithm/random_shuffle
+ * \ref https://www.boost.org/doc/libs/1_87_0/doc/html/boost_random/reference.html
+ */
+template <class RandomIt, class URBG>
+void shuffle(RandomIt first, RandomIt last, URBG&& g)
+{
+  typedef typename std::iterator_traits<RandomIt>::difference_type difference_type;
+  typedef boost::random::uniform_int_distribution<difference_type> distribution_type;
+  typedef typename distribution_type::param_type param_type;
+
+  distribution_type D;
+  for (difference_type i = last - first - 1; i > 0; --i) {
+    std::swap(first[i], first[D(g, param_type(0, i))]);
+  }
 }
 } // namespace champsim
 
--- a/inc/vmem.h
+++ b/inc/vmem.h
@@ -21,7 +21,6 @@
 #include <deque>
 #include <map>
 #include <optional>
-#include <random>
 
 #include "address.h"
 #include "champsim.h"
--- a/replacement/random/random.h
+++ b/replacement/random/random.h
@@ -1,14 +1,14 @@
 #ifndef REPLACEMENT_RANDOM_H
 #define REPLACEMENT_RANDOM_H
 
-#include <random>
+#include <boost/random.hpp>
 
 #include "cache.h"
 #include "modules.h"
 
 struct random : public champsim::modules::replacement {
-  std::mt19937_64 rng{};
-  std::uniform_int_distribution<long> dist;
+  boost::mt19937_64 rng{};
+  boost::random::uniform_int_distribution<long> dist;
 
   explicit random(CACHE* cache);
   random(CACHE* cache, long ways);
--- a/replacement/ship/ship.cc
+++ b/replacement/ship/ship.cc
@@ -2,7 +2,7 @@
 
 #include <algorithm>
 #include <cassert>
-#include <random>
+#include <boost/random.hpp>
 
 #include "champsim.h"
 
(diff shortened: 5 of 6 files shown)


src/vmem.cc
```diff
@@ -70,7 +72,7 @@ void VirtualMemory::populate_pages()
 void VirtualMemory::shuffle_pages()
 {
   if (randomization_seed.has_value())
-    std::shuffle(ppage_free_list.begin(), ppage_free_list.end(), std::mt19937_64{randomization_seed.value()});
+    champsim::shuffle(ppage_free_list.begin(), ppage_free_list.end(), boost::mt19937_64{randomization_seed.value()});
```
ngober, line 75: Is it necessary to reimplement `shuffle()` rather than simply pass the boost engine?

cmolder: From what I can tell, yes. Simply changing the boost engine modified the results but there was still a discrepancy. I think it's because libstdc++'s version of shuffle uses std::uniform_int_distribution, which can differ between versions.

This [blog post](https://codingnest.com/generating-random-numbers-using-c-standard-library-the-problems/) I found explains many issues of \<random\> in detail. Most relevant here is the part about how "Distributions are nonportable".

maccoymerrell: From what I understand, mersenne twister is invariant across different platforms. It is a strictly defined algorithm.
I would assume that this failure is because of std::shuffle itself, and swapping out shuffle for champsim::shuffle should be sufficient by itself while still using mersenne twister and not using boost::random.
Could we confirm this?

cmolder: @maccoymerrell Yup, it looks like this is the case. Putting back std::mt19937_64 but keeping champsim::shuffle yields identical results on both g++-9 and 14.

cmolder:
I know adding boost to a project can be a big change, so I went ahead and limited the usage of it to a single header file. Consequently, if we write our own implementation of `uniform_int_distribution`, we can get away with not using boost here.

ngober:
Is it as simple as something like

```
template <typename T>
class uniform_int_distribution {
  T a;
  T b;
public:
  uniform_int_distribution(T a_param, T b_param) : a(a_param), b(b_param) {}

  template <typename Gen>
  T operator()(Gen& generator) const {
    const auto width = b-a;
    constexpr auto genwidth = Gen::max() - Gen::min();
    const auto upperbound = Gen::min() + width*(genwidth / width);

    T retval = generator();
    while (retval >= upperbound) {
      retval = generator();
    }

    return ((retval - Gen::min()) % b) + a;
  }
```

Of course, that's a very, _very_ loose sketch and doesn't include the full interface, but is it more complicated than just discarding out-of-range values from the generator? I'd want to test this to the moon and back, but I think it's conceptually pretty simple.

On the topic of adding boost, because we manage dependencies with vcpkg, I'm not too troubled by adding an additional package dependency. It will require all users to run `vcpkg install` again, but if it's justified, it's justified. My thinking is this: external libraries are (ideally) tested and maintained by someone else, so it's generally worth our brain bandwidth to rely on them, unless the thing is so simple we can just write it ourselves.

If we can find justifiable uses of Boost::random somewhere else, then I think we should add it. Example: our dependency on nlohmann::json, currently used only for stats output, but hopefully we can use it to parse the configuration soon.

maccoymerrell:
I don't think you need to discard anything actually.
I would assume that a random number generator of any kind produces a fully uniform distribution, so wouldn't it be sufficient to just modulus the generated value by the width of the uniform distribution? Why the while loop and upper bound?

```
template <typename T>
class uniform_int_distribution {
  T a;
  T b;
public:
  uniform_int_distribution(T a_param, T b_param) : a(a_param), b(b_param) {}

  template <typename Gen>
  T operator()(Gen& generator) const {
    const auto width = b-a;

    T retval = generator() % width;
    return retval + a;
  }
};
```

ngober:
If you just slap a modulus on it, it biases the distribution towards lower numbers. There's an explanation in the blog post that Carson linked above. So, I set the upper bound at an integer number of widths over `Gen::min()`, then dropped things that fell between that and `Gen::max()`.

maccoymerrell:
I don't believe this has a bias in distribution towards lower numbers. I don't know why it would mathematically. Modulus operations are the backbone of integer-based cryptography, so I highly doubt that is true.

Here is a short test I did in python.
Figure 2025-02-01 143702

ngober:
That's not the right way to think about this or to evaluate the claim.

Consider the following scenario:
 - The generator yields integer values in [0,1,2] with equal probability
 - The distribution maps to [0,1] with equal probability

Before proceeding, I note that this scenario is entirely within the specification. The values are smaller than we might consider common, but they are legal values.

If we perform a modulo 2 on the generated values, we will map `{ [0,2] => 0, [1] => 1 }`. This is far from a uniform distribution. If instead we discard all 2s, we will map `{ [0] => 0, [1] => 1 }`. This will hold true for any instance where the range of the generator is not an integer multiple of the range of the distribution. Since the values `a` and `b` are given at runtime, we cannot assume anything about them (beyond the specification).

We also cannot evaluate this claim with your python script. What you have presented is but one realization of the probability distribution. Your modulus will produce the values of 0-15 with a probability `16/184467440737095516 ~= 8.67e-17` higher than values 16-99. Your run of 100million evaluations simply cannot illustrate this bias above the noise. You would need at least 10 orders of magnitude more evaluations to illustrate what I'm describing. None of this, however, denies that the bias is in fact present.

maccoymerrell:
What kind of performance overhead to the randomness would we experience? Mainly, how many resamples?
I guess performance is worst when the bias is worst, mainly when the width is just over half of the genwidth? Something like a 2x overhead on the average worst case? I guess that is acceptable, as long as our generators remain over 2x the desired field width then I don't see this being a major problem.

Here is data showing as much

Figure 2025-02-01 162414

maccoymerrell:
Figure 2025-02-01 162746

cmolder:
After looking at a bunch of sample implementations, I've written a custom implementation of `uniform_int_distribution` that should be a drop-in replacement to the one in `<random>` while being platform-agnostic.

inc/util/random.h
```diff
@@ -0,0 +1,152 @@
...
+    // Get the ranges of the URBG and distribution
+    static_assert(URBG::max() > URBG::min(), "URBG must have a range greater than zero");
+    constexpr u_result_type range_g = static_cast<u_result_type>(URBG::max() - URBG::min());
+    const u_result_type range_d = static_cast<u_result_type>(params.b() - params.a() + 1);
+
+    if (range_g < range_d) {
+      throw std::invalid_argument("URBG range is too small for the distribution");
+    }
+
+    // Use rejection sampling to avoid modulo bias
```
cmolder, line 140: re: the branch prediction issue, I think this branch should be easy to predict. We should rarely enter the while loop, unless the first call to `g()` happens to be an extremely small number. For std::mt19937_64, the odds of that happening are super low since (I think) it uniformly samples from `0` to `max(uint64_t)`.

inc/util/random.h
```diff
@@ -0,0 +1,152 @@
...
+    const u_result_type reject_limit{range_g % range_d};
+    u_result_type sample{g() - URBG::min()};
+    while (sample <= reject_limit) {
```
cmolder, line 143: re: the branch prediction issue, I think this branch should be easy to predict. We should rarely enter the while loop, unless the first call to `g()` happens to be an extremely small number. For std::mt19937_64, the odds of that happening are super low since (I think) it uniformly samples from `0` to `max(uint64_t)`.

cmolder: Also, while I was playing with this in compiler explore, it looks like the compiler optimizes out the `- URBG::min()` when it's zero, which is the case most of the time (a nice little bonus).

ngober:
I like tossing out low values rather than high values. That's clever. We should put plenty of tests on this. Let's test at least the degenerate `[0,1,2] => [0,1]` case the blog above describes. I might also be curious to see benchmarks for various ratios of `(range_g % range_d)/range_g`, to address the concerns Maccoy raises. He's probably right that there are performance pessimizations in this implementation if that ratio gets too close to 1/2. But, if benchmarks show that processors can parallelize the wrong-path speculation, that would be good.

cmolder:
Okay, I pushed some tests for `champsim::uniform_int_distribution`. Next step is to test `champsim::shuffle` and expand the test suite further if we think it's necessary. I'd like to test `vmem::shuffle`, but that may be tricky.

PR #593, opened 2025-02-08
Rename bandwidth variables in handle_memory_return()

akihikodaki:
In O3_CPU::handle_memory_return(), l1i_bw represents the fetch bandwidth while to_read represents the L1I bandwidth, which is confusing.

Rename l1i_bw to fetch_bw and to_read to l1i_bw.


ngober:
This is a trivial change, but I agree that it increases the clarity of the source code.

Issue #382, opened 2023-09-04
Towards a new trace format


ngober:
The existing ChampSim trace format has a few issues that have limited development. As an initial example, the Cloudsuite traces and the standard traces have different number of destination registers, and these values are fixed, making the traces difficult to interoperate.

We ought to consider moving to a new trace format that removes the impedance of the old format and leaves the opportunity open for future growth. I'm opening this Issue to document the steps we will be taking towards making a new format. This might be a place where new requirements are added, and it might be a place where alternatives are considered.

### Desired Features

Here is a list of features that should be in the new format:

- In-built compression: the existing trace is emitted as raw values that must be immediately compressed by an external tool. The existing trace is typically too large to be stored on disk.
- File validation: the file should have a header and/or footer that validates the file contents. This can include a magic number, an instruction count, a checksum, etc.
- Source program information: the current trace format has no information about the underlying program. This information is usually encoded in the file name of the trace, but this is extremely fragile.
- Branch information in instructions: the current trace does not encode the branch type for any branch instruction or the branch target for non-taken branches. This makes branch prediction difficult.
- Memory access sizes: the size of a load, store, or a fetch is not encoded. If the access were to span multiple blocks, we would want to issue multiple memory requests.
- Well-defined format: The current format does not make strict correctness guarantees about the order in which bytes are stored on disk. If endianness or padding of struct elements varies between C++ implementations, two traces will be horribly broken between the two machines.
- Not using zero as a non-value: In both the registers and the memory accesses, the value 0 is used as an empty value. However, it may be that 0 should be a meaningful value, particularly in register values. 
- Support for multiple threads and thread interactions: the current trace format is strictly sequential and single-threaded. This makes support for multiple threads impossible since they cannot be sequenced.

### Deliverables

The deliverables on this issue are:

- A detailed file format specification, perhaps in the same format as an RFC
- A PIN tracer to replace the current tracer
- A conversion tool for the current trace format to a new format. This does not need to be an online tool (to stream the old format into ChampSim), though that could be an interesting alternative.

alberto-ros:
I would add:
 
 - The target of a branch should be present in the trace, even if the branch is not taken.

 - Optionally, more information about FP operations: e.g., a FP mul takes more cycles than a FP add and more than an integer operation.

ngober:
Ah, yes. Having opcodes in the trace would be very useful. I'll add that to the list.

aperais:
[...]

PR #594, opened 2025-02-11
Add Kanata logging

akihikodaki:
Kanata is a log format to record the behavior of processor pipelines,
which can be used with Konata, an instruction pipeline visualizer.

I'm planning to implement a feature to generate Kanata logs in a
gradual approach to minimize the burden for both me and the reviewers. This
is the first commit to add the basic infrastructure and instrument to the
instruction pipeline. Instrumenting the memory subsystem needs
significant code changes in the subsystem itself to track memory
accesses across merging and forwarding, which this commit does not
include.

However, the added Kanata logging infrastructure is already designed to
be versatile enough to cover the memory subsystem in the
future. More concretely:

- It has its own representation of "instruction"; currently it only
  logs ooo_model_instr, but it can also log prefetches in the future.
- It implements a concept of "lane", which is a unit of execution and
  part of an instruction. An instruction can have multiple lanes to
  track multiple memory accesses initiated by an instruction.
- A lane can keep its instruction alive even after the instruction
  retires from the ROB so that it can log the following stores, for example.

Please see the README.md change to know how to use this new feature.

I attached generated Kanata logs for reference: [kanata.log.gz](https://github.com/user-attachments/files/18746562/kanata.log.gz)
Below is a screenshot of Konata visualizing the attached logs:
Screenshot
The logs were generated with the following command:
```sh
bin/champsim --kanata >(gzip > kanata.log.gz) -i 100000 600.perlbench_s-210B.champsimtrace.xz
```

### DIFF
--- a/inc/event_listeners.h
+++ b/inc/event_listeners.h
@@ -8,8 +8,9 @@
 
 #include "events.h"
 #include "listeners/heartbeat.h"
+#include "listeners/kanata.h"
 
-inline auto listeners = std::make_tuple(Heartbeat(&std::cout));
+inline auto listeners = std::make_tuple(Heartbeat(&std::cout), champsim::kanata::Kanata());
 
 template <typename>
 struct listener_names_helper {
@@ -22,10 +23,19 @@ constexpr inline auto listener_names = listener_names_helper<decltype(listeners)
 
 inline std::bitset<std::tuple_size_v<decltype(listeners)>> listener_activation_map;
 
+template <std::size_t... Is>
+void event_listener_cli(std::index_sequence<Is...>, CLI::App& app)
+{
+  (std::get<Is>(listeners).cli(app), ...);
+}
+
+inline void event_listener_cli(CLI::App& app) { event_listener_cli(std::make_index_sequence<std::tuple_size_v<decltype(listeners)>>{}, app); }
+
 inline void init_event_listeners(const std::vector<std::string>& requested_listeners)
 {
   listener_activation_map.reset();
   listener_activation_map[0] = true; // heartbeat is always enabled
+  listener_activation_map[1] = true; // kanata has its own logic for enablement
   for (std::string name : requested_listeners) {
     bool found = false;
     for (size_t i = 0; i < listener_names.size(); i++) {
--- a/inc/instruction.h
+++ b/inc/instruction.h
@@ -102,6 +102,7 @@ struct ooo_model_instr : champsim::program_ordered<ooo_model_instr> {
   bool branch_taken = false;
   bool branch_prediction = false;
   bool branch_mispredicted = false; // A branch can be mispredicted even if the direction prediction is correct when the predicted target is not correct
+  bool stack_pointer_folded = false;
 
   std::array<uint8_t, 2> asid = {std::numeric_limits<uint8_t>::max(), std::numeric_limits<uint8_t>::max()};
 
@@ -197,5 +198,7 @@ public:
 
   [[nodiscard]] std::size_t num_mem_ops() const { return std::size(destination_memory) + std::size(source_memory); }
 };
+
+std::ostream& operator<<(std::ostream& os, const ooo_model_instr& instr);
 
 #endif
--- a/inc/listeners/kanata.h
+++ b/inc/listeners/kanata.h
@@ -1,0 +1,270 @@
+/*
+ *    Copyright 2025 The ChampSim Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef KANATA_H
+#define KANATA_H
+
+#include <deque>
+#include <functional>
+#include <optional>
+#include <CLI/CLI.hpp>
+#include <fmt/os.h>
+#include <fmt/ostream.h>
+
+#include "cache.h"
+#include "dram_controller.h"
+#include "events.h"
+#include "instruction.h"
+#include "ooo_cpu.h"
+#include "ptw.h"
+
+namespace champsim
+{
+namespace kanata
+{
+class Kanata final
+{
+public:
+  static constexpr auto cli_key = "Kanata";
+
+  void cli(CLI::App& app)
+  {
+    app.add_option("--kanata", file_name, "The name of the file to receive Kanata output");
+    app.add_option("--kanata-max", max, "The maximum number of instructions for Kanata logging");
+    app.add_option("--kanata-skip", skip, "The number of instructions to skip before starting Kanata logging");
+  }
+
+  template <Event e, typename... Args>
+  inline void handle_event(Args&&... args);
+
+private:
+  struct Instr final {
+    Kanata& kanata;
+    const uint64_t id;
+    const bool flushed;
+
+    Instr(Kanata& kanata_, uint64_t id_, bool flushed_) : kanata(kanata_), id(id_), flushed(flushed_) {}
+
+    ~Instr();
+    void start(uint8_t lane, const std::string& stage);
+  };
+
+  template <typename T>
+  class Queue final
+  {
+  public:
+    T& emplace(uint64_t index)
+    {
+      if (storage.empty())
+        begin = index;
+      else
+        assert(index - begin >= storage.size());
+
+      storage.resize(index - begin + 1);
+
+      return storage[index - begin];
+    }
+
+    T* find(uint64_t index) { return index >= begin && index - begin < storage.size() ? &storage[index - begin] : nullptr; }
+
+    T& front() { return storage.front(); }
+
+    bool empty() const { return storage.empty(); }
+
+    void pop()
+    {
+      storage.pop_front();
+      ++begin;
+    }
+
+  private:
+    uint64_t begin = UINT64_MAX;
+    std::deque<T> storage;
+  };
+
+  template <typename ID, typename Label>
+  std::shared_ptr<Instr> init(uint32_t cpu, ID id, Label label, bool flushed)
+  {
+    auto index = num;
+
+    ++num;
+
+    if (index < skip || index - skip > max)
+      return {};
+
+    if (!file) {
+      if (!file_name) {
+        skip = UINT64_MAX;
+        return {};
+      }
+
+      file.emplace(fmt::output_file(*file_name));
+      file->print("Kanata\t0004\nC=\t{}\n", current_cycle);
+    }
+
+    auto instr = std::make_shared<Instr>(*this, num - skip, flushed);
+    file->print("I\t{}\t{}\t{}\nL\t{}\t0\t{}\n", instr->id, id, cpu, instr->id, label);
+    return instr;
+  }
+
+  void init(const O3_CPU& cpu, const ooo_model_instr& o3)
+  {
+    auto instr = init(cpu.cpu, o3.instr_id, fmt::streamed(o3), false);
+    if (!instr)
+      return;
+
+    if (cpu.cpu >= cpus.size())
+      cpus.resize(cpu.cpu + 1);
+
+    auto& emplaced = cpus[cpu.cpu].emplace(o3.instr_id);
+    emplaced.swap(instr);
+    emplaced->start(0, "F");
+  }
+
+  void retire();
+
+  void start(const O3_CPU& cpu, const ooo_model_instr& o3, const std::string& stage)
+  {
+    if (cpu.cpu >= cpus.size())
+      return;
+
+    auto kanata = cpus[cpu.cpu].find(o3.instr_id);
+    if (!kanata)
+      return;
+
+    (*kanata)->start(0, stage);
+  }
+
+  std::optional<fmt::ostream> file;
+  std::optional<std::string> file_name;
+  std::vector<Queue<std::shared_ptr<Instr>>> cpus;
+  uint64_t current_cycle = 0;
+  uint64_t max = UINT64_MAX;
+  uint64_t num = 0;
+  uint64_t skip = 0;
+
+  template <Event e, typename... Args>
+  friend void handle_event(Kanata& kanata, const Args&... args);
+};
+
+template <Event e, typename... Args>
+inline void handle_event([[maybe_unused]] Kanata& kanata, [[maybe_unused]] const Args&... args)
+{
+}
+
+template <>
+inline void handle_event<Event::COMPLETE>(Kanata& kanata, const O3_CPU& cpu, const ooo_model_instr& instr)
+{
+  kanata.start(cpu, instr, "C");
+}
+
+template <>
+inline void handle_event<Event::CYCLE>(Kanata& kanata)
+{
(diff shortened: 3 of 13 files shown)


inc/trace_instruction.h
```diff
@@ -25,6 +25,28 @@ namespace champsim
...
+    "zmm14",   "zmm15",     "zmm16",        "zmm17",   "zmm18",   "zmm19",   "zmm20",    "zmm21",    "zmm22",    "zmm23",    "zmm24",    "zmm25",      "zmm26",
+    "zmm27",   "zmm28",     "zmm29",        "zmm30",   "zmm31",   "k0",      "k1",       "k2",       "k3",       "k4",       "k5",       "k6",         "k7",
+    "mxcsr",   "mxcsrmask", "orig_rax",     "fpcw",    "fpsw",    "fptag",   "fpip_off", "fpip_sel", "fpopcode", "fpdp_off", "fpdp_sel", "fptag_full", "st0",
+    "st1",     "st2",       "st3",          "st4",     "st5",     "st6",     "st7",      "dr0",      "dr1",      "dr2",      "dr3",      "dr4",        "dr5",
+    "dr6",     "dr7",       "cr0",          "cr1",     "cr2",     "cr3",     "cr4",      "tssr",     "ldtr",     "tr",       "tr3",      "tr4",        "tr5",
+    "tr6",     "tr7",       "status_flags", "df_flag", "gs_base", "fs_base", "g0",       "g1",       "g2",       "g3",       "g4",       "g5",         "g6",
+    "g7",      "g8",        "g9",           "g10",     "g11",     "g12",     "g13",      "g14",      "g15"};
```
ngober, line 49: I understand why you've added this, but it further highlights to me how there are many x86 assumptions in ChampSim that I would like to remove. See some discussion about generalizing the registers [here](https://github.com/ChampSim/ChampSim/issues/382#issuecomment-1706266807).

inc/instruction.h
```diff
@@ -198,4 +201,89 @@ struct ooo_model_instr : champsim::program_ordered<ooo_model_instr> {
   [[nodiscard]] std::size_t num_mem_ops() const { return std::size(destination_memory) + std::size(source_memory); }
 };
 
+inline std::ostream& operator<<(std::ostream& os, const ooo_model_instr& instr)
```
ngober, line 204: I feel like this is a lot of definition to put in a header. This should probably be in a .cpp file somewhere.

inc/register_allocator.h
```diff
@@ -25,12 +31,12 @@ class RegisterAllocator
 
 public:
   RegisterAllocator(size_t num_physical_registers);
-  PHYSICAL_REGISTER_ID rename_dest_register(int16_t reg, champsim::program_ordered<ooo_model_instr>::id_type producer_id);
-  PHYSICAL_REGISTER_ID rename_src_register(int16_t reg);
-  void complete_dest_register(PHYSICAL_REGISTER_ID physreg);
-  void retire_dest_register(PHYSICAL_REGISTER_ID physreg);
-  void free_register(PHYSICAL_REGISTER_ID physreg);
-  bool isValid(PHYSICAL_REGISTER_ID physreg) const;
+  PHYSICAL_REGISTER_ID rename_dest_register(int16_t reg, const ooo_model_instr& producer);
```
ngober, line 34: It looks like you're passing a reference to the entire instruction model here to get to its innards. I feel like this is bad design. See the discussion [here](https://github.com/ChampSim/ChampSim/pull/532#discussion_r1823564660).

ngober (commented):
This is a very interesting patch. I hadn't known that that Kanata exists. I like it and I know it could be very useful for people.

I suspect we will probably need to iterate on this a bit before we can merge it, because in my opinion this patch adds significant logging noise to the source code, and I'd like to minimize that for clarity.

You and @dpuckett98 should have a conversation. This patch has some similarities with #483, and I think they could benefit one another. Your Kanata logging could be implemented as a listener, which would provide a lot of code isolation. I have some comments on that also that need a follow-up.

inc/trace_instruction.h
akihikodaki: Indeed, it will need new code to map the register names for the new trace format discussed in #382 when the format gets implemented. This code will be used to map the register names until then.

akihikodaki:
@ngober I made the suggested changes so please check them out.
@dpuckett98 #483 may benefit my proposed change as @ngober pointed out. Do you have a plan to update your pull request?

dpuckett98:
@akihikodaki Kanata looks very helpful! I would love to see it merged with ChampSim.

I do plan to update my pull request, but I don't expect to get to it until after April since I'm working towards a deadline.

ngober:
Deadlines are very understandable. I've been clear about how my capacities have reduced lately as well.

These two patches are quite complimentary. This patch serves as excellent justification for #483, an example for why it's extremely useful. As such, I think I want to prioritize this patch after #483, for a few reasons:
 1. Merging this patch first will create substantial merge conflicts that will delay #483, since they are hitting the very same spots in the code (pipeline stage triggers, cache hit/miss, etc.).
 2. The needs of this patch may help inform the design of #483, as you mentioned there may be refactoring necessary in the cache model.
 3. This patch decreases test coverage by 1.7%. Refactoring this patch out into an event listener will enable better testing and actually improve test coverage.

So, I don't want to shut this patch down. I would absolutely love to see it merged. But, given the scale of this patch (cache activity is still to come), and the benefits of rebasing on #483, I intend to let this one sit a while before merging. @akihikodaki, if you would like to coordinate with @dpuckett98 to contribute to #483, that would probably be beneficial.

akihikodaki:
For anyone interested in Kanata logging, I pushed a tree that instruments the memory subsystem on top of this pull request and #618: https://github.com/akihikodaki/ChampSim/tree/tyche2
A sample Kanata log file generated with memory subsystem instrumentation is attached with this comment: [400.perlbench-41B.log.gz](https://github.com/user-attachments/files/19998308/400.perlbench-41B.log.gz)

akihikodaki:
Rebased and converted the implementation to an event listener.

inc/event_listeners.h
```diff
@@ -22,10 +23,19 @@ constexpr inline auto listener_names = listener_names_helper<decltype(listeners)
...
 inline std::bitset<std::tuple_size_v<decltype(listeners)>> listener_activation_map;
 
+template <std::size_t... Is>
+void event_listener_cli(std::index_sequence<Is...>, CLI::App& app)
+{
+  (std::get<Is>(listeners).cli(app), ...);
+}
+
+inline void event_listener_cli(CLI::App& app) { event_listener_cli(std::make_index_sequence<std::tuple_size_v<decltype(listeners)>>{}, app); }
+
```
ngober, line 33: I like this idea. It's good design to give the responsibility for command line options to the individual listeners. However, I'm concerned that passing the CLI11 object to the listeners is a bad API. It leaks implementation details. (Even though I've been quite happy with the choice of CLI11, we should leave the opportunity open to transition away from it.)

I would suggest a pull-style model, rather than a push-style.

```cpp
struct MyListener {
  static std::string name() { return "mylistener"; }
  static std::array<std::tuple<std::string, int&, std::string>, 2> args() {
    return {
      {"max", MyListener::max, "the maximum widgets"},
      {"skip", MyListener::skip, "the skippy widgets"}
    };
  }

  static inline bool enable = false;
  static inline int max = 0;
  static inline int skip = 100000;
};

/* main.cpp */
void listener_options(CLI::App& app) {
  for (const auto& listener : listeners) { // I know this isn't how this works. Bear with me here...
    app.add_option("--"+listener::name(), listener::enable, "Enable listener "+listener::name());
    for (const auto& opt : listener::args()) {
      app.add_option(fmt::format("--{:s}.{:s}", listener::name(), std::get<0>(opt)), std::get<1>(opt), std::get<2>(opt));
    }
  }
}
```

I know there would probably be some implicit dependency between the listeners and the CLI, but I hope there's a way to do this that doesn't `#include <CLI/CLI.hpp>` in the listener sources.

inc/instruction.h
```diff
@@ -198,4 +199,89 @@ struct ooo_model_instr : champsim::program_ordered<ooo_model_instr> {
...
+{
+  fmt::print(os, "{}: ", instr.ip);
```
ngober:
```suggestion
  fmt::print(os, "{:x}: ", instr.ip);
```
I suspect this will be more informative

inc/instruction.h
ngober: Unresolving this, because I insist that a better place to put this is in a source file.

Also, I have two critiques here:
 1. You have a combination of `fmt::print(os, ...` and `os << ...` going on. Try to bring the two together into one idea.
 2. I find the `bool needs_comma` somewhat yucky. It's saying how to do the thing, rather than what the thing is you're doing. I would try accumulating a `std::vector<std::string> string_parts` then doing `fmt::join(string_parts, ", ")`.

inc/listeners/kanata.h
```diff
@@ -0,0 +1,270 @@
...
+      cpus.resize(cpu.cpu + 1);
+
+    auto& emplaced = cpus[cpu.cpu].emplace(o3.instr_id);
+    emplaced.swap(instr);
+    emplaced->start(0, "F");
+  }
+
+  void retire();
+
+  void start(const O3_CPU& cpu, const ooo_model_instr& o3, const std::string& stage)
```
ngober, line 139: Passing `const O3_CPU&` as a parameter here is not a very good interface. You're exposing the entire core model to the listener. You should be more selective about which pieces of information you're passing here. It seems like you're only using the cpu index right now, so maybe stick to that?

inc/listeners/kanata.h
```diff
@@ -0,0 +1,270 @@
...
+  {
+    if (cpu.cpu >= cpus.size())
+      return;
```
ngober, line 142: Is "return" the proper behavior here? This seems like a bug if it's true.

inc/listeners/kanata.h
```diff
@@ -0,0 +1,270 @@
...
+    const bool flushed;
+
+    Instr(Kanata& kanata_, uint64_t id_, bool flushed_) : kanata(kanata_), id(id_), flushed(flushed_) {}
+
+    ~Instr();
+    void start(uint8_t lane, const std::string& stage);
+  };
+
+  template <typename T>
+  class Queue final
```
ngober, line 66: What is the purpose of this class? It seems to be a circular queue, but your underlying storage is a `std::deque`, which is stable but isn't contiguous. I suppose you want `std::vector` for the storage, but in general, why would `std::queue` not suffice?

ngober (changes requested):
This looks great. The changes to `src/ooo_cpu.cc` are exactly what I envisioned: minimally invasive and unified with other events. I'm pleased that it worked out that way. I have some design comments that I think need to be worked through. This patch now particularly raises the concern about CLI options. Now that we have event listeners, we need to thing through the CLI design, or we'll end up with unmaintainable code down the line.

inc/event_listeners.h
akihikodaki: I agree it leaks implementation details, but this is the result of my attempt to balance the interface stability and the separation of concerns.

What you suggested provides better interface stability by encapsulating CLI11 to `main.cpp` and makes it easier to change the implementation of `main.cpp` like migrating away from CLI11. However, it also puts more assumptions on how arguments are interpreted, which is a concern of listeners.

For example, some arguments of Kanata is interpreted as `std::optional<std::string>` instead of an integer, which cannot be handled with the code you provided. Adding an interface for the type will solve the issue, but it is an ad-hoc approach that leaks the concern of listeners. Making the interface generic will results in a duplication of CLI11.

Directly exposing CLI11 will make transition away from it relatively harder as you pointed out, but my intuition is that it will be still tractable. In most cases, options will be defined with simple `add_option` that is easy to convert. There may be more complex options, but I expect they will be exceptions and do not hinder transition, hence there is no need to prohibit them.

inc/instruction.h
akihikodaki: `instr.ip` is `champsim::address`, which already has a nice formatter.

inc/listeners/kanata.h
akihikodaki: `std::queue` does not provide `operator[]()`. `std::vector` does not provide `pop_front()`.

`std::deque` may not be contiguous and may perform worse than `std::vector` for some operations, but all operations we need are O(1) and that should be sufficient; the Kanata listener is I/O-intensive and the computational cost is tiny compared to the I/O cost, as long as we don't use an operations like O(n*2).

inc/listeners/kanata.h
akihikodaki: Yes, it is intentional. It means the instructions in the CPU are not tracked by the Kanata listener because they are initialized before the listener gets triggered.

inc/listeners/kanata.h
akihikodaki: This is another design question that needs careful balancing. The problem with selecting information to pass to listeners is that listeners may have different concerns and need different information.

If you look at this particular example, passing the CPU index may look just fine, but in the upcoming change to log memory accesses[1] add event handlers that need more information. A question arises when having such a handler: what information will event listeners need?

I don't think there is a universal answer to this question. When we have more event listeners, they will have different requirements, and event definitions can grow significantly if we add a parameter for each. Instead of answering the question, I apply the idea of inversion of control here. Instead requiring the event emitters to know what data event listeners will need, I let the listeners choose what they want.

However, it must be noted that listeners should still access information that is *safe* to do so. Generally speaking, in OOP, objects should encapsulate states to maintain its invariants and to avoid accidental dependencies to internal states. C++ enables this by having the concept of accessiblity (e.g., `private`) and `const`, which can also be applied to objects passed to event listeners if wanted.

[1] https://github.com/akihikodaki/ChampSim/commit/035182f991e5ccb08cc387089130b75ba72d6b7a#diff-662ee17cb2e3488e280dafd2901fb2789007bbda3a4d3eff77601e0fcbb95d23

akihikodaki:
Commit ff7340ac93530ccde440c0856b99ce6840cb5d2d addresses https://github.com/ChampSim/ChampSim/pull/594#discussion_r2676621745.

inc/instruction.h
ngober: I should know this, I wrote it... Don't mind me.

inc/listeners/kanata.h
ngober: I think we agree on philosophy, and we need to work out how to build this out properly. I hear your concern that the number of parameters will scale out of control. You are concerned about an interface like

```cpp
void interface(int a, int b, int c /* and many others */) { /*...*/ }
```

and I am concerned about an interface like

```cpp
void interface(const O3_CPU& cpu, const CACHE& cache, const MEMORY_CONTROLLER& mem) {
    int a = cpu.get_a();
    int b = cache.get_b();
    int c = mem.get_c();
    /* and many others */
    /* ... */
}
```

I think there is a difference between the two. The second exposes the entire state of the simulator to each event, and trusts the client not to use it improperly. For example, I might suggest that the listener should itself count the number of prefetch hits in a `CACHE_HIT` event, and record later in the `RETIRE` event, rather than exposing the cache model to the `RETIRE` event. It's an imperfect example, but I mean to say that we should limit the information available at an event to be the information pertinent to that event. I don't know where that line lies, but I'm open to iterating on the interface for individual events.

akihikodaki: We do not have to trust event listeners for proper usage. The point is, C++ objects can enforce (and should have enforced) its proper usage by accessibility and `const`. In your example, `CACHE` should not make the number of prefetches `public`.

Exposing a model object to events do not prevent having different interfaces for individual events. For example, events can have parameters to pass information valid only at the time they are triggered. On the other hand, information that can be accessed while a simulator model is alive is better to be exposed via the simulator model itself; by doing so, we can have one common interface definition for such information instead of having 1 + N definitions where 1 is for the accessibility the class defines and N is for each event.

PR #595, opened 2025-02-15
Update champsim.cc to extend livelock period from 100K cycles to 10M cycles

maccoymerrell:
The current livelock duration is too short. This can result in significant stall-drain behavior in some traces triggering the livelock. As a resolution, we can extend the livelock period which should help prevent these false positives at the expense of potentially spending longer livelocked. Champsim can chew through 10M cycles like nothing, so I don't think this will be an issue.

This is causing problems for ECEN 676 at the university, so I would appreciate this PR having high prio.

### DIFF
--- a/src/champsim.cc
+++ b/src/champsim.cc
@@ -75,7 +75,7 @@ phase_stats do_phase(const phase_info& phase, environment& env, std::vector<trac
                                             [](const auto acc, const operable& y) { return std::min(acc, y.clock_period); });
 
   bool livelock_trigger{false};
-  uint64_t livelock_period{100000};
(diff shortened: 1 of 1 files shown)


ngober:
I don't see a problem with this, since it isn't a change in function. It goes from dying if the simulation fails to retire 1 kinst in 100 kcycles to dying if it fails to retire 10 kinst in 1000 kcycles. That is, the ratio is the same. 

I will merge this, but I would like to consider for the future two alternatives:
 1. Add a command-line option to configure the livelock duration
 2. Create your own fork and branch for use at the university. This would expedite the process for getting changes through.

PR #600, opened 2025-02-21
PTW PSCL Fix

maccoymerrell:
Fixes a bug that causes PTW walks that were previously cached to begin at 1 level lower than expected.

I am still not entirely sure whether this is a bug or not.
Since VMEM has to allocate more pages for the page tables after this fix, I would assume this was indeed a bug. If it wasn't, those page table pages would have been allocated originally.
However, this also means that the top-level of the PSCL doesn't seem to reduce the number of translations required when it hits.

Essentially what is happening is that each level of the PSCL should cache entries from its respective level of the page table hierarchy.
Level 5 (top) should cache entries in the level 5 table, level 4 should cache level 4 entries, and so on.
We don't have a level 1 table, since this is the actual translation.

When we hit in the PSCLs, we should be able to start our walk at the level just below it. So hitting in the level 5 PSCL should allow the walk to begin at level 4 using the entry in the level 5 pscl as the address.  What seems to be happening is something strange with the level indexing. Somehow, we are hitting in the pscl at a level that we shouldn't. It matches the upper bits, so either the shifting is wrong for each level, or we need to begin our walk 1 higher, I can't tell.

I have modified test 603 to account for this change, and do some more thorough testing as well.

### DIFF
--- a/src/ptw.cc
+++ b/src/ptw.cc
@@ -90,7 +90,7 @@ auto PageTableWalker::handle_fill(const mshr_type& fill_mshr) -> std::optional<m
   }
 
   const auto pscl_idx = std::size(pscl) - fill_mshr.translation_level;
-  pscl.at(pscl_idx).fill({fill_mshr.v_address, *fill_mshr.data, fill_mshr.translation_level - 1});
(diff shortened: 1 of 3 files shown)


ngober (approved):
This looks good to me. I like the additions to test 603. This version is more rigorous. I've tinkered with this a couple times, it's been a bit frustrating to get right.

PR #619, opened 2025-04-30
Check dependencies of the main file

akihikodaki:
The dependencies of the main file in a build not for testing was not checked, which could result in an invalid build when one of the dependencies like inc/channel.h was modified. Check them properly by including dependency rules of the build specified in _configuration.mk.

### DIFF
--- a/config/makefile.py
+++ b/config/makefile.py
@@ -84,6 +84,7 @@ def get_makefile_lines(build_id, executable, module_info):
     if legacy_paths:
         yield from append_variable('prereq_for_generated', *legacy_paths, targets=['$(generated_files)'])
 
+    yield from append_variable('build_ids', build_id)
     yield from append_variable('executable_name', exe_basename)
 
     yield ''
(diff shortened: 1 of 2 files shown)


config/makefile.py
```diff
@@ -75,7 +75,7 @@ def get_makefile_lines(build_id, executable, module_info):
     exe_dirname, exe_basename = os.path.split(os.path.normpath(executable))
     exe_basename = os.path.join('$(BIN_ROOT)', exe_basename)
     yield from hard_assign_variable('BIN_ROOT', exe_dirname)
-    yield from hard_assign_variable('build_id', build_id, targets=["compile_commands", exe_basename])
+    yield from hard_assign_variable('build_id', build_id)
```
ngober, line 78: Does this break if a configuration would build multiple executables? For example, with the configuration

```json
[
  { "name": "small", "rob_size": 200 },
  { "name": "large", "rob_size": 500 }
]
```

Then, running

    ./config.sh <file>
    make

Does this build two executables or one? I think if you hard-assign a variable without the condition of the targets, the two executables will get the same build ID.

akihikodaki: Indeed, the case with multiple executables was not properly handled. I fixed the issue with commit 07460f325291df9f3c716df58488d758c132ec22 and tested it with `test/config/compile-only/multi-configuration.json`.

ngober (approved):
Great fix!

PR #620, opened 2025-04-30
Stop testing on Ubuntu 20.04 on GitHub Actions

akihikodaki:
Ubuntu 20.04 on GitHub Actions became fully unsupported by 2025-04-15.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,22 +38,12 @@ jobs:
         config:
         # GCC-9
         - {
-            name: "Ubuntu 20.04 GCC 9",
-            os: ubuntu-20.04,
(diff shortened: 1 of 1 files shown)


ngober:
Would you add `ubuntu-24.04` to the runner list? We might as well keep this up to date.

ngober:
It looks like now the GCov version is mismatched with the GCC version. You might need to update it with `apt` in the setup step.

akihikodaki:
I changed to measure the coverage using GCC 13 with commit c619c085bb00edab6c8cd0a53dbe6bef49509814, assuming that the default compiler is the most reliable.
Commit f3d171b63b412941342305342cc15a04f9f67361 resolves another LCov failure.

PR #624, opened 2025-05-01
Branch predictor champ sim

yardboy27:


yardboy27 (commented):
should be fine

yardboy27:
this was a mistake

PR #663, opened 2025-11-02
Adding bzip2 compression to the ChampSim tracer

joshualmashburn:
I'm opening a PR for visibility and feedback.

Someone asked why we use pipes to compress the tracer output and I remembered I had this mod lying around. Bzip2 is not the most performant algorithm, but it cuts down on IPC overhead and it's a happy medium compression-wise between gz and lzma.

I did have to replace the C++ file I/O API usage with C stdio for compatibility with libbzip2.
Originally I added libbzip2 as a submodule, but I rewrote the README instructions to use the vcpkg package source cache.

I envision further mods in this vein, like "-g" for gzip, "-x" for xz/lzma.

### DIFF
--- a/tracer/pin/champsim_tracer.cpp
+++ b/tracer/pin/champsim_tracer.cpp
@@ -19,12 +19,14 @@
  *  and could serve as the starting point for developing your first PIN tool
  */
 
-#include <fstream>
+//#include <fstream>
+#include <cstdio>
 #include <iostream>
 #include <stdlib.h>
 #include <string.h>
 #include <string>
 
+#include "bzip2/bzlib.h"
 #include "../../inc/trace_instruction.h"
 #include "pin.H"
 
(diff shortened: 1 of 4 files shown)


tracer/pin/champsim_tracer.cpp
```diff
@@ -87,9 +98,22 @@ BOOL ShouldWrite()
 
 void WriteCurrentInstruction()
 {
-  typename decltype(outfile)::char_type buf[sizeof(trace_instr_format_t)];
   std::memcpy(buf, &curr_instr, sizeof(trace_instr_format_t));
-  outfile.write(buf, sizeof(trace_instr_format_t));
+  fwrite(buf, 1, sizeof(trace_instr_format_t), outfile);
+}
+
+void WriteCurrentInstruction_bz2()
```
ngober, line 105: Is it going to be more efficient to buffer multiple instructions before performing a bz write? When we're just dumping to file it might be fine, but maybe it's different with the compression library?

ngober (approved):
I could be fussy about C++ style here, but PIN is weird, and it's probably not worth it. I wanted to add this in such a long time ago, but never did. Thanks for the patch.

tracer/pin/champsim_tracer.cpp
joshualmashburn: You know, it probably would. When I get some free time (like Thanksgiving-ish, after DAC submissions), I can time some runs of varying buffer sizes.

joshualmashburn:
I'm definitely willing to touch it up if you have feedback on the coding style. And I know the build instructions, at least the monolith of a build command for bzip, could possibly be moved to the makefile.

PR #665, opened 2025-11-05
Fix PT page allocation

gh1403:
Currently each `VirtualMemory::get_pte_pa` function call would allocate a new PT page, as `champsim::page_offset{next_pte_page} == champsim::page_offset{0}` is always true. 

Also, PT page insertion logic seems to use the wrong level for index comparison. Consider level = 1, the function will try to insert the full PFN into `page_table`. While in `VirtualMemory::va_to_pa` it will do another page address translation. Therefore, the fault penalty and the overall memory usage will be doubled.

### DIFF
--- a/src/vmem.cc
+++ b/src/vmem.cc
@@ -124,18 +124,17 @@ std::pair<champsim::page_number, champsim::chrono::clock::duration> VirtualMemor
 
 std::pair<champsim::address, champsim::chrono::clock::duration> VirtualMemory::get_pte_pa(uint32_t cpu_num, champsim::page_number vaddr, std::size_t level)
 {
-  if (champsim::page_offset{next_pte_page} == champsim::page_offset{0}) {
-    active_pte_page = ppage_front();
-    ppage_pop();
-  }
-
-  champsim::dynamic_extent pte_table_entry_extent{champsim::address::bits, shamt(level)};
+  champsim::dynamic_extent pte_table_entry_extent{champsim::address::bits, shamt(level + 1)};
   auto [ppage, fault] =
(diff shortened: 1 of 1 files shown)


maccoymerrell:
I think that the PT page insertion logic is correct (from our work with R-MAX, we rebuilt the page table structure from tracing and found it was correctly done). The PTW is providing the level index for the get_pte_pa calls.
If you believe this is an error, could you generate some printouts showing that we are triggering additional page faults / allocating additional pages without this change?

gh1403:
please use my test branch [here](https://github.com/gh1403/ChampSim/tree/check-pt) and run random workloads like:

`bin/champsim --warmup-instructions 0 --simulation-instructions 2500000 ../traces/dpc3/436.cactusADM-1804B.champsimtrace`

The output logs would have unrealistically high numbers of 'new active pte page', and the count of 'new next pte page' would also be more than 'new data page'. For example, 436.cactusADM-1804B workload has about only 4MB memory footprint but allocates thousands of pte pages (should only need tens of them).

maccoymerrell:
@ngober This seems like a major issue. We are overallocating a significant number of pages for pte entries. Like mentioned above, translation is going to be much slower as well.

maccoymerrell:
This probably needs a test added to the test suite.

Also, from some initial runs, this could account for as much as a 5% performance differencd for some workloads. Others may be closer to 1%. Drops off as the TLB is filled and ptw walks become less frequent.

ngober:
Oh, yea, that does look like a bug. That will over-allocate pages especially when warming up. Let's add a test on this and get it merged.

ngober:
Resolved with #688

PR #666, opened 2025-11-06
changes to vmem physical page allocation policy

eshaanmandal:
This PR updates the virtual-to-physical page allocation policy to use a hash-based index instead of sequential allocation. This makes physical page selection deterministic and reproducible across runs, addressing variability discussed in [#656](https://github.com/ChampSim/ChampSim/discussions/656)

Sequential allocation caused differences in mappings depending on timing/order of page faults, leading to inconsistent behavior even with identical seeds. A hash-based mapping stabilizes this.

### DIFF
--- a/src/vmem.cc
+++ b/src/vmem.cc
@@ -50,14 +50,7 @@ VirtualMemory::VirtualMemory(champsim::data::bytes page_table_page_size, std::si
   std::mt19937_64 rng(randomization_seed.value_or(0));
   std::uniform_int_distribution<uint64_t> dist(0, ppage_free_list.size() - 1);
 
-  // std::vector<uint64_t> hash_constants;
-  hash_constants.resize(NUM_CPUS);
-
-  
-
-  for(std::size_t i=0; i < hash_constants.size(); i++){
-    hash_constants[i] = dist(rng);
-  }
+  std::generate_n(std::back_inserter(hash_constants), NUM_CPUS, [rng, dist]{ return dist(rng); });
 
 }
 
--- a/src/vmem.cc
+++ b/src/vmem.cc
@@ -46,6 +46,19 @@ VirtualMemory::VirtualMemory(champsim::data::bytes page_table_page_size, std::si
   }
   populate_pages();
   shuffle_pages();
+
+  std::mt19937_64 rng(randomization_seed.value_or(0));
+  std::uniform_int_distribution<uint64_t> dist(0, ppage_free_list.size() - 1);
+
+  // std::vector<uint64_t> hash_constants;
+  hash_constants.resize(NUM_CPUS);
+
+  
+
+  for(std::size_t i=0; i < hash_constants.size(); i++){
+    hash_constants[i] = dist(rng);
+  }
+
 }
 
 VirtualMemory::VirtualMemory(champsim::data::bytes page_table_page_size, std::size_t page_table_levels, champsim::chrono::clock::duration minor_penalty,
@@ -53,6 +66,19 @@ VirtualMemory::VirtualMemory(champsim::data::bytes page_table_page_size, std::si
     : VirtualMemory(page_table_page_size, page_table_levels, minor_penalty, dram_, {})
 {
 }
+
+// void VirtualMemory::populate_pages()
+// {
+//   assert(dram.size() > 1_MiB);
+//   ppage_free_list.resize(((dram.size() - 1_MiB) / PAGE_SIZE).count());
+//   assert(ppage_free_list.size() != 0);
+//   champsim::page_number base_address =
+//       champsim::page_number{champsim::lowest_address_for_size(std::max<champsim::data::mebibytes>(champsim::data::bytes{PAGE_SIZE}, 1_MiB))};
+//   for (auto it = ppage_free_list.begin(); it != ppage_free_list.end(); it++) {
+//     *it = base_address;
+//     base_address++;
+//   }
+// }
 
 void VirtualMemory::populate_pages()
 {
@@ -62,7 +88,8 @@ void VirtualMemory::populate_pages()
   champsim::page_number base_address =
       champsim::page_number{champsim::lowest_address_for_size(std::max<champsim::data::mebibytes>(champsim::data::bytes{PAGE_SIZE}, 1_MiB))};
   for (auto it = ppage_free_list.begin(); it != ppage_free_list.end(); it++) {
-    *it = base_address;
+    it->first = base_address;
+    it->second = false;
     base_address++;
   }
 }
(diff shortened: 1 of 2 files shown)


src/vmem.cc
```diff
@@ -54,6 +67,19 @@ VirtualMemory::VirtualMemory(champsim::data::bytes page_table_page_size, std::si
...
+//   ppage_free_list.resize(((dram.size() - 1_MiB) / PAGE_SIZE).count());
+//   assert(ppage_free_list.size() != 0);
+//   champsim::page_number base_address =
+//       champsim::page_number{champsim::lowest_address_for_size(std::max<champsim::data::mebibytes>(champsim::data::bytes{PAGE_SIZE}, 1_MiB))};
+//   for (auto it = ppage_free_list.begin(); it != ppage_free_list.end(); it++) {
+//     *it = base_address;
+//     base_address++;
+//   }
+// }
+
```
ngober: Please remove all of the commented code here. The source code is under version control, so we can easily go back and see old behavior.

src/vmem.cc
```diff
@@ -86,10 +113,52 @@ uint64_t VirtualMemory::get_offset(champsim::address vaddr, std::size_t level) c
...
 
-champsim::page_number VirtualMemory::ppage_front() const
+// champsim::page_number VirtualMemory::ppage_front() const
+// {
+//   assert(available_ppages() > 0);
+//   return ppage_free_list.front();
+// }
+
+
+std::deque<std::pair<champsim::page_number, bool>>::iterator
```
ngober: This should just be `auto`. There's no point in spelling it out.

src/vmem.cc
```diff
@@ -86,10 +113,52 @@ uint64_t VirtualMemory::get_offset(champsim::address vaddr, std::size_t level) c
...
+    auto vaddr_val = vaddr.to<uint64_t>();
+    std::size_t idx = (hash_constants[cpu_num] ^ vaddr_val) % ppage_free_list.size();
+
+    auto it = ppage_free_list.begin() + idx;
+    
+    // for keeping track of the original ptr
+    auto start_itr = it;
+    while(it != ppage_free_list.end() && it->second){
+      it++;
+    }
```
ngober, line 111:
```suggestion
    auto start_itr = ppage_free_list.begin() + idx;
    
    auto it = std::find_if(start_itr, ppage_free_list.end(), [](const auto& free_list_entry) { return !free_list_entry.second; });
```

There's a standard library function for what you're doing here.

src/vmem.cc
```diff
@@ -46,6 +46,19 @@ VirtualMemory::VirtualMemory(champsim::data::bytes page_table_page_size, std::si
...
+  std::uniform_int_distribution<uint64_t> dist(0, ppage_free_list.size() - 1);
+
+  // std::vector<uint64_t> hash_constants;
+  hash_constants.resize(NUM_CPUS);
+
+  
+
+  for(std::size_t i=0; i < hash_constants.size(); i++){
+    hash_constants[i] = dist(rng);
+  }
```
ngober:
```suggestion
  std::generate_n(std::back_inserter(hash_constants), NUM_CPUS, [rng, dist]{ return dist(rng); });
```
I think this can be expressed more succinctly with an algorithm call.

src/vmem.cc
```diff
@@ -86,10 +113,52 @@ uint64_t VirtualMemory::get_offset(champsim::address vaddr, std::size_t level) c
...
+
+void VirtualMemory::ppage_pop_idx(std::deque<std::pair<champsim::page_number, bool>>::iterator it){
+  it->second = true;
+  // ppage_free_list.erase(it);
+  // if (available_ppages() == 0) {
+  //   fmt::print("[VMEM] WARNING: Out of physical memory, freeing ppages\n");
+  //   populate_pages();
+  //   shuffle_pages();
+  // }
 }
```
ngober: This probably doesn't need to be a function, you can just do `it->second = true` at the call site.

ngober (changes requested):
This looks like it will do what the issue suggests. I have some code cleanliness comments.

src/vmem.cc
eshaanmandal: I will apply the changes, also I had one more thing to report. So when doing make test, one of the test fails. The test started with the number 802. The reason being that the original algo removed the pages from the free_pa_list, and this algo doesn't.  And the test checks for the fact if x pages have been removed from the free list after allocation. So, if we go ahead with this change, do we need to modify the test as 
well?


`test/cpp/src/801-vmem-duplicated.cc:51: FAILED:
  REQUIRE( original_size - 2 == uut.available_ppages() )
with expansion:
  130814 (0x1fefe) == 130816 (0x1ff00)
`

eshaanmandal: Also could i keep the return type the way it is rather than "auto", as it makes it easier to understand what data type is being returned?

ngober: The return type is a contract with the calling code. If you say "this function returns a `std::deque<std::pair<champsim::page_number, bool>>::iterator`", then the calling code can depend on any of the following:

 - The container is a `std::deque`, and not a `std::vector`.
 - The value type is a `std::pair` and not a structural type with two members
 - The pair contains a `champsim::page_number` and not a `champsim::address`.
 - The return type gives a view into an underlying range structure, that is, that I can increment it and maybe decrement it to move around in the range.

None of that is what the function is meant to do. What the function actually does is return a handle to a value that can be popped. Maybe it should just return a `std::pair<champsim::page_number, bool>&`.

ngober: > I will apply the changes, also I had one more thing to report. So when doing make test, one of the test fails. The test started with the number 802. The reason being that the original algo removed the pages from the free_pa_list, and this algo doesn't. And the test checks for the fact if x pages have been removed from the free list after allocation. So, if we go ahead with this change, do we need to modify the test as well?
> 
> `test/cpp/src/801-vmem-duplicated.cc:51: FAILED: REQUIRE( original_size - 2 == uut.available_ppages() ) with expansion: 130814 (0x1fefe) == 130816 (0x1ff00) `

All of the tests should pass before we can merge the changes. If you make changes to the tests, we should all understand why these changes reflect more correct behavior.

eshaanmandal: That makes sense, i did change it to auto.  Regarding the test, all the other tests pass except one (the one for vmem). Since the list size remains fixed. Should I update the ppages_available() function so that it returns how many unoccupied locations are there in the dequeue?

ngober: Yes. You have changed how the underlying structures work. You will need to make sure that `ppages_available()` reports the number of available pages remain. That's what the tests are there to catch.

maccoymerrell: You could just use a counter to track pages that have been allocated so far. Initialize to total free pages, then decrement when a page is allocated. Report that from those functions.

src/vmem.cc
eshaanmandal: done

PR #667, opened 2025-11-16
Fix fwcounter bug

maccoymerrell:


maccoymerrell:
This fixes a recently discovered bug in inc/msl/fwcounter.
This resulted in the decremented operator (--) incrementing instead.

There were no tests for this operator, so those have been added as well.

ngober (approved):
Somebody had emailed me about this a couple months ago, and I told them "Nice find, please submit a patch." and, well, thanks for the patch!

seemakurthi-anand:
> This fixes a recently discovered bug in inc/msl/fwcounter. This resulted in the decremented operator (--) incrementing instead.
> 
> There were no tests for this operator, so those have been added as well.

This fix is really important

PR #670, opened 2025-11-28
Trace by symbol instead of instruction count.

Easyoakland:
# Feature
Adds two new command line options to the pin tracer:  `start_symbol` and `stop_symbol`.

These allow precisely controlling when tracing occurs. When the start_symbol is reached tracing is enabled, and when the stop_symbol is reached tracing is disabled. This can be used on existing binaries (possibly by using `objdump -d -j .text` or similar to find the symbol) or by purposefully creating an unmangled function and inserting it into the binary which is intended on being traced.

If the new arguments are unspecified, behavior is the same as before the feature except that:

This pr also changes the tracer so that after the number of instructions specified by `-t` have been traced, the pin tool terminates the traced program, since there is no reason to continue execution.

# Misc
From what I can see prs are expected to pull request into the `develop` branch. Let me know if I'm mistaken.
I wrote this because I found it useful. If inclusion is unwanted, let me know and I'll close the pr.

### DIFF
--- a/tracer/pin/champsim_tracer.cpp
+++ b/tracer/pin/champsim_tracer.cpp
@@ -82,7 +82,13 @@ void ResetCurrentInstruction(VOID* ip)
 BOOL ShouldWrite()
 {
   ++instrCount;
-  return (instrCount > KnobSkipInstructions.Value()) && (instrCount <= (KnobTraceInstructions.Value() + KnobSkipInstructions.Value()));
+  if (instrCount > (KnobTraceInstructions.Value() + KnobSkipInstructions.Value())) {
+    // Stop the application with code 0.
+    // Note we don't use `PIN_Detach()` because that would stop instrumentation, but let the process continue, and there's no reason to do that.
+    PIN_ExitApplication(0);
+    return false;
+  }
+  return (instrCount > KnobSkipInstructions.Value());
 }
 
 void WriteCurrentInstruction()
--- a/tracer/pin/champsim_tracer.cpp
+++ b/tracer/pin/champsim_tracer.cpp
@@ -157,11 +157,6 @@ VOID StopTracing(VOID* ip)
 // Instrument image loads to find symbols
 VOID ImageLoad(IMG img, VOID* v)
 {
-  /* // List all symbols in the image to debug
-  for (SYM sym = IMG_RegsymHead(img); SYM_Valid(sym); sym = SYM_Next(sym)) {
-    std::cout << "Found symbol: " << SYM_Name(sym) << " at " << std::hex << SYM_Address(sym) << std::dec << std::endl;
-  } */
-
   // Look for start symbol
   if (!KnobStartSymbol.Value().empty() && !foundStartSymbol.load()) {
     RTN rtn = RTN_FindByName(img, KnobStartSymbol.Value().c_str());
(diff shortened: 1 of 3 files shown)


tracer/pin/champsim_tracer.cpp
```diff
@@ -106,6 +137,77 @@ void WriteToSet(T* begin, T* end, UINT32 r)
...
+  }
+}
+
+// Instrument image loads to find symbols
+VOID ImageLoad(IMG img, VOID* v)
+{
+  /* // List all symbols in the image to debug
+  for (SYM sym = IMG_RegsymHead(img); SYM_Valid(sym); sym = SYM_Next(sym)) {
+    std::cout << "Found symbol: " << SYM_Name(sym) << " at " << std::hex << SYM_Address(sym) << std::dec << std::endl;
+  } */
```
ngober: You should remove these debugging comments before merge.

ngober (approved):
I like this change, I anticipate it will be very useful for people writing benchmarks. Thanks for the patch!

tracer/pin/champsim_tracer.cpp
Easyoakland: Done.

Easyoakland:
Oh by the way, this should close https://github.com/ChampSim/ChampSim/issues/455

ngober:
@joshualmashburn Have you looked any further into #455? Were you able to learn anything about it?

joshualmashburn:
I am unsure if this closes that issue. See my new comments in #455 . We should test this version with a nonzero number to skip, but from what I see, that particular mechanism hasn't changed, so I'm somewhat doubtful.

ngober:
If this doesn't actually solve #455 as-is, I would rather fix it in a separate patch, rather than trying to work it into this patch. If they can be independent, let's let them be independent.

joshualmashburn:
Separate PR for #455 inbound. Upon a quick skim, it looks orthogonal enough.

ngober:
As I suspected, the two patches have a merge conflict. @Easyoakland if you can resolve this, we'll get this patch merged.

Easyoakland:
Why's that PR make the global volatile? It should be atomic like this PR made it. It's not a machine register for performing I/O.

Easyoakland:
> I am unsure if this closes that issue. See my new comments in #455 . We should test this version with a nonzero number to skip, but from what I see, that particular mechanism hasn't changed, so I'm somewhat doubtful.

@joshualmashburn, did you actually test this PR? It stopped tracing for me exactly as expected. (Yes, this includes if you specify `-s` with a nonzero value)

joshualmashburn:
@Easyoakland I didn't imply that I did; I am traveling currently. I just tested it and yes it works. Good work. Keep the skip and trace instruction counts global to reduce those function call overheads, but go ahead with your changes, including making instruction count atomic.

Edit: I haven't thoroughly tested the symbol functionality. Boarding a flight now. I'll test it more in a moment.

joshualmashburn:
@Easyoakland could you please document the new feature in the README as well? Please give an example usage of the new options.

Easyoakland:
@joshualmashburn 

> Keep the skip and trace instruction counts global to reduce those function call overheads

This doesn't make any sense. The implementation for `Knob.Value()` is just `const TYPE& Value() const { return _value; }`. The compiler should easily inline this. There's no reason to make our own copy of an already cached scalar value.

> @Easyoakland could you please document the new feature in the README as well? Please give an example usage of the new options.

I've added an example and updated the README.

@ngober
> ... if you can resolve this, we'll get this patch merged.

Merge conflict resolved.

ngober (approved):

PR #674, opened 2025-12-04
Fix for pin tracer instruction count bug

joshualmashburn:
Fixes issue #455 . I made the numbers of instructions to trace and skip into globals, filled in main, instead of in the instrumentation. I also made the instruction count volatile. It appears to have rectified the issue.

### DIFF
--- a/tracer/pin/champsim_tracer.cpp
+++ b/tracer/pin/champsim_tracer.cpp
@@ -34,7 +34,9 @@ using trace_instr_format_t = input_instr;
 // Global variables
 /* ================================================================== */
 
-UINT64 instrCount = 0;
+volatile UINT64 instrCount = 0;
+UINT64 skipInstrs = 0;
+UINT64 traceInstrs = 1000000;
 
 std::ofstream outfile;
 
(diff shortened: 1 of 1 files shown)


tracer/pin/champsim_tracer.cpp
```diff
@@ -82,7 +84,10 @@ void ResetCurrentInstruction(VOID* ip)
 BOOL ShouldWrite()
 {
   ++instrCount;
-  return (instrCount > KnobSkipInstructions.Value()) && (instrCount <= (KnobTraceInstructions.Value() + KnobSkipInstructions.Value()));
+  if (instrCount > (traceInstrs + skipInstrs)) {
+    PIN_ExitApplication(0);
+  }
+  return (instrCount > skipInstrs) && (instrCount <= (traceInstrs + skipInstrs));
```
ngober, line 90: The second half of the and-statement is always true (the program would have exited if it were false). Either it should be removed or the call to `PIN_ExitApplication()` should go elsewhere.

joshualmashburn: Ah you're right. That was me being overly thorough about it. I'll remove it and ensure it still works.

joshualmashburn: Oddly enough, removing the latter half of the condition on line 90 reintroduces the bug.
I've instead moved the exit check to ResetCurrentInstruction()

joshualmashburn: Tested the change. All appears well.

ngober (approved):
This looks like a good and simple change.

Issue #683, opened 2026-01-29
L1I  is swapped with the ITLB in `O3_CPU`


aaarunan:
**Describe the bug**

I have been trying to make a prefetcher for the L1I and realized the `CACHE::prefetcher_branch_operate` function never got called. Then I noticed when running simulations, the L1I is swapped with the ITLB. In addition looking at `.csconfig/core_inst.cc.inc` the order of caches in `champsim.cc` should be:

0: LLC
1: cpu0_DTLB
2: cpu0_ITLB
3: cpu0_L1D
4: cpu0_L1I
5: cpu0_L2C
6: cpu0_STLB

In fact this order is reversed in the `champsim.cc`.

**To Reproduce**
1. Print the `NAME` of the caches in `champsim.cc` 
2. Print the `NAME` of l1i in `ooo_cpu.cc`

**Expected behavior**
 the l1i `CACHE::prefetcher_branch_operate`  in `ooo_cpu.cc`  should call the function for the L1I prefetcher not ITLB.

**Desktop (please complete the following information):**
 - OS:  6.12.63+deb13-amd64 \#1 SMP PREEMPT_DYNAMIC Debian 6.12.63-1 (2025-12-30) x86_64 GNU/Linux
 - Compiler gcc (Debian 14.2.0-19) 14.2.0
 - Version commit 06de8d3d03d9ba39b4726166aa6364881812365f

maccoymerrell:
Yes, this seems like an issue with how the config script is generating the instantiation file. 

I am under the impression that the l1i and l1d pointers were added later to the core model, and the config file is providing these with the wrong cache pointer. Calling the branch operate hook seems to be it's only purpose, so this seems to be why it hasn't been caught.

The ordering of simulation should mostly be okay due to fill latency in the caches, but you are correct that we were originally enforcing a particular simulation order and that seems to be broken as well.

aaarunan:
It looks like `generated_environment.cc` creates the caches in the reverse order. I fixed this by reversing it after creation.

```c++
auto build(PTWs... builders)
{
  std::forward_list<R> retval{};
  (..., retval.emplace_front(builders));
  retval.reverse();
  return retval;
}
```
This does not affect the L1D as it will be third in the list in both reversed and normal direction.

maccoymerrell:
Yes, this seems to be a result of us iterating forward for the builders but emplace_front() as we go, resulting in a flipped final order.
[...]

PR #684, opened 2026-01-29
Reverse the order after building in `generate_environment.cc`

aaarunan:
This fixes were the caches and PTW are built in the reversed direction addressed in #683 .


ngober:
Resolved with #687

PR #687, opened 2026-02-13
Fix/reversed caches - with tests

### LINKED ISSUE #683: L1I  is swapped with the ITLB in `O3_CPU`

maccoymerrell:
This pr addresses issues identified by @aaarunan, adds testing, and replaces the forward_list containers with regular vectors.

To facilitate testing, the build() function which was previously defined in environment.cc had to be moved to environment.h.


inc/environment.h
```diff
@@ -40,6 +40,14 @@ namespace configured
 {
 template <unsigned long long ID>
 struct generated_environment;
+
+template <typename R, typename... PTWs>
+auto build(PTWs... builders)
+{
+  std::vector<R> retval{};
```
ngober, line 47:
```suggestion
  std::vector<R> retval{};
  retval.reserve(sizeof...(builders));
```

I'm keeping #471 in the back of my mind. Reserving space in this vector will avoid moves caused by the emplace.

ngober (changes requested):
I have a small change request.

ngober (approved):

maccoymerrell:
> I have a small change request.

I wonder if having modules referenced exclusively by name would be worthwhile - we incur the overhead of the std::map string -> pointer conversion but it prevents modules from getting bad pointers (assuming that the map is maintained properly and everything is provided a unique name).

This also relies on enforcing that modules cannot and should not be moved after instantiation. Using a centralized system like what I am trying for module runtime selection could work - but it would need applied to everything.
I guess what I am saying is that moving pointers is probably better than moving actual objects, even when considering the de-reference cost.

ngober:
I'm pretty averse to prohibiting moves. This is a philosophical position: an object's identity is not defined by its location in memory, it is defined by it's value (in the case of a cache, perhaps only its name, but also maybe it's contents). One should be able to construct it and move it around in memory without violating its identity. We're not quite there in ChampSim, but we're pretty close. Most of the time we look through queues to find an identity match. Anyway, that's a topic for another patch.

PR #689, opened 2026-02-14
Drrip set dueling - with tests

maccoymerrell:
Extension of PR from @Quangmire with the following improvements:
1. set sampling duel counter type (dscounter) added, which implements the set sampling duel counter feature utilized by many replacement policies and prefetchers
2. categorizer type, which implements the sampling utilized by 1 as a categorizer, evenly splitting a given value into a sampled number of categories (where 0 and 1 are used as indicators by the dscounter of those sets to be sampled). How input values are translated to integers to be sampled relies on a user-defined projection object.
3. static inlines of get_num_samples() and get_sample_rate() which when provided a population? provides the recommended number of samples and the corresponding sample rate.
4. The addition of a new msl/stat_methods.h header which includes 1, 2, and 3 (and in the future could include other structures like this)
5. Tests for features introduced in 1, 2, and 3


maccoymerrell:
I don't know what is causing this? The includes seem correct.

ngober:
The problem is that you've got `#include "msl/bits.h"`. That defines `champsim::msl::lg2()` and friends. So, you either need to spell it `champsim::msl::lg2()` (fully-qualified) or `lg2()` (since you're already in the `champsim::msl` namespace.

Good idea to pull the set sampling into the MSL, by the way. I anticipate that could be useful.

maccoymerrell:
Before we consider merging this, I should probably do some performance comparisons between lru, drrip, and ship to make sure these are working right.

maccoymerrell:
Realized how I was using the new class in DRRIP was leading to a double-inversion and correlating positively on miss instead of negatively.

DRRIP updates on miss, so an update from a set with one policy should increment the counter towards the other policy. If I just flip the duel result that doesn't work, since the sample sets will get inverted as well so the entire system ends up behaving exactly the same.

Instead, I added a flag to the update() method to invert the way each category increments the counter, which seems to be the most-elegant solution to this problem. Settings the inv flag to true means you train against the policy, and leaving it unspecified/false means you count towards the policy.

Performance checks seem to confirm that this is correct, I am seeing a strong lru < DRRIP < SHiP relationship on the workloads I grabbed from the SHiP paper.

ngober:
Hot take: boolean parameters are an antipattern because they leak a function's control flow to the caller. I think it may actually be a more clear API design to have two update functions: `dscounter::update_good()` and `dscounter::update_bad()`. Calling the parameter "invert" isn't very clear what we should be inverting. Calling it "good" and "bad" tells the user how to use it, rather than what it does.

inc/msl/stat_methods.h
```diff
@@ -0,0 +1,101 @@
...
+#define STAT_METHODS_H
+
+#include <cstdint>
+#include "msl/fwcounter.h"
+#include <cassert>
+#include "extent.h"
+#include "msl/bits.h"
+#include "util/detect.h"
+#include "util/span.h"
+#include "util/type_traits.h"
```
ngober:
```suggestion
#include "msl/bits.h"
```

I think these are unused.

replacement/drrip/drrip.cc
```diff
@@ -9,22 +9,18 @@
 
 drrip::drrip(CACHE* cache) : replacement(cache), NUM_SET(cache->NUM_SET), NUM_WAY(cache->NUM_WAY), rrpv(static_cast<std::size_t>(NUM_SET * NUM_WAY))
```
ngober:
```suggestion
drrip::drrip(CACHE* cache) : replacement(cache), NUM_SET(cache->NUM_SET), NUM_WAY(cache->NUM_WAY), PSEL(NUM_CPUS, champsim::msl::dscounter<long, PSEL_WIDTH>(champsim::msl::get_sample_rate(NUM_SET)), rrpv(static_cast<std::size_t>(NUM_SET * NUM_WAY))
```

Specifying it here would be less verbose.

src/modules.cc
ngober, line 1: Please remove the whitespace change here.

ngober (approved):
I have one final pass of cleanup feedback, but I think this is good to go.

src/modules.cc
maccoymerrell: My git  is not showing any whitespace change at all. I see the change highlighted here, but there isn't actually a change? A direct diff of the modules.cc on develop and the one on this pr also come back identical.

ngober: Try `git checkout develop -- src/modules.cc`

maccoymerrell: That worked

PR #696, opened 2026-03-09
make dib_inorder_width and dib_hit_buffer_size configurable

Easyoakland:
Make `dib_inorder_width` and `dib_hit_buffer_size` configurable with the config.json.

I was trying to configure champsim to get an unrealistically high IPC just to confirm that there aren't missing pieces in my understanding of how everything works and found I couldn't make IPC above 5. I believe the culprit is these configuration values being hard-coded.

### DIFF
--- a/config/instantiation_file.py
+++ b/config/instantiation_file.py
@@ -51,6 +51,8 @@ core_builder_parts = {
     'dib_set': '  .dib_set({dib_set})',
     'dib_way': '  .dib_way({dib_way})',
     'dib_window': '  .dib_window({dib_window})',
+    'dib_inorder_width': '  .dib_inorder_width(champsim::bandwidth::maximum_type{{{inorder_width}}})',
+    'dib_hit_buffer_size': '  .dib_hit_buffer_size({DIB[inorder_width]})',
     'L1I': ['.l1i(&{^l1i_ptr})', '.l1i_bandwidth({^l1i_ptr}.MAX_TAG)', '.fetch_queues(&{^fetch_queues})'],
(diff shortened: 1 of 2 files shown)


config/instantiation_file.py
```diff
@@ -62,7 +64,9 @@
 dib_builder_parts = {
     'sets': '  .dib_set({DIB[sets]})',
     'ways': '  .dib_way({DIB[ways]})',
-    'window_size': '  .dib_window({DIB[window_size]})'
+    'window_size': '  .dib_window({DIB[window_size]})',
+    'inorder_width': '  .dib_inorder_width(champsim::bandwidth::maximum_type{{{DIB[inorder_width]}}})',
+    'hit_buffer_size': '  .dib_hit_buffer_size({DIB[inorder_width]})',
```
ngober, line 69: Could you add a couple tests to `test/python/instantiation_file.py`? At the very least, I'd like to cause a merge conflict with #564 and make sure Maccoy doesn't accidentally regress this.

ngober (changes requested):

config/instantiation_file.py
maccoymerrell: Yes, this will need added to the legacy environment module (explicit should have already fixed this by design).

ngober: What would you prefer here? Would it be better to hold off on this until after #564? It probably has at least a couple more weeks left on it. Or would it be better to put this in real quick and make sure that the legacy environment handles it?

maccoymerrell: You can go ahead and merge this at will, I will add it into the legacy environment when needed.

PR #698, opened 2026-03-15
Ship tag check fix

maccoymerrell:
The previous implementation of the sampler-set tag check mechanism was convoluted, and would result in false matches / false mismatches under some cache configurations.

This pr simplifies the check to the block number, functionally equivalent to what should have been implemented but far simpler and avoids any confusion or potential bugs by introducing additional slicing to remove variable amounts of set index and sampler index bits.

A test was added to ensure the matcher was behaving properly.

Fixes issue raised in #697

### DIFF
--- a/test/cpp/src/445-ship-replacement.cc
+++ b/test/cpp/src/445-ship-replacement.cc
@@ -6,16 +6,20 @@
 TEST_CASE("SHIP sampler matches at cache block granularity")
 {
   /*
-   * The SHIP sampler compares addresses using champsim::block_number so that
+   * The SHIP sampler compares addresses at cache block granularity so that
    * two addresses in the same cache block are treated as the same entry.
    *
-   * With BLOCK_SIZE=64 (LOG2_BLOCK_SIZE=6):
+   * With BLOCK_SIZE=64 (OFFSET_BITS=6):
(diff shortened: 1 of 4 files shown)


test/cpp/src/445-ship-replacement.cc
```diff
@@ -0,0 +1,69 @@
...
+    champsim::address addr2{32};
+    // First access: sampler miss, stores entry for block 0
+    uut.update_replacement_state(0, 0, 0, addr1, ip, champsim::address{}, access_type::LOAD, 0);
+
+    // Second access: same block → sampler HIT → SHCT decremented (clamped at 0)
+    uut.update_replacement_state(0, 0, 1, addr2, ip, champsim::address{}, access_type::LOAD, 0);
+
+    // Sampler hit: SHCT was 0, decremented → clamped to 0
+    // A buggy shamt would treat these as different → sampler miss → SHCT incremented to 1
+    CHECK(uut.SHCT[0][SHCT_idx].value() == 0);
```
ngober: As a general rule, I think it's good to test externally-visible behavior, rather than the values of internal structures. For these tests, that would look like training the replacement on a buggy block, and showing how `find_victim` chooses the wrong way on a follower set. I hope that wouldn't be too difficult.

PR #700, opened 2026-03-25
chore: standardize catch.hpp import

gigamonster256:
catch2 dependency includes should be angle brackets, not double quotes - otherwise you will get errors such as the following

only a few test files had quotes, most already have angle brackets

```bash
champsim> g++ @global.options @absolute.options -MM -MT .csconfig/test/602-walk-latency.d -MT .csconfig/test/602-walk-latency.o -I.csconfig -MF .csconfig/test/602-walk-latency.d test/cpp/src/602-walk-latency.cc
champsim> test/cpp/src/602-walk-latency.cc:1:10: fatal error: catch.hpp: No such file or directory
champsim>     1 | #include "catch.hpp"
champsim>
```

ran clang format to sort the imports

### DIFF
--- a/test/cpp/src/030-address-ops.cc
+++ b/test/cpp/src/030-address-ops.cc
@@ -1,9 +1,9 @@
+#include <catch.hpp>
 #include <iomanip>
 #include <sstream>
 #include <fmt/core.h>
 
 #include "address.h"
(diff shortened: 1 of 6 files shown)


ngober (approved):
This looks like a good change!

Issue #590, opened 2025-02-07
Benchmark History Tracking in Continuous Integration


ngober:
**Is your feature request related to a problem? Please describe.**
Performance is a very important part of ChampSim. If the simulations are slower, research is more difficult. Recently, I've pushed in a few occations ([two](https://github.com/ChampSim/ChampSim/pull/584#issuecomment-2621942858) [examples](https://github.com/ChampSim/ChampSim/pull/521#issuecomment-2408780494)) to add benchmarks to our CI runs. Currently, this merely runs the benchmarks and puts the results in the terminal. Tracking their change over time is a very manual process and their benefits are likely to be missed.

**Describe the solution you'd like**
I would like a solution with at least the following two features:
 - Alerts if the performance of a PR drifts significantly (perhaps in a comment on that PR)
 - Tracks the performance of the benchmarks over time (perhaps in the docs)

The solution would need to interface with the existing body of Catch2 benchmarks and extend to additional benchmarks easily.

**Describe alternatives you've considered**
[This GitHub Action](https://github.com/benchmark-action/github-action-benchmark) seems to do many of these things, but it would probably need to be implemented carefully. Its documentation seems to assume starting from scratch with GitHub Pages.

PR #715, opened 2026-07-07
tools: add benchmark reporting utilities

skku970412:
## Summary
- Add a Catch2 XML benchmark normalizer that writes stable JSON and Markdown reports.
- Add a baseline/current benchmark comparison tool with JSON and Markdown output.
- Add a manual, artifact-only benchmark report workflow.
- Document the local and CI benchmark reporting flow.

## Motivation
Issue #590 asks for benchmark history tracking and PR performance drift visibility. This first pass focuses on durable, machine-readable benchmark artifacts without changing simulator behavior or adding PR comments.

## What changed
- Added `tools/bench/normalize_catch2_benchmarks.py`.
- Added `tools/bench/compare_benchmark_results.py`.
- Added fixture-based Python tests for valid, malformed, duplicate, unit conversion, threshold, zero-baseline, Markdown, and JSON edge cases.
- Added `.github/workflows/benchmark-report.yml` as a manual workflow with read-only permissions and uploaded artifacts.
- Added workflow diagnostics for Catch2 reporters, test discovery, benchmark declaration discovery, generated XML validation, and strict artifact upload checks.
- Added `docs/src/Benchmark-reporting.rst`.

## Out of scope
- No simulator behavior changes.
- No automatic PR comments.
- No benchmark history publishing.
- No default CI gating on benchmark drift.

## Verification
- [x] `git diff --check`
- [x] `python3 -m compileall tools/bench test/python`
- [x] `python3 -m unittest discover -v --start-directory test/python` passed 270 tests with 1 skipped test
- [x] `./config.sh`
- [x] `make -j"$(nproc)" test/bin/000-test-main`
- [x] `test/bin/000-test-main --list-reporters`
- [x] `test/bin/000-test-main --list-tests`
- [x] `grep -R "BENCHMARK" -n test/cpp src inc || true`
- [x] Generated Catch2 XML contained `BenchmarkResults`
- [x] Catch2 XML benchmark output normalized successfully from a real XML run with 9 benchmarks
- [x] Self-compare with `--fail-on-regression` passed with 9 pass, 0 warn, and 0 fail
- [ ] `actionlint` was not available locally
- [ ] GitHub Actions workflow execution is not yet proven in CI. A fork-only push-trigger verification branch created a run, but GitHub did not start the job because the fork account is locked due to a billing issue.

## Follow-ups
- Add optional PR comments after maintainers choose a comment update policy.
- Add benchmark history publishing after maintainers choose storage and retention.
- Decide whether drift thresholds should ever fail CI.


ngober:
This is an interesting patch. It might do what the Issue is asking for. I have a couple of initial concerns:

 - Putting all of this code in the ChampSim repository means we need to maintain it over time. The principle of separation of concerns would suggest that this might be better as a standalone Github Action that we link with the `uses` key. But, then, why would this be better than an existing solution (such as the one I linked in the issue)? It may reinvent the wheel in that regard.
 - Why did you choose to convert the XML files to JSON, only to store the JSON files and compare them. Would it not make more sense to store the XML files and ingest them each time?

PR #716, opened 2026-07-08
Fix scheduler and pipt timing

maccoymerrell:
Two fixes for bugs discovered during optimizations.

1. Rename happens IN-ORDER, so walks stop once they cannot allocate registers for an instruction. **However, this check was happening against all instructions, even scheduled ones**, resulting in some scheduler walks ending earlier than they should. Performance impact was minimal as long as the scheduler was not the bottleneck of the machine (or for that matter, the RAT table was at a realistic size).

2. Translation timing and tag check timing were allowed to overlap. At the moment, caches are modeled PIPT so this was not correct. Impacted L1D latency, scaled with DTLB/ITLB latency (typically only 1-2 cycles), minimal IPC impact on realistic configurations where L1D latency >> DLTB/ITLB latency.

Tests were added for each.
Follow up work on caches may reintroduce this timing path for VIPT caches when implemented.

### DIFF
--- a/test/cpp/src/202-scheduler-rename-gate.cc
+++ b/test/cpp/src/202-scheduler-rename-gate.cc
@@ -1,0 +1,74 @@
+#include <catch.hpp>
+
+#include <algorithm>
+#include <array>
+#include <cstdint>
+
+#include "instr.h"
+#include "mocks.hpp"
+#include "ooo_cpu.h"
+#include "defaults.hpp"
+
+// The free-physical-register rename gate must apply only to instructions being
+// renamed this cycle. An already-scheduled instruction has already claimed its
+// registers, so evaluating the gate against it -- and breaking scheduling when
+// the free pool is smaller than its destination count -- wrongly stalls ready
+// instructions sitting behind it in the ROB whenever registers are scarce.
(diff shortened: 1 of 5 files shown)


test/cpp/src/202-scheduler-rename-gate.cc
```diff
@@ -0,0 +1,67 @@
...
+
+    uut.ROB.push_back(ahead);
+    uut.ROB.push_back(behind);
+    for (auto& instr : uut.ROB)
+      instr.ready_time = champsim::chrono::clock::time_point{};
+
+    // Exhaust every free physical register.
+    for (uint64_t filler_id = 100; uut.reg_allocator.count_free_registers() > 0; ++filler_id) {
+      uut.reg_allocator.rename_dest_register(10, filler_id);
+    }
```
ngober: I think it might make more sense to fill the rob with instructions rather than manually going in and exhausting all of the physical registers. It would be more representative of the failure case that we're trying to demonstrate.

ngober (commented):
These look like real bugs and the fixes seem appropriate.

ngober (approved):

maccoymerrell:
@ngober Glad to see this approved. Once this is merged, I will open the frontend modularization pr.

PR #718, opened 2026-07-22
Fix heap-buffer-overflow in SPP prefetcher

LokeshR926:
## Bug Description
The SPP prefetcher has a heap-buffer-overflow issue in `PATTERN_TABLE::read_pattern()` at line 363. The `pf_q_tail` index is incremented without bounds checking against the vector sizes, causing out-of-bounds writes when many prefetch candidates are generated.

## Root Cause
In `PATTERN_TABLE::read_pattern()`, the code writes to `delta_q[pf_q_tail]` and `confidence_q[pf_q_tail]` and then increments `pf_q_tail` without checking if it exceeds the vector size.

## Fix
Added bounds checking before writing to the vectors:
```cpp
if (pf_q_tail >= delta_q.size() || pf_q_tail >= confidence_q.size()) {
    break; // Queue is full, stop adding entries
}


ngober:
Thanks for the patch!

PR #719, opened 2026-07-28
Frontend modularization and orchestration rework

maccoymerrell:
Modularize the frontend and stats system, rework orchestration with origin provenance and scoped globals, add the identity registry and stream sources. Carries the scheduler and PIPT bugfixes forward. Behavior matches the prior compile-time configurations.


maccoymerrell:
Valgrind check, why did you betray me?

inc/util/type_traits.h
ngober, line 1: Are these only formatting changes to this file? Why make them, when clang-format will simply undo them?

test/config/explicit-1core-altmodules.json
ngober, line 1: Was this reformat necessary?

src/main.cc
```diff
@@ -86,8 +94,7 @@ int main(int argc, char** argv) // NOLINT(bugprone-exception-escape)
   }
 
   // Enable dump mode if requested
-  if (knob_dump)
-    fmt::print("=== Module Builder Dump ===\n");
+  if (knob_dump) fmt::print("=== Module Builder Dump ===\n");
```
ngober: Here's another arbitrary reformat.

ngober (changes requested):
This patch is enormous. It would be much easier to read if you didn't include a bunch of formatting changes. Note that clang-format will catch any inconsistencies.

ngober:
Valgrind is a strict tutor, but you'll be better for her discipline.

maccoymerrell:
This is happening because of some unfortunate issues with the way I split my branch into the 3 commits, I will fix it when I can, including the random formatting changes.

maccoymerrell:
Running some quick checks to ensure we are still byte identical, and stripping out some excessive comments.

maccoymerrell:
Cut a good portion of changed lines due to excessive commenting or formatting changes.
Also confirmed output is byte-identical, unchanged from develop for single-core, stat collection changed for multi-core due to an implicit bugfix. Previously, multi-core would recapture stats for every core when any core finished, which resulted in the final roi stats being from the entire execution of every core up until that point, instead of each core's independent ROIs. This was fixed out of necessity under the new system, stat collection hooks are invoked per-core, not broadcast.

maccoymerrell:
Doing some cleanup, demangling, and fixing some naming schemes.

maccoymerrell:
@ngober Okay, I have cleaned this up a lot. Should be more palatable.

inc/identity_registry.h
```diff
@@ -0,0 +1,102 @@
+/*
+ *    Copyright 2023 The ChampSim Contributors
```
ngober: The copyright on a new file should probably be the current year. (I wish I had a robot to maintain this)

inc/listener.h
```diff
@@ -0,0 +1,55 @@
...
+#include <vector>
+
+#include "modules.h"
+
+namespace champsim::modules
+{
+/**
+ * Listener interface — observes run-wide events for reporting.
+ *
+ * Listeners are ordinary modules: declare them as top-level children in an
```
ngober: I appreciate the temptation to define listeners as modules, but I think we need to be concerned about the user experience here. See [my comment](https://github.com/ChampSim/ChampSim/pull/483#discussion_r1660052729) from when these were originally introduced. The listener is not a module in the same way that a prefetcher is. We don't expect the user (in the common course of using ChampSim) to roll their own listeners, though they may want to do so.

inc/ptw.h
```diff
@@ -30,20 +30,29 @@
...
     std::size_t level;
+    // The address space this cached walk step belongs to. A walker is
+    // hardware owned by a consumer, not by an address space: entries are
+    // stream-tagged so concurrent streams never hit each other's steps.
+    champsim::origin::id_type stream = 0;
   };
 
   struct pscl_indexer {
     champsim::data::bits shamt;
-    auto operator()(const pscl_entry& entry) const { return entry.vaddr.slice_upper(shamt); }
```
ngober, line 43: Why was this line changed? You are throwing away the typed guarantees of the address and going back to integers.

inc/util/stat_format.h
ngober, line 1: Toss this in `inc/util/`. It's just a small utility.

ngober (changes requested):
A coarse-grained pass, mostly focused on design:

 - I like the idea of a `champsim::origin` instead of a `std::array<uint8_t, 2>` for an address space identifier. You warm my heart with your strong types.
 - Some names... I like the idea of "phase controllers". This concisely describes what it does. However, "token" seems to me to be the wrong word to describe this. Despite being in vogue, it implies that the unit of work is standalone and fungible, which these aren't. Perhaps "message" or "packet" would make more sense, because those would communicate that it is one unit in a stream. Also, the dual of "source" is conventionally "sink". A "consumer" is usually dualled with a "producer".
 - I'm worried that listeners got unfairly caught up in Claude's enthusiasm. The basics for what you're trying to do here don't seem to have much to do with them. Can you explain why they were moved around so much?

maccoymerrell:
> A coarse-grained pass, mostly focused on design:
> 
> * I like the idea of a `champsim::origin` instead of a `std::array<uint8_t, 2>` for an address space identifier. You warm my heart with your strong types.
> * Some names... I like the idea of "phase controllers". This concisely describes what it does. However, "token" seems to me to be the wrong word to describe this. Despite being in vogue, it implies that the unit of work is standalone and fungible, which these aren't. Perhaps "message" or "packet" would make more sense, because those would communicate that it is one unit in a stream. Also, the dual of "source" is conventionally "sink". A "consumer" is usually dualled with a "producer".
> * I'm worried that listeners got unfairly caught up in Claude's enthusiasm. The basics for what you're trying to do here don't seem to have much to do with them. Can you explain why they were moved around so much?

I will work on these when I can. Packet is probably best then, I was thinking tokens in terms of how a text processor would work, but packet is probably better. Producer/consumer is probably best here, unless source/sink has any advantages I am not seeing.

I missed that event listeners were caught up in this. I don't think it is necessarily the case they need to remain exactly as they were (particularly because the heartbeat is tied into them); I will think on this. They don't need to be modules, I need to make sure they mesh with the phase controllers properly.

maccoymerrell:
@ngober @dpuckett98 Looking through the logs, the listeners were pulled in was because of the shape and the fact I had touched the phase orchestration when I modularized it, which was tied into the heartbeat event listener. The templated hooks matched the old module shape, so Claude pulled them into the module system. I don't think this was wrong, but I think that the implementation of the listener module was inverted from what it should have been.

Looking at the code, there are two routes I can go with (excluding just reverting to develop), but I would prefer not to leave the listeners in their current state. They reintroduce the failure mechanisms that the new module system avoids (unenforced interfaces, silent hook mismatches). Regardless of any final shape, the heartbeat listener has to be modified to work with the new phase system, and that means being instruction agnostic (report packet heartbeats, not instruction heartbeats) or adding new instruction-specific event enums.

1. Listener system (we can use the module system, nothing prevents them from coexisting in the same tables), but instead of event -> listener modules, we do event modules -> listener. The listener system allows for the registering of events and their respective callbacks, but the listener class itself remains outside the module system. This costs overhead from the indirection at the event_type ->  callback over the template approach. In practice, I didn't see any performance difference when doing this with the other module hooks, so I think it is fine. Keeps all the event listener features: lightweight code and cl-configurable silencing. New event types can be registered anywhere alongside their callbacks. register_event<event_type>(callback).

2. Listener system remains templated, but I introduce loud errors when a hook mismatch happens. Adding new events and hooks requires editing the listener file. Existing heartbeat callback is curated towards instruction packets only, will need to generalize or add instruction-specific events inside the listener.

maccoymerrell:
Fixed a major interface issue with the phase controller, much cleaner now. I also bundled stats and phases tightly together. I reverted the listener modularization, pending what that final shape should be.

Not entirely sold on that shape either. I left stats and phases disjoint, so it is never up to the phase controller to collect stats, but that leaves this awkward ferrying of phase-controller-specific info into the champsim.cc loop when stat collection actually takes place.

dpuckett98:
While I'm still digging into the details of this PR, I do think moving event listeners into the new module system is a good idea. Performance characterization is an active research area, and event listeners are the way that we implement performance characterization techniques in ChampSim. If we elevate listeners to first-class citizens alongside prefetchers, replacement policies, and branch predictors, then we encourage performance characterization research in ChampSim. Additionally, when I use event listeners, I tend to repeatedly modify some of their configuration parameters in some .cc or .h file then recompile the binary. If listeners are moved into the module system, then I can change their configuration without recompilation.

maccoymerrell:
Simplified the stats interface further, I think it is far cleaner now.
I also implemented something that I hope satisfies both of y'all's requests.

Listeners are split into two parts. Hooks and listeners.
Hooks are subscription-based, they are registered with a particular function invocation but there is not necessarily something listening on it. So you can include lots of hooks in modules, but they are effectively nops without a listener.

A listener is a module, and it subscribes to one or more hooks, when that hook is invoked it triggers a callback (lambda function matching hook invocation). Since listeners are modules, they are enabled/disabled by being defined, rather than the old --listeners argument (as dpuckett requested).

The legacy environment by default creates a heartbeat listener (unless heartbeat is disabled), the new environment requires you to manually define the module somewhere in the config.
@dpuckett98 Thoughts?

PR #722, opened 2026-09-04
Fix strict weak ordering comparator in ChampSim DRAM channel

abejgonzalez:
`DRAM_CHANNEL::populate_dbus`uses `std::min_element` to select the next ready bank request. The previous comparator returned `!rhs.valid || (lhs.valid && lhs.ready_time < rhs.ready_time)`, which evaluates to true when both lhs and rhs are invalid. This violates irreflexivity and asymmetry required by strict weak ordering. Thus require lhs.valid so invalid requests are never considered strictly less than other invalid requests.

### DIFF
--- a/src/dram_controller.cc
+++ b/src/dram_controller.cc
@@ -249,7 +249,7 @@ long DRAM_CHANNEL::populate_dbus()
   long progress{0};
 
   auto iter_next_process = std::min_element(std::begin(bank_request), std::end(bank_request),
-                                            [](const auto& lhs, const auto& rhs) { return !rhs.valid || (lhs.valid && lhs.ready_time < rhs.ready_time); });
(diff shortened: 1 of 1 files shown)


ngober:
Once upon a time, we had some adaptors that looked like `invalid_is_minimal([](const auto& lhs, const auto& rhs) { return lhs.ready_time < rhs.ready_time; })` that would handle this. I think I removed them and thought "It's too verbose, and I won't make that mistake easily anyway." No programmer should think such thoughts.

Anyway, thanks for the patch!

maccoymerrell:
we need this on develop too, right @ngober? Do you want to handle that on the next release or do it now?

ngober:
I just merged master into develop. It's in both branches now.

